PR #2, opened 2015-01-23
host: Use subparsers rather than arguments

cgwalters:
This allows matching the existing model for /usr/bin/rpm-ostree:
 - atomic upgrade -r
 - atomic status
etc.


rhatdan:
Will need man pages for this now.

cgwalters:
Er yes, sorry I meant `atomic host upgrade`, etc.

PR #8, opened 2015-01-27
If no image name specified, it should print Default configuration details

jeremyeder:
To fix this:

```
-bash-4.2# ./atomic defaults
  File "./atomic", line 355, in <module>
    atomic.set_args(args)
  File "./atomic", line 118, in set_args
    self.name = self.image.split("/")[-1].split(":")[0]
AttributeError: 'NoneType' object has no attribute 'split'
```

### DIFF
--- a/atomic
+++ b/atomic
@@ -115,9 +115,10 @@ class Atomic:
             pass
 
         if not self.name:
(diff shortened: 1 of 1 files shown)


cgwalters:
This looks like it's fixed.

PR #12, opened 2015-01-28
atomic: Print help by default if no arguments supplied

cgwalters:
Thanks to http://stackoverflow.com/questions/4042452/display-help-message-with-python-argparse-when-script-is-called-without-any-argu

### DIFF
--- a/atomic
+++ b/atomic
@@ -312,9 +312,15 @@ def SetFunc(function):
             setattr(namespace, self.dest, function)
     return customAction
 
+class HelpByDefaultArgumentParser(argparse.ArgumentParser):
+    def error(self, message):
+        sys.stderr.write('error: %s\n' % message)
+        self.print_help()
+        sys.exit(2)
+
 if __name__ == '__main__':
     atomic=Atomic()
-    parser = argparse.ArgumentParser(description=atomic.help())
+    parser = HelpByDefaultArgumentParser(description=atomic.help())
     subparser = parser.add_subparsers(help=_("Commands"))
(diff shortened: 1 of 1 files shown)


rhatdan:
I am not crazy about this though, I thought it was less complete.  The problem I have with this type of interaction is as the commands become more complex the data scrolls off the screen.

rhatdan:
I prefer this type 

 ls --foobar
ls: unrecognized option '--foobar'
Try 'ls --help' for more information.
[Exit 2]

cgwalters:
Right now it's just 18 lines, enough to fit even within the default 24 terminal lines.  Running `docker` is 89, which is getting to be a lot.  `systemctl` shows you the units by default, which I never use.

If it gets longer, we can customize it to be shorter?

rhatdan:
But I still loose the error in the noise.

rhatdan:
How about we put the error at the bottom rather then the top

rhatdan:
./atomic
usage: atomic [-h] {defaults,host,info,install,run,uninstall,update} ...

Atomic Management Tool

positional arguments:
  {defaults,host,info,install,run,uninstall,update}
                        Commands
    defaults            list default commands with which Atomic will
                        RUN/INSTALL/UNINSTALL containers
    host                execute Atomic host commands
    info                dispaly label information about an image
    install             execute container image install method
    run                 execute container image run method.
    uninstall           execute container image uninstall method
    update              pull latest container image from repository

optional arguments:
  -h, --help            show this help message and exit

error: too few arguments

cgwalters:
Error at bottom is :+1: from me

PR #19, opened 2015-02-18
build: "all" should be default, don't run tests on install

cgwalters:
Neither "make" or "make install" should not run tests.  Further,
change the default target to be "all" by listing it first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,10 @@ SYSCONFDIR ?= $(DESTDIR)/etc/sysconfig
 PROFILEDIR ?= $(DESTDIR)/etc/profile.d
 PYTHON ?= /usr/bin/python
 
+all: python-build
+
 test:
 	sh ./test.sh
-
-all: python-build
 
 python-build: atomic
 	$(PYTHON) setup.py build
@@ -17,7 +17,7 @@ clean:
 	$(PYTHON) setup.py clean
 	-rm -rf build *~ \#* *pyc .#*
 
-install: test all 
+install: all 
 	$(PYTHON) setup.py install `test -n "$(DESTDIR)" && echo --root $(DESTDIR)`
 	[ -d $(SYSCONFDIR) ] || mkdir -p $(SYSCONFDIR)
 	install -m 644 atomic.sysconfig $(SYSCONFDIR)/atomic


lsm5:
LGTM, tests aren't run in rpmbuild by default with this update.

cgwalters:
That's what `%check` is for, no?  And we can't run the tests anyways in koji/mock/rpmbuild as they require docker which requires root...

lsm5:
yup, the test could be placed in `%check` but docker and its root requirement would pose a problem

rhatdan:
We can do it like we do the docker test, so that we only do the check if the docker socket is available.

rhatdan:
Then we can run the tests in Jenkens.

lsm5:
@rhatdan hmm, I guess as long as mock doesn't pull in docker given that it's a runtime requirement and trigger tests, we're probably fine. I'll give this a shot and get back

cgwalters:
How about a new target `make rootcheck` or something?

cgwalters:
(Because we *could* have tests that run as non-root, e.g. internal mocked unit tests and the like)

lsm5:
haven't tried in mock yet, but on a non-mock build, `make test` did show me some errors:

    atomic_test
    PID   USER     COMMAND
        1 root     /bin/sh
        9 root     ps
    Error response from daemon: Conflict, cannot delete 9dd6ed7c29a4 because the container 3152734b6250 is using it, use -f to force
    FATA[0000] Error: failed to remove one or more images


Also:

    Step 0 : FROM busybox
     ---> 4986bf8c1536
    Step 1 : LABEL 
    # Skipping unknown instruction LABEL
     ---> 4986bf8c1536
    Step 2 : LABEL 
    # Skipping unknown instruction LABEL
     ---> 4986bf8c1536
    Step 3 : LABEL 
    # Skipping unknown instruction LABEL
     ---> 4986bf8c1536

lsm5:
so running tests in check works fine, mock doesn't complain either.

scratch build results: http://koji.fedoraproject.org/koji/taskinfo?taskID=9000034
updated spec: https://github.com/lsm5/atomic-rpm/blob/master/atomic.spec

rhatdan:
LABEL would currently require the LABEL patch, which is only in RHEL7 at the moment.

PR #20, opened 2015-02-18
build: Move manpage rules upstream here

cgwalters:
From initial code by Lokesh Mandvekar in https://github.com/lsm5/atomic-rpm


lsm5:
yup, this works great :+1:

PR #26, opened 2015-03-13
use /usr/local as prefix in Makefile

lsm5:
This will keep the 'make install'-ed package isolated from distro managed
packages.

rpms can use PREFIX=%{prefix} to install to /usr.
My guess is other formats can do something similar too.

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>


lsm5:
ping @rhatdan @cgwalters

rhatdan:
WHy would I want to do this?

PR #33, opened 2015-03-30
Redirect stderr to /dev/null instead of creating an unread pipe

cgwalters:
(Note, only lightly tested - the test suite is failing for me right now even without this patch, due to what I think is the automatic prefixing with `docker.io` for images)

If the docker client output enough stderr, we'd end up with a
deadlock.  I think the original idea here was to suppress error
messages - if so, /dev/null works.  Perhaps in the future though we
should store the error buffer somewhere in case it is needed.

### DIFF
--- a/atomic
+++ b/atomic
@@ -27,6 +27,11 @@ import gettext
 import docker
 import json
 import subprocess
+try:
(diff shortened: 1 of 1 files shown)


cgwalters:
Looks fine to me, I rebased with that suggestion.

PR #40, opened 2015-04-16
atomic-run: Add --volume option

cgwalters:
First, this is a horrendous hack.  We need something like a `RUNOPTS`
metadata as a JSON array or something that can be clearly appended to
instead of having shell script.

Second and more importantly, it doesn't actually work because if the
container is stopped we have no way to change its mount points
currently.  We'd have to detect the case where additional mounts were
passed in on the command line - and AFAIK Docker doesn't support
dynamically changing the mounts right now, though it would be possible
at the kernel level.


rhatdan:
I think the --opt1, --opt2, --opt3 should take care of this now.

PR #42, opened 2015-04-23
fixed a tiny typo

goern:
its SYS_TIME and not SYS_TYPE


cgwalters:
Thanks, it's about time for this fix.

PR #46, opened 2015-04-24
Adding the install/build steps for Fedora

LalatenduMohanty:
Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

### DIFF
--- a/docs/install/Fedora.md
+++ b/docs/install/Fedora.md
@@ -1,0 +1,31 @@
+#INSTALL
+atomic can be installed through below methods
+
+##YUM/DNF
+atomic RPM is available on Fedora 21 and later. So a simple `yum/dnf install atomic` will install it fine.
+
+##Make
+On Fedora, You need to install required build dependencies
+```
+yum-builddep atomic
+```
+Or using DNF
+```
+dnf builddep atomic
+```
+To use the builddep plugin in DNF you need to install dnf-plugins-core
+```
+dnf install dnf-plugins-core
+```
+
+Get the code
+```
+git clone https://github.com/projectatomic/atomic.git
+cd atomic
+```
+Build and install it.
+````
+make all
+
+make install
+```


docs/install/Fedora.md
```diff
@@ -0,0 +1,23 @@
+#INSTALL
+atomic can be installed through below methods
+
+##YUM/DNF
+atomic RPM is available on Fedora 21 and later. So a simple `yum/dnf install atomic` will install it fine.
+
+##Make
+On Fedora 21, You need to install required build dependencies 
+```
+yum install pylint python-docker-py go-md2man
```
cgwalters: Why not recommend `yum-builddep atomic`?  That's what I do to get a starting point.

lsm5: ah yup, `yum-builddep atomic` or `dnf builddep atomic` would work better than yum install.

I think all yum commands including `yum-builddep` have been deprecated starting with f22, so might need to replace with `dnf builddep`. yum would still work but will just throw a deprecation warning

LalatenduMohanty:
@cgwalters @lsm5 Thanks for the suggestion. I will update the PR. However `dnf builddep atomic` is as of now not working as expected (refer https://github.com/rpm-software-management/dnf-plugins-core/blob/master/doc/builddep.rst)

# dnf builddep atomic
Using metadata from Mon Apr 27 12:21:22 2015
Error: Failed to open: 'atomic', not a valid spec file.

lsm5:
ahh, that's probably only a rawhide thing in that case, my bad :|

rhatdan:
@lsm5 Is this fixed.  @LalatenduMohanty This should probably be a bugzilla not in here.

lsm5:
@rhatdan this is just a documentation update to be added to the upstream repo itself and not a bug.

rhatdan:
ok

lsm5:
@rhatdan could you please re-open (and hopefully merge) this?

PR #48, opened 2015-04-25
Add ability to upload/push to a registry

rhatdan:
This patch will allow users to upload to either a docker registry or
to a pulp crane server.  We want to support Sattelite in the future as
well.

Will read authorization data out of ~/.pulp/admin.conf if it exists, otherwise
it will prompt for Username and Password.

### DIFF
--- a/atomic
+++ b/atomic
@@ -27,6 +27,8 @@ import gettext
 import docker
 import json
 import subprocess
+import atomicpulp
+import getpass
 try:
     from subprocess import DEVNULL # pylint: disable=no-name-in-module
 except ImportError:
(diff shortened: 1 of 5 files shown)


rhatdan:
@cgwalters Any chance you could give this a quick review?
@aweitieka?

cgwalters:
I'm a fan of making it easier to upload to Pulp, but why are we wrapping `docker push`?

cgwalters:
Though this does pull in the pulp client tools (and a whole other bag of Python libraries) into the host.

Can we make it easier to install/run the Pulp tools as a container?  That's already been prototyped some in https://github.com/pulp/pulp_packaging/tree/master/dockerfiles

cgwalters:
Also, (and I'm just speaking personally) - I think `docker push` is something for developers to use for quick feedback in staging environments.  Production containers should be reliably built from source code via continuous delivery services using well-known and stable build environments.  In other words you want some server doing Jenkins -> push to pulp -> exported as read-only registry.

`docker push` from developer laptops and the like means you're at grave risk of having binaries uploaded without the corresponding source code, etc.

aweiteka:
The code LGTM.

1. No opinion on wrapping `docker push`
1. Running pulp-admin CLI or a more container-focused pulp client as a container is very doable. The issue is it presents another barrier to using pulp as a registry. Atomic CLI is "just there."
1. I agree 100% the ideal workflow is to *never* push a locally built image from a dev workstation to any type of "real" registry server. That said it is really nice to have a tool available in non-ideal environments. I image this would be used less frequently on atomic host and more common on a traditional OS like RHEL server.

rhatdan:
> Though this does pull in the pulp client tools (and a whole other bag of Python libraries) into the host.
> 
> Can we make it easier to install/run the Pulp tools as a container? That's already been prototyped some in https://github.com/pulp/pulp_packaging/tree/master/dockerfiles

No we are shipping pulp.py which is all we need.

Issue #49, opened 2015-05-07
Currently there is no way atomic run can take a shell environment variable like docker run can


arilivigni:
Use case.

I add a label with docker run to an image/Dockerfile 

ex.

LABEL RUN="docker run -it --name NAME --privileged -e PS1=\"[\\u@\\h \\W]\\$\" -e PROJECT_DEFAULTS=ci-ops-projex/config/project_defaults -e SSH_KEYFILE=ci-factory/targets/keys/ci-ops-central -e RESOURCES_DIR=/root/workspace/ci_dev_env -v \${CI_DEV_ENV}:/root/workspace/ci_dev_env IMAGE"

On my system I then export CI_DEV_ENV=/home/ari/workspace

Then I do an atomic run it picks the variable and replaces it for the running container.

stc proposed adding a -D key=value for doing this with atomic run which makes sense.

PR #52, opened 2015-05-12
When running atomic commands we need to pass the users environment

rhatdan:
The run and install commands were not seeing the environment, this will
allow users to pass environment variables into atomic commands.

### DIFF
--- a/atomic
+++ b/atomic
@@ -305,7 +305,7 @@ removes all containers based on an image.
         return command
 
     def _new_env(self):
-        env = os.environ
+        env = dict(os.environ)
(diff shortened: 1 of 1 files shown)


rhatdan:
@cgwalters @ari could you take a look at this.

arilivigni:
LGTM this will be extremely helpful!

cgwalters:
Otherwise looks like a nice cleanup to me, so :+1: with the above change.

miabbott:
This appears to address #49.  Can it be closed?

PR #53, opened 2015-05-14
Move the Atomic class into python library directory.

rhatdan:
I am beginning to work on dbus bindings, this will allow the atomic cli and
the atomic_dbus_server app to share the same code.

Renamed atomicpulp to Atomic in the python library.


cgwalters:
Capitalization looks unusual (but not unheard of) for Python libraries.  Now of course there is at least https://pypi.python.org/pypi/atomic/0.7.2

`atomiccontainer`?

Dunno...naming things is hard =)  I'm fine with the patch as is too if we want to just ignore the conflict with that random PyPi module.

Python really made a mistake by not encouraging namespacing like Java did, but anyways.

rhatdan:
Lets leave it as is and change if we need to later.

PR #54, opened 2015-05-15
Use image metadata if container does not exist on atomic stop

vpavlin:
When you call atomic stop it expects existing container which does not
have to be always true. We `--rm` our container in label `RUN` for atomicapp
because it's not useful after it finishes it's work.

This commit checks if the inspect of container was successful and if
not, it uses loads image metadata and uses `STOP` label from it

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -340,7 +340,12 @@ removes all containers based on an image.
 
 
     def stop(self):
+        inspected_image = False
         self.inspect = self._inspect_container()
+        if self.instpect is None:
+            self.inspect = self._inspect_image()
+            inspected_image = True
+
         args = self._get_args("STOP")
         if args:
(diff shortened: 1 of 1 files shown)


vpavlin:
That actually does not solve my issue at all because that just blocks me from using stop completely. All I want is for stop not being strictly tied to running container. We in atomicapp run a container that delegates the work down to kubernetes which means the atomicapp in a container does only need to do some calls and than can die. And if I want to undo what I did in kubes, I need to start a container, it does few calls to kubes and dies again. 

But as container dies after run, I cannot use stop anymore (although all the stuff is running in kubes). We could use uninstall, but that would break UX - `atomic run` to start and `atomic uninstall` to stop? I don't think so..

So all I need is that `atomic` queries not only container, but also an image for `STOP` label.

PR #56, opened 2015-05-18
Allow stop to execute STOP command if container does not exist.

rhatdan:


rhatdan:
@vpavlin Does this fix your problem?

vpavlin:
It does, yes:)

PR #58, opened 2015-05-27
Pass additional arguments to the atomic install/uninstall line

rhatdan:
This patch will allow a user to pass additional arguments to the command specified in the INSTALL/UNINSTALL lines


markllama:
LGTM (drool)

PR #59, opened 2015-05-27
Allow users to add ${IMAGE} and ${NAME} to INSTALL/RUN/STOP/UNINSTALL…

rhatdan:
… labels

This will allow users to add random content based off image and container names

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -417,6 +417,8 @@ removes all containers based on an image.
     def gen_cmd(self,cargs):
         args = []
         for c in cargs:
+            c = c.replace("${IMAGE}", self.image)
+            c = c.replace("${NAME}", self.name)
             if c == "IMAGE":
                 args.append(self.image)
                 continue
--- a/docs/atomic-install.1.md
+++ b/docs/atomic-install.1.md
@@ -16,12 +16,13 @@ IMAGE, if this field does not exist, `atomic install` will install the IMAGE
 
 If the container image has a LABEL INSTALL instruction like the following:
 
-```LABEL INSTALL /usr/bin/docker run -t -i --rm --privileged -v /:/host --net=host --ipc=host --pid=host -e HOST=/host -e NAME=NAME -e IMAGE=IMAGE -e CONFDIR=${CONFDIR} -e LOGDIR=${LOGDIR} -e DATADIR=${DATADIR} --name NAME IMAGE /bin/install.sh```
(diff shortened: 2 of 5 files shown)


stefwalter:
The regular expression for a basic expansion of variables isn't that hard:

https://github.com/cockpit-project/cockpit/blob/master/pkg/base1/cockpit.js#L2142

fche:
Both a substring-search ${FOO} and simple regex methods suffer from poor quoting-tolerance.  (Imagine a user wants to pass the literal string ${FOO} as an argument, by single-quoting it or \-prefixing at some point.)

rhatdan:
Are either of these comments actually related to the pull request?

stefwalter:
> Are either of these comments actually related to the pull request?

Yes, the variable substitution method is brittle and prone to error.

shaded-enmity:
I think that the simplified way how TeX works can be used to achieve proper substitutions:

```python
class Substitute(object):
       def __init__(self, s = ''):
               self.data = s

       def expand(self):
               return Substitute(self.data.replace(...))

       def __eq__(self, other):
               return self.data == other.data

sub, old = Substitute(""), None
while sub != old:
        old, sub = sub, sub.expand()
```
I.E: keep expanding until both before/after strings are the same

But shouldn't that be a subject for another PR? the original LGTM.

rhatdan:
Lets get the original merged and then we can work on improving it.

fche:
Dan, have you considered making this work by having the atomic driver arrange to put into the environment those two values, and let the INSTALL/RUN/... scripts to find it there (with the same single-quote trick in your 'using-environment-substitution' blog?  So instead of massaging the command strings, automatically pass a standard -e VAR1=value1 -e VAR2=value2 tuple to docker.

rhatdan:
On 05/29/2015 09:32 AM, Frank Ch. Eigler wrote:
>
> Dan, have you considered making this work by having the atomic driver
> arrange to put into the environment those two values, and let the
> INSTALL/RUN/... scripts to find it there (with the same single-quote
> trick in your 'using-environment-substitution' blog? So instead of
> massaging the command strings, automatically pass a standard -e
> VAR1=value1 -e VAR2=value2 tuple to docker.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/projectatomic/atomic/pull/59#issuecomment-106803396>.
>

I guess that would work also.  It would leak the information into the
container, but probably not a bad idea.

rhatdan:
I like @DirectXMan12 fix better.

rhatdan:
https://github.com/projectatomic/atomic/issues/66

PR #60, opened 2015-05-27
First pass at a portable docker-mount API for image mounting/umounting.

baude:


rhatdan:
We have merged in modified version of this by @willmtemple 
So closing.

PR #62, opened 2015-05-29
Basic atomic mount/unmount commands

willmtemple:
CLI integration with python docker_mount module provided by @baude. This PR includes PR#60 (and docker_mount.py here is more up-to-date than the one in PR#60).

Mount a container image via `atomic mount repo/image[:tag] /path/to/mountpoint`, unmount via `atomic unmount /path/to/mountpoint`

Would like thoughts on how to eliminate the required repo in the image name, I would think that `atomic mount fedora /mnt` would be valid syntax. Not sure how to check the registries (or how to prioritize registries with identically named images)

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -7,6 +7,8 @@ import subprocess
 import getpass
 import requests
 import pipes
+import mount
+import selinux
 try:
     from subprocess import DEVNULL # pylint: disable=no-name-in-module
 except ImportError:
@@ -183,6 +185,14 @@ class Atomic(object):
             if self.spc:
                 self.name = self.name + "-spc"
 
+    def _get_selinux_label(self):
+        if selinux.is_selinux_enabled() != 0:
+            fd = open(selinux.selinux_lxc_contexts_path())
+            for i in fd.readlines():
+                name,context = i.split("=")
+                if name.strip() == "file":
+                    return context.strip("\n\" ")
+        return ""
 
     def _getconfig(self, key, default=None):
         assert self.inspect is not None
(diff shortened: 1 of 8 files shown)


willmtemple:
@rhatdan Rewrote the history into a single patch. Changed how the DockerMount class uses exceptions (it was originally using them just as an error mechanism internally).

I didn't intend to remove your comments; it was a side effect of updating the PR and removing the other commits.

willmtemple:
@rhatdan updated with bind mounting rootfs/ to given mount point via a bind mount, completions, man pages. Unmount is now significantly more robust and can deal with cases where a user mounts another filesystem over/under the same mountpoint or mounts within the container's filesystem.

docker_mount.py is now mount.py

_namesearch() is much more effective now, and is capable of determining when a given name input does not refer to a unique image, and offers a list of possible alternatives. This prevents decision problems i.e. when two registries have different versions of identically named and tagged images.

rhatdan:
White space cleanup these patches also.

willmtemple:
@rhatdan: updated with default selinux context, mount options, completions for -o flag, man page attribution.

willmtemple:
@rhatdan The test script does not run on my machine. I have written some tests that should demonstrate the core functionality, but have no indication of whether or not they perform, since the tests fail with keyerrors.

rhatdan:
I fixed the problem.  Please update your pull and we can merge it.

willmtemple:
@rhatdan, I added my tests. I confirmed that they work by removing some nonfunctioning other tests. I also fixed an incidental bug in 'atomic info' that would cause it to return an error code even on success. Changed sector size for correctness and (re)blacklisted the 'rw' argument.

rhatdan:
LGTM

PR #63, opened 2015-06-01
pulp: Hack around python-requests/urllib3 being different in Fedora/RHEL

cgwalters:
Fixes a crash when `atomic` is deployed with current RHEL.


rhatdan:
LGTM

chuanchang:
@cgwalters, Great! it works well for me, thanks a lot :)

PR #67, opened 2015-06-17
atomic mount --live (and other improvements)

willmtemple:
Changes:
- Added --live option to mount running containers. Default behavior is to mount *snapshots*, rather than live.
- Refactored to avoid creating new thin devices. Now we just use docker's or create identical ones.
- By default image-mounts create a container tagged with a sentinel environment variable.
- Non-live container mounts commit the container to an image (tagged with a sentinel label) then image-mount it. On cleanup, the container mounted at the mount path and its image are inspected for sentinel values and removed if they are tagged.
- a --no-bind option was added to prevent atomic mount from bind-mounting 'mntpoint/rootfs' to 'mntpoint'. The default behavior is to perform the bind-mount.
- More easily support extension to other backends. e.g. to support the btrfs backend, one only needs to write `DockerMount._mount_btrfs()`, and the mount command will automatically invoke that method when on the btrfs backend. At the moment, only LVM devicemapper is supported.
- If a --live mount is specified, setting any mount option is illegal.
- Setting the 'context' mount option is forbidden. Security context is provided by the container.

**NOTE**: strongly recommended to add '--storage-opt dm.use_deferred_removal=true' to docker daemon command line. It sounds like this setting will become a default in at least the Red Hat distributions if not upstream.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -9,6 +9,7 @@ import requests
 import pipes
 import mount
 import selinux
+import util
 try:
     from subprocess import DEVNULL # pylint: disable=no-name-in-module
 except ImportError:
@@ -184,15 +185,6 @@ class Atomic(object):
             self.name = self.image.split("/")[-1].split(":")[0]
             if self.spc:
                 self.name = self.name + "-spc"
-
-    def _get_selinux_label(self):
-        if selinux.is_selinux_enabled() != 0:
-            fd = open(selinux.selinux_lxc_contexts_path())
-            for i in fd.readlines():
-                name,context = i.split("=")
-                if name.strip() == "file":
-                    return context.strip("\n\" ")
-        return ""
 
     def _getconfig(self, key, default=None):
         assert self.inspect is not None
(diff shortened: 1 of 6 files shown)


rhatdan:
+                if o.startswith('context='): 

Would be better to have 

if o.find(context=) != -1

rhatdan:
As in other comments, Change DockerMount to just Mount

rhatdan:
Otherwise it looks great.

willmtemple:
@rhatdan Added a commit with the following changes:
- Restructured mount module such that DockerMount is a subclass of a larger Mount class, which contains the static methods which might be common among different container platforms.
- Moved mount-related utility code into Mount class as static methods.
- Made the app smarter about merging default options 'ro,nodev,nosuid' with user options.
- SELinux context is specifiable for non-live containers. If no context is specified, it will pick the host's default container type and mount the device with that context.
- Man pages changed to reflect above.

rhatdan:
Can you just squash the requests and then we can merge.

willmtemple:
@rhatdan Done.

rhatdan:
LGTM lets merge for docker-1.7 release.

PR #66, opened 2015-06-17
Add IMAGE and NAME as environment variables

DirectXMan12:
This commit adds `IMAGE` and `NAME` as environment variables
passed to install, uninstall, stop, and run (similarly to
`CONFDIR`, `DATADIR`, and `LOGDIR`).  This allows users to write
`NAME=${NAME}` instead of `NAME=NAME`, and also enables `${NAME}`
and `${IMAGE}` to be used in arbitrary parts of the commands.

This commit is based off of ideas from #59

PR #69, opened 2015-06-18
Add SUDO_UID and SUDO_GID variable substitions

bexelbie:
These substitutions are useful with docker -u SUDO_UID:SUDO_GID ...
for userspace tools such as format translators

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -10,6 +10,7 @@ import pipes
 import mount
 import selinux
 import util
+import pwd
 try:
     from subprocess import DEVNULL # pylint: disable=no-name-in-module
 except ImportError:
@@ -411,10 +412,29 @@ removes all containers based on an image.
 
     @property
     def cmd_env(self):
-        return {'NAME': self.name, 'IMAGE': self.image,
-                "CONFDIR": "/etc/%s" % self.name,
-                "LOGDIR": "/var/log/%s" % self.name,
-                "DATADIR":"/var/lib/%s" % self.name}
+        env = {'NAME': self.name,
+               'IMAGE': self.image,
+               'CONFDIR': "/etc/%s" % self.name,
+               'LOGDIR': "/var/log/%s" % self.name,
+               'DATADIR':"/var/lib/%s" % self.name}
+
+        with open("/proc/self/loginuid") as f:
+            default_uid = f.readline()
+
+        if "SUDO_UID" in os.environ:
+            env["SUDO_UID"] = os.environ["SUDO_UID"]
+        else:
+            env["SUDO_UID"] = default_uid
+
+        if 'SUDO_GID' in os.environ:
+            env['SUDO_GID'] = os.environ['SUDO_GID']
+        else:
+            try:
+                env['SUDO_GID'] = pwd.getpwuid(int(env["SUDO_UID"]))[3]
+            except:
+                env["SUDO_GID"] = self.default_uid
+
+        return env
 
     def gen_cmd(self,cargs):
         args = []
--- a/README.md
+++ b/README.md
@@ -77,6 +77,14 @@ directories if necessary.
 In my example the INSTALL method will execute the install.sh which we add to
 the image.  The root sub directory contains the following scripts:
 
+The `atomic install` will set the following environment variables for use in the command:
+
+**SUDO_UID**
+  The `SUDO_UID` environment variable.  This is useful with the docker `-u` option for user space tools.  If the environment variable is not available, the value of `/proc/self/loginuid` is used.
+
+**SUDO_GID**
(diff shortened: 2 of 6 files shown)


rhatdan:
How about we just add

-u and -g 

And substitute ${UID} and ${GID}

bexelbie:
I am not sure I follow you here Dan?  Are you suggesting two new options to `docker run` ?

My goal was to find out the UID and GID of whomever typed `sudo atomic run foo` and use that in the resulting docker command.  I am cool with using a process similar to pull #66 for passing the values around if we want.

rhatdan:
Ok that is a tougher problem.

We have loginuid which we could get this infor from but not something from GID>

bexelbie:
What is the source of loginuid?  I am not familiar with it.  The SUDO_UID and SUDO_GID are generally available, I thought.

rhatdan:
Funny I never new about SUDO_UID and SUDO_GID, but that makes it easy to pass them along.  One problem is they would not be guaranteed, since the user could have come from su.

loginuid is set at login time and is used for audit logging, it is not supposed to be overridable.

cat /proc/self/loginuid.

But We could pass the SUDO_UID and SUDO_GID into the atomic command, I am wondering if it would make most sense to just pass the users environment to the commands.

rhatdan:
@DirectXMan12

bexelbie:
My concern with absorbing the whole environment was name collisions.  We already seem to have enough discussion over substitution methodology that I didn't want to add collisions.

If we can structure the replacement code right, even lower-level coders can submit a PR to get a new variable in.

A non-colliding alternative might be to pull specific environment variables when wanted and to get all variables meeting the pattern ATOMIC_*

DirectXMan12:
I like the idea of taking anything `ATOMIC_xyz` and importing it automatically.  FWIW, importing the whole environment makes me a bit uneasy -- pulling in specific variables (or variables with a specific prefix) seems like it would probably cause fewer headaches in the future.

Doing something similar to #66 should be fairly easy (esp since after #66, you just have to add a new element to the results of the `cmd_env` property), and sidesteps any issues about how to properly do replacement.

rhatdan:
@bexelbie Since we merged https://github.com/projectatomic/atomic/pull/66, you would need to change to use this method, then I guess we could merge.

bexelbie:
@rhatdan I have made modifications.  I am open to further guidance.  I'll try to handle the ATOMIC_ variables in a separate PR

DirectXMan12:
Additionally, the docs should indicate that the variables are being passed as environment variables (like `CONFDIR`, etc), not as text substitutions (like `NAME=NAME`).

bexelbie:
@rhatdan is there a better way to read /proc/self/loginuid than I did?
@DirectXMan12 good catch - I hope to push a doc fix through after this is accepted that cleans up some other issues as well.  I am holding off because I reformatted the variable substitutions so I don't want to have to rebase.

rhatdan:
@bexelbie Looks fine other then you should call f.close() and fix the documentation.

bexelbie:
@rhatdan an explicit f.close() has been added.  I had read that using a `with open ...` structure automatically closed the file handle, so I didn't know whether it was idiomatic to call it explicitly or not.

Also the docs are correct for things this patch modifies.  Do you want me to roll a larger docs update into this patch as well to clean up old `NAME=NAME` references, fix backslash requirements and rewrite all substitutions as environment variables?  I was planning to do that separately for cleanliness, but am happy to roll it in.

rhatdan:
Ok, maybe the f.close() was not necessary.  But I don't think it will hurt.  

I am talking about the references to 1001 for UID, you should probably mention login_uid.

Changes to the other substitution stuff would be appreciated in a separate pull request.

bexelbie:
@rhatdan thanks for pointing out the 1001s.  I completely read over them.

DirectXMan12:
@rhatdan, @bexelbie : 
For future reference, `__exit__` (the method called when leaving the `with` block) calls `close`
(you can see what's happening here: https://hg.python.org/cpython/file/15c95b7d81dc/Objects/fileobject.c#l2015)

bexelbie:
@rhatdan I forgot to explicitly say it, the docs are fixed.  @DirectXMan12 would you suggest deleting the f.close() or is it idiomatically better to leave it?

DirectXMan12:
@bexelbie : idiomatically, it's better to remove the explicit `f.close()` (from a idiomatic perspective, it looks odd).

bexelbie:
I made one tweak to add an attempt to look up the default GID of the SUDO_UID value we use, if SUDO_GID isn't present.  @rhatdan do you want the f.close removed?

rhatdan:
Yes remove the f.close

bexelbie:
@rhatdan done.  Thank you.

PR #71, opened 2015-06-22
Bugfix atomic CLI for users without GraphDriver API exposure, simpler mount options

willmtemple:
There is a bug in the atomic cli which prevents users who do not have the GraphDriver API exposure patch from mounting containers. This patch fixes that issue.

In addition, it also fixes a but in mount options in such a way that default options will be honored only if no mount options are provided. The old logic was cumbersome, and would have been prohibitive of any other Graphdriver backend having its own set of default options. The man pages have been updated to reflect the simpler behavior.


rhatdan:
LGTM

PR #74, opened 2015-06-26
Updates the default for `atomic run --spc` to use `docker create`

bexelbie:
This makes `atomic run foo` and `atomic run --spc foo` behave
the same way


rhatdan:
LGTM, although I still think we need to easily change the default behaviour to differentiate use cases.  


atomic run rhel-tools 

Is one use case where people see the container as a "admin shell", such that updates to the shell persist.  

The second use case is atomic run SERVICE, where users often think the run creates a container for the single run, and then the container is destroyed,  IE Each atomic run SERVICE, is independent of the next atomic run SERVICE.

bexelbie:
@rhatdan does #75 provide a forum for that discussion?

rhatdan:
Yes

bexelbie:
@rhatdan if you have time to look at this for feedback, I would be appreciative.  I have written this to meet the ideas expressed in #75 pending sctweedie's response.  It definitely needs docs before it can be merged and probably some tests in the new test framework.

rhatdan:
@bexelbie This one has fallen through the cracks.  Is this still something you want?

bexelbie:
I'd like to still see this considered, however, I haven't looked at how the codebase has changed in a while.  Does it make sense for atomic still?

PR #76, opened 2015-06-30
use NAME and IMAGE variables, not literals in README.md

ibotty:
I assume that is a typo and not some magic that escapes me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,8 @@ RUN yum -y update; yum -y install httpd; yum clean all
 
 LABEL Vendor="Red Hat" License=GPLv2
 LABEL Version=1.0
-LABEL INSTALL="docker run --rm --privileged -v /:/host -e HOST=/host -e LOGDIR=${LOGDIR} -e CONFDIR=${CONFDIR} -e DATADIR=${DATADIR} -e IMAGE=IMAGE -e NAME=NAME IMAGE /bin/install.sh"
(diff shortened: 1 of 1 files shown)


rhatdan:
Why didn't you add IMAGE=${IMAGE}

Also don't we need to escape all of these.

IMAGE=\${IMAGE}

ibotty:
> Why didn't you add IMAGE=${IMAGE}

because I missed it. :-|

If we need to escape them, don't we need to escape all environment variables in LABELs? I don't really understand how substitution in LABELs is supposed to work. Is `NAME=NAME` supposed to be `NAME=my_image_name` when used or `my_image_name=my_image_name`. If the latter, how can one supply a `NAME=my_image_name`?

rhatdan:
NAME=my_image_name

ibotty:
I am not entirely sure my patch is correct. testing it, it substitutes NAME=NAME (i.e. without any `$`) and IMAGE=IMAGE correctly, so that part is at least unnecessary. I don't get any expanded $LOG_DIR regardless of escaping it though. (using atomic-0-0.7.gita7ff4cb.fc22.x86_64).

rhatdan:
Yes we will continue to support NAME=NAME and IMAGE=IMAGE ,but I think we should move docs to talk about ${IMAGE}.

$LOG_DIR  should be $LOGDIR

PR #78, opened 2015-07-03
Make atomic code python3 compatible. Fixes #77

### LINKED ISSUE #77: python3 support for atomic
Looking at atomic code it looks like that it is not python3 ready - `import ConfigParser` at  atomic/Atomic/config.py@2 , do you have any plans to fix Python3 incompatibilities (if there is more of them)? Or would you accept PR regarding this issue?

bkabrda:


rhatdan:
LGTM

@cgwalters @baude @willmtemple

cgwalters:
Looks fine to me - if there was more than this I'd say we should probably be looking at python-six, but we can go farther without it.

bkabrda:
Thanks for accepting this.
As for `from __future__ import print_function`, I think that'd be a reasonable thing to do. I totally agree that if py3 compat gets more complex, six should be added as dependency. The reason why I didn't do that right away is that I didn't want to introduce a new dependency.

rhatdan:
BTW, we just merged a change that should allow us to support atomic on Python3 in Rawhide going forward.

PR #79, opened 2015-07-06
Push-to-Satellite feature

jramseye:
Started adding support for satellite in addition to pulp. Need to test it and integrate it into the rest of the system. Waiting on Satellite server to test with. Jenny Ramseyer

Added a push-to-satellite function, might have gotten the config stuff to work.  Jenny Ramseyer

refactoring and adding documentation

rechecked API calls.  More documentation

it builds.  Still not sure the config file parser is going to work--need to check section headings, but it builds.  Jenny Ramseyer

switched false to true in config.py

debugging the REST API calls

more bug fixing. May have fixed payload, need to check.  Have to decide what to do about publish

Fixed.  It should work now.  Need to fiddle around with create_repo more, but the rest should be good.  Thanks to David Davis for help with uploading an image.

debugging create_repo

fixing build error.

fixing some build errors.  Still can't push to Satellite or Pulp.  Not sure why, working on that.  Jenny Ramseyer

more bug fixing, cleaning up CLI

minor tweaks

fixed the push to pulp/satellite problem!  Thanks to Alec and Will for the help!  Now new error

Seems to be working now

Hooray it works.  Now we're getting connection errors, which I will try to debug. But I think it works.  Might have messed up an API call, which is why it won't connect?  Or maybe it is a server problem?  I will investigate.  --Jenny Ramseyer

committing before merge

Sending in this as a pull request.

satellite and pulp are now mutually exclusive arguments

Fixed the authentication problems in the code.

Debugging upload calls


rhatdan:
Delete the commented out session.

jramseye:
Will do!  

After talking more with Aaron, I will need to restructure this code--we have decided to switch to using the activation key given by Satellite, instead of the user credentials.  So that should require some code restructuring.  Once I get that working, I will update this request again.

jramseye:
Hi!  I got it to work now.  It successfully uploads images.  Still some minor work to do on fixing the tags and the user's config file, but it works!

rhatdan:
@cgwalters @willmtemple @bexelbie @baude Could one of you guys review the code?

rhatdan:
@jramseye I thought I stated this but could you move the push_to_pulp and push_to_satellite to pulp.py and satellite.py respectively.  I think that makes more sense.

jramseye:
Good point.  I've moved them now.

jramseye:
Some of the authentication to the server will probably change--I am talking to the Satellite team next week about how best to authenticate the user to the server.

rhatdan:
@jramseye Is this ready to merge?

jramseye:
@rhatdan Yes it can be merged.  
Tomorrow I am talking to the Satellite team to see if we can figure out a better way for the user to tell the server where to put the image (currently it takes an activation key and a repo id, which is only visible in the URL of the page.  It works fine, but is a little hard for the user to find).  So it may change based on that.  But I can open another PR instead once we change that, if we decide to change it.

rhatdan:
@jramseye Please rebase

jramseye:
@rhatdan fixed!

rhatdan:
LGTM

jramseye:
Thanks! I'll make a new pull request if we get the subscription-manager version working.

jramseye:
I caught a bug in satellite.py where there was extra whitespace in one of the upload calls (introduced when I made it pep8 compatible).  I've fixed the whitespace bug, see PR: https://github.com/projectatomic/atomic/pull/127
With the bug, it won't upload your image and will silently fail, so the PR will fix that.

PR #81, opened 2015-07-07
host: Proxy the `rpm-ostree rebase` operation too

cgwalters:
See https://lists.projectatomic.io/projectatomic-archives/atomic/2015-June/msg00000.html
for an example use case.


rhatdan:
LGTM

PR #83, opened 2015-07-08
Unit/Integration Testing Framework

willmtemple:
### Adds a testing framework and some starter tests.

The unit tests are python-unittest files located in `tests/unit/`. Integration tests are shell scripts located in `tests/integration`. If an integration test produces nonzero exit status, it is considered to have failed.

Example files are provided in `tests/`.

New integration tests must be named `test_*.sh` and will be automatically discovered.

The test harness prepares some test images automatically. By default, it pulls a busybox and centos image, as well as building any file matching `Dockerfile.*` located in `tests/test-images/`.

Currently, the tests must be run as root, otherwise some integration tests which depend on mount functionality will not pass. In order to invoke the tests, run `sudo test.sh`.

The harness will generate a coverage report and write it to stdout. As such, python-coverage is now a dependency for testing the package. For a more detailed report, run `coverage html` in the directory where `test.sh` is located and then browse to `file:///path/to/atomic/htmlcov/index.html`.

The old test script is now `test-old.sh`.

###### TL;DR: The basic idea is that you make modifications to the stub Dockerfiles in `tests/test-images/` and then use those images as your test data.

@rhatdan RFC. I've tried to build this to be as extensible as possible, so if something isn't right let me know and I can add/remove features to get this where we want it.

Also a lot of this is pretty hackish so if anyone has any comments on how to clean it up, please leave feedback!


rhatdan:
if not self.args.no_bind:
+            driver = self.d.info()['Driver']
+            if not self.args.no_bind and driver == 'devicemapper':

Could we be clever about this and force no_bind when driver==devicemapper?

rhatdan:
Why are we saving  test-old.sh?

rhatdan:
Other then that this looks good.

willmtemple:
@rhatdan `--no-bind` only has an effect when on devicemapper. It isn't needed on overlay/btrfs. I suppose we could be clever and force `no-bind=False` when driver != devicemapper.

rhatdan:
That is what I was thinking, to simplify the code.

willmtemple:
@rhatdan Updated with changes.

rhatdan:
LGTM

PR #84, opened 2015-07-09
BindTo docker in example systemd service

ibotty:
BindTo=docker.service ensures, that the service is stopped or restarted
whenever docker is. This is more sane behavior if no docker restart policy is
given (i.e. --restart=always).


rhatdan:
@lsm5 what do you think?

It looks good to me. Until we can break containers away from the docker daemon.

lsm5:
LGTM

PR #85, opened 2015-07-09
OverlayFS DockerMount backend for atomic mount command.

willmtemple:
Depends on/contains #83, wouldn't be *too* much trouble to isolate this if needed.

DockerMount backend for the overlay docker storage driver.

By observation, this backend is remarkably faster than devicemapper, but has some restrictions due to  technical limitations.

- Mounts are read-only. Any attempt to mount a container writeable will cause the program to terminate due to:
- Live mount cannot be supported on OverlayFS at this time. Mounting a writeable overlay filesystem while another writeable overlay is running produces undefined behavior. The only way to view the container live is to `nsenter` the mount namespace of the docker daemon and navigate to `/var/lib/docker/overlay/<uuid>/merged`, where the overlay is already mounted. (However, it seems we should be able to mount and modify a non-running container as long as we can ensure that docker wouldn't try to start the container, but afaik we can't do that right now).


rhatdan:
@willmtemple I think this needs rebase?

rhatdan:
LGTM

PR #86, opened 2015-07-10
Rename atomic_server.py to atomic_dbus.py

rhatdan:
Also cleanup setup.py to handle most of the setup.
setup handles python3 issues better then standard Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,6 @@ PREFIX ?= $(DESTDIR)/usr
 SYSCONFDIR ?= $(DESTDIR)/etc/sysconfig
 PROFILEDIR ?= $(DESTDIR)/etc/profile.d
 PYTHON ?= /usr/bin/python
-BASHCOMPLETIONDIR ?= $(PREFIX)/share/bash-completion/completions/
 
 all: python-build docs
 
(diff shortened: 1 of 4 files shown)


rhatdan:
@cgwalters Could you take a look at this,  I was playing with getting atomic_dbus.py interface, and making setup.py do what I want.  Tricky getting py files into /usr/share/atomic/

cgwalters:
I'm not a setuptools expert, looks OK to me though.  I tested it locally via `make && make install DESTDIR=$(mktemp -d)` as non-root, and needed:

```
diff --git a/Makefile b/Makefile
index 8fab102..466b991 100644
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ clean:
 
 install: all 
        $(PYTHON) setup.py install --install-scripts /usr/share/atomic `test -n "$(DESTDIR)" && echo --root $(DESTDIR)`
+       install -d -m 0755 $(DESTDIR)/usr/bin
        ln -fs ../share/atomic/atomic $(DESTDIR)/usr/bin/atomic
 
        [ -d $(SYSCONFDIR) ] || mkdir -p $(SYSCONFDIR)
```

rhatdan:
Added your install line and I am going to merge.

PR #91, opened 2015-07-16
atomic.sh: fix up command forwarding to rhel-tools

jlebon:
Small tweaks to make command forwarding work properly.

### DIFF
--- a/atomic.sh
+++ b/atomic.sh
@@ -19,6 +19,7 @@ command_not_found_handle () {
 
 	# run the command, or just print a warning
 	if [ $runcnf -eq 1 ]; then
+		echo "Redirecting to container ${TOOLSNAME}..."
 		atomic run ${TOOLSIMG} "$@"
 		retval=$?
 	else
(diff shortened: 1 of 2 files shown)


bexelbie:
Should this be modified in the upstream to check for a TOOLSIMG and only call it if defined and default to undefined?  Then in the rhel version add the name?

rhatdan:
LGTM

rhatdan:
@bexelbie Not sure what you mean, This is defaulted to not execute the atomic command.  If you uncomment the sysconfig line, it will run.

jlebon:
@rhatdan Thanks! I think @bexelbie means that TOOLSIMG shouldn't be defined on Fedora (is that correct @bexelbie?).

bexelbie:
@jlebon yes, that is correct.  Therefore I would expect that this repo wouldn't contain an entry there.  Each distro should modify it before packaging as needed.  However, I may be mistaken.

jlebon:
@bexelbie Agreed. I'll tweak it. (BTW, I looked but it doesn't seem like there's an official equivalent to rhel-tools for Fedora, is there?)

PR #93, opened 2015-07-16
WIP: Add --display to run|install

sallyom:
Signed-off-by: Sally O'Malley <somalley@redhat.com>

### DIFF
--- a/tests/test-images/Dockerfile.1
+++ b/tests/test-images/Dockerfile.1
@@ -1,4 +1,9 @@
-FROM scratch
-MAINTAINER "William Temple <wtemple at redhat dot com>
+FROM centos
+MAINTAINER "Sally O'Malley <somalley at redhat dot com>
+ENV container docker
 
 LABEL "Name"="atomic-test-1"
+
+LABEL RUN "/usr/bin/docker run -t --user \${SUDO_UID}:\${SUDO_GID} \${OPT1}  -v \${LOGDIR}:/var/log -v \${DATADIR}:/var/lib \$OPT2 --name \${NAME} \${IMAGE} \$OPT3 echo I am the run label." 
+
+LABEL INSTALL "docker \${OPT1} run  -v \${CONFDIR}:/etc/${NAME} -v \${LOGDIR}:/var/log/${NAME} -v \${DATADIR}:/var/lib/${NAME} \$OPT2 --name \${NAME} \${IMAGE} \$OPT3 echo I am the install label."
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -229,13 +229,20 @@ class Atomic(object):
                 cmd += self.command
             else:
                 cmd += self._get_cmd()
-            return subprocess.check_call(cmd, stderr=DEVNULL)
+            if self.args.display:
+              return self.display(cmd)
+            else:
+                return subprocess.check_call(cmd, stderr=DEVNULL)
         else:
             if self.command:
-                return subprocess.check_call(["/usr/bin/docker", "exec", "-t", "-i", self.name] + self.command,
-                                             stderr=DEVNULL)
+                if self.args.display:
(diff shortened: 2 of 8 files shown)


sallyom:
list of test cmds i've run with output: http://fpaste.org/246219/42462114/

rhatdan:
Why do you need to check self.args.display all over the place?

rhatdan:
This should only be for run and install.  Remove all other checks.

sallyom:
there were a few extra checks in there :)

sallyom:
note WIP, I have the integration test running i'm still working on it, for example, when I change the LABEL RUN in the Dockerfile and leave the expected outcome the same in test_display.sh I expect the test to fail but it passes. @willmtemple please review if you have time, thanks :)

rhatdan:
Need updated man pages.

sallyom:
@willmtemple , I implemented your fb with the tests, thanks again.

sallyom:
added def display(self, cmd)

sallyom:
implemented fb, updated man pages

rhatdan:
LGTM.

@willmtemple @cgwalters 
Could you guys take a look.

willmtemple:
@rhatdan @sallyom I'm not sure we should include an image that might actually manipulate the host as a test. This sounds like in order to test `atomic run` on this image, I would have to actually start an httpd on my system. Could we construct a toy image instead that would test the core functionality in the same way?

rhatdan:
@willmtemple Good catch.  

Lets change the Dockerfile to only have the labels, to not do anything other then echo something.

LABEL INSTALL "docker run ...\${image} echo I am the install label"
LABEL RUN "docker run ...\${image} echo I am the run label"
LABEL UNINSTALL "docker run ...\${image} echo I am the uninstall label"

willmtemple:
@rhatdan @sallyom That sounds good. We should try to work the --opt<n> arguments, and any settable environment variables into those labels.

sallyom:
ok, will do

sallyom:
Done, let me know if i covered everything with the LABEL RUN, LABEL INSTALL

rhatdan:
I think you should put legitimate commands into your Dockerfile, these commands should cause the tests to blow up.  You should also check with out the --display to make sure --display did not break anything.

tests/test-images/Dockerfile.1
```diff
@@ -1,4 +1,11 @@
...
+FROM centos
+MAINTAINER "Sally O'Malley <somalley at redhat dot com>
+ENV container docker
 
 LABEL "Name"="atomic-test-1"
+
+LABEL INSTALL "docker run --rm --spc \$SUDO_UID:\$SUDO_GID --privileged \$OPT1 FOO1 \$OPT2 FOO2 -v /:/host -e HOST=/host -e LOGDIR=${LOGDIR} -e CONFDIR=${CONFDIR} -e \${DATADIR} --name NAME IMAGE  echo I am the install label /bin/install.sh"
+LABEL RUN "docker run -t -i --rm --spc \$SUDO_UID:\$SUDO_GID \$OPT1 FOOBAR1 \$OPT2 FOOBAR2 \$OPT3 FOOBAR3 -v LOGDIR:/var/log -v DATADIR:/var/lib --name NAME IMAGE echo I am the run label run.sh"
+
+EXPOSE 80
```
willmtemple: I would drop this line, as well as the --spc and --privileged options in the run and install labels.

sallyom:
@willmtemple carriage returns (^M) were messing up my test output, finally got past it.
@rhatdan I added the run,install tests and checked against output of 'docker logs container'

willmtemple:
@sallyom @rhatdan If we get rid of privileged options (`--cap-add=SYS_ADMIN`, `--net=host`, `--ipc=host`) in the dockerfile, then it LGTM. All tests are passing on my end.

rhatdan:
Yes remove them, and then we can merge.

sallyom:
OK, done

rhatdan:
LGTM

PR #94, opened 2015-07-17
atomic.sysconfig: leave default TOOLSIMG undefined

jlebon:
We leave it up to each distro to configure the TOOLSIMG to be used.

Related PR: https://github.com/projectatomic/atomic/pull/91


rhatdan:
We probably want to make this distribution specific. So for the RHEL packaging we change this back.

PR #95, opened 2015-07-20
Bugfix for docker create_container API change.

willmtemple:
In docker API version 1.19, mem_limit was moved to the host_config section of the container metadata. This changed the docker-py api and attempts to mount an image or non-live container would result in the following error:

`mem_limit has been moved to host_config in API version 1.19`

This patch addresses the problem by removing the memory limit, which was probably unnecessary in the first place, as these containers are only temporary for the purposes of mounting the internal filesystems.


rhatdan:
Have yo you submitted this upstream?

rhatdan:
Oops never mind previous comment.

PR #96, opened 2015-07-20
Integrate remote inspection with Atomic info.

willmtemple:
Rather than pull a repository when it isn't found locally, with this patch the `atomic info` command will now use remote inspection. It will first attempt to find the image locally, and then will query configured registries.

Additionally, a `--remote` flag will force `atomic info` to use only remote repositories.

While writing tests for this feature, I also changed the invocation of `atomic` in the test harness. As a result, the test harness is now capable of including the CLI integration tests in the coverage report.

Tested against `rhatdan/docker:fedora-1.7`

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -482,20 +482,45 @@ removes all containers based on an image.
         return " ".join(args)
 
     def info(self):
-        self.inspect = self._inspect_image()
-        if not self.inspect:
-            self.update()
-            self.inspect = self._inspect_image()
+        """
+        Retrieve and print all LABEL information for a given image.
+        """
+        def _no_such_image():
+            raise ValueError('Could not find any image matching "{}".'
+                             ''.format(self.args.image))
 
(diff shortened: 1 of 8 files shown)


rhatdan:
Why do I need --remote?

willmtemple:
@rhatdan It allows you to inspect a remote copy of an image you may already have.

I have a RHEL 7.1-6 image tagged as `rhel7:latest`, but that same tag on the registry is actually RHEL 7.1-9.

```
[wtemple@laptop ~]$ atomic info rhel7 > ./rhel7-local
[wtemple@laptop ~]$ atomic info --remote rhel7 > ./rhel7-remote
[wtemple@laptop ~]$ diff rhel7-local rhel7-remote 
6c6
< Release: 6
---
> Release: 9
```

This allows us to force-check the registry even when we have a matching local image.

rhatdan:
Ok.

rhatdan:
LGTM

PR #97, opened 2015-07-21
Allow arbitrary files in test images.

willmtemple:
This patch will allow arbitrary files to be inserted into the test images.

If you have a file hierarchy that looks like this:
- test-images/
    - Dockerfile.\<id\>
    - Dockerfile.\<id\>.d/
        - foo

Then the directory `Dockerfile.<id>.d` will be available to the image at build time, and can be added to the image, e.g. (with an \<id\> of 'foo') as
`Dockerfile.foo`:
```
FROM rhel7:7.1-9
MAINTAINER "William Temple <wtemple at redhat dot com>"

ADD ./Dockerfile.foo.d/some_script.sh /usr/bin/some_script
RUN "chmod +x /usr/bin/some_script"
```

The script will also now rebuild any test-images whose files have changed.

@rhatdan RFC. I think this is the cleanest way to add arbitrary files to these test images.

@sallyom This should enable you to add scripts/etc. to the test images, and you should then be able to take the "docker build" logic out of your test script.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -27,6 +27,17 @@ cleanup () {
 }
 trap cleanup EXIT
 
+_checksum () {
+    if [[ -d "${1}" ]]; then
+        CHK=`find "${1}" -type f -exec sha256sum {} \; | sha256sum`
+        echo "${CHK}"
+        return 0
+    fi
+    if [[ -e "${1}" ]]; then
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

PR #101, opened 2015-07-27
Add version support for atomic command

rhatdan:


rhatdan:
@willmtemple @lsm5 @bexelbie @cgwalters could you take a quick look at this?

@cgwalters Any good idea of how we could have setup.py modify the content that atomic -v shows?

lsm5:
`atomic -v` and `atomic --version` give me `1.1`

rhatdan:
@lsm5 They are supposed to, this is current version of atomic in github.  We will update this for the next release for RHEL

lsm5:
ah yup, i meant lgtm :) ... i'll add this for fedora and centos, sometime after today, rel-eng seems to have some trouble pushing updates :|

PR #104, opened 2015-07-29
Bugfix atomic mount command & incorrect test configuration

willmtemple:
Bugfix for atomic mount not detecting the correct docker thin-pool device prefix and failing to recognize containers mounted as docker containers. Also changed test_info.sh to use CentOS rather than RHEL images, as the test harness was changed to ensure that the test environment contains a CentOS image, but not necessarily a RHEL image.


rhatdan:
LGTM

PR #107, opened 2015-08-04
atomic: fix errors in display option

chuanchang:
atomic install with --display doestn't work, and atomic run still preview the command without --display option, the patch will fix them.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -591,10 +591,13 @@ class Atomic(object):
                 self.inspect = self._inspect_image()
         else:
             args = self._get_args("INSTALL")
-        if args:
-            cmd = self.gen_cmd(args + list(map(pipes.quote, self.args.args)))
-            self.display(cmd)
-            if not self.args.display:
+            if not args:
+                args = self.INSTALL_ARGS + self._get_cmd()
+                cmd = self.gen_cmd(args)
+
+            if self.args.display:
+                self.display(cmd)
+            else:
                 return subprocess.check_call(cmd, env=self.cmd_env, shell=True)
 
     def help(self):


rhatdan:
@chuanchang We have always been displaying the command before executing.  The new code was just to allow a "test" run where it would only display the message but not actually run it.

We could open a discussion on whether we should continue to display the command before executing it.

chuanchang:
@rhatdan, got it, good to know this, thanks.

chuanchang:
@rhatdan, there are two questions are mentioned in this PR, except above one, I think we should fix issue for atomic install with --display doestn't work.

rhatdan:
@chuanchang Could you fix your patch for install --display then and I will merge

@sallyom ^^

sallyom:
thanks for the fix @chuanchang  

**from Options: section of 'man atomic install'**
--display Display the image's install options and environment variables populated into the install command. The install command will not execute if --display is specified. If --display is not specified the install command will execute.

chuanchang:
@rhatdan, @sallyom, done, you're welcome :)

rhatdan:
LGTM

willmtemple:
@rhatdan This has a test failure due to a runtime exception. There's a code path in which `cmd` is never set.

```
++ /usr/bin/coverage run --source=./Atomic/ --branch --append atomic install --display -n TEST2 atomic-test-1
  File "atomic", line 306, in <module>
    sys.exit(args.func())
  File "/home/wtemple/Development/atomic/Atomic/atomic.py", line 599, in install
    self.display(cmd)
UnboundLocalError: local variable 'cmd' referenced before assignment
```

chuanchang:
@willmtemple, oh, I forgot to declare 'cmd' variable before if statement, for the name of the flag, I have no idea, @rhatdan, what do you think? thx

rhatdan:
I think this is the correct way.

https://github.com/projectatomic/atomic/pull/109

PR #108, opened 2015-08-06
Revert "atomic: fix errors in display option"

willmtemple:
PR #107 introduces a change which causes uncaught runtime exceptions. See comment on the Pull Request's thread.


rhatdan:
How about this one?

https://github.com/projectatomic/atomic/pull/109

PR #109, opened 2015-08-07
Fix install code with display option

rhatdan:


willmtemple:
@rhatdan Code LGTM but there are test failures. @sallyom is working on the test.

If we're going to commit forward rather than revert then I'll close the revert PR.

sallyom:
submitted test fix https://github.com/projectatomic/atomic/pull/111

----- Original Message -----

PR #110, opened 2015-08-07
Append command to container run method when container is not running.

rhatdan:
We have two cases when calling atomic run, In one case the container
exists and we will just execute the COMMAND into the container.

If the container is not running, we want to append the COMMAND onto the
container RUN method, to allow more complex interaction.

For example if i had container image TEST that had the run label:

LABEL RUN="docker run -ti \${IMAGE} command"

atomic run TEST -h

Would execute

docker run -ti \${IMAGE} command -h

Displaying help information about the container.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -332,6 +332,7 @@ class Atomic(object):
         return command
 
     def run(self):
+        missing_RUN = False
         self.inspect = self._inspect_container()
 
         if self.inspect:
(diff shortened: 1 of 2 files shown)


rhatdan:
@aweiteka Does this work for you?

@willmtemple @cgwalters @sctweedie Please review.

aweiteka:
@rhatdan that's it. Simple as that. Thanks.

willmtemple:
@rhatdan I checked this out locally for tests, and I'm getting an error in `install`. This is confusing to me because I don't think you touched the install code in this PR, but I do not get the following error on master.

```
++ /usr/bin/coverage run --source=./Atomic/ --branch --append atomic install --display -n TEST2 atomic-test-1
  File "atomic", line 309, in <module>
    sys.exit(args.func())
  File "Atomic/atomic.py", line 597, in install
    self.display(cmd)
UnboundLocalError: local variable 'cmd' referenced before assignment
```

It looks like the error is that this PR doesn't include #109, but GitHub hasn't figured out that this branch is out-of date. I think if you merge changes from master into this PR it should be good.

rhatdan:
@willmtemple Please check it again.  Docker is hosed up on my machine right now, so I can not test.  I don't see this happening with the current code.

willmtemple:
@rhatdan LGTM

PR #113, opened 2015-08-10
Fix setup py (fixes #112)

### LINKED ISSUE #112: Fix setup.py
Issues:
- Importing ``__version__`` from Atomic package causes a race condition, where a new user cannot access ``setup.py`` without pre installing dependencies.
- ``python setup.py develop`` does not work.
- Install Python dependencies on running ``setup.py {develop,install}``
- Fix linting errors

rtnpro:
Fixes #112 
- Access ``setup.py`` without installing dependencies
- Make ``python setup.py`` develop work.
- Install python dependencies from ``requirements.txt`` on running ``setup.py {develop,install}``
- Fix linting errors

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+requests
+setuptools
+docker-py
+websocket-client>=0.11.0
+six>=1.3.0
--- a/setup.py
+++ b/setup.py
@@ -1,14 +1,13 @@
 #!/usr/bin/env python
 
 # Author: Dan Walsh <dwalsh@redhat.com>
-import os
 from setuptools import setup
 import pkg_resources
 
 __version__ = pkg_resources.require('Atomic')[0].version
 
 setup(
-    name = "atomic", scripts=["atomic", "atomic_dbus.py"], version=__version__,
+    name="atomic", scripts=["atomic", "atomic_dbus.py"], version=__version__,
     description="Atomic Management Tool",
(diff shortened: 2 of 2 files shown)


rhatdan:
This looks ok to me.  Although some of the python code looks a little more advanced then I am familiar with.

@cgwalters @willmtemple @baude  How does this look to you?

willmtemple:
@rhatdan It LGTM based on my tests, but I'm not really familiar with the setuptools ecosystem, so I'm a little bit confused.

@rtnpro How can I test to make sure this works as intended?

rtnpro:
@willmtemple Here's a very good comparison between ``setuptools`` and ``distutils``. I am not sure about how to write tests for checking for regression from this migration to ``setuptools``. May be, we can run ``python setup.py install`` in a chroot (or mock) environment and check if it places the proper files in the proper locations. Does this test plan sound good?

willmtemple:
@rtnpro I'm not really concerned about unit or integration tests for this. I just wanted to check and make sure that it does work as intended. I've played around with it since yesterday and figured out how it works.

It's worth noting, however, that `libselinux-python` is a dependency not available via pip, so it has to be installed/linked into the virtual env manually using a `.pth` file or other.

@rhatdan LGTM. This is a Quality of Life improvement for developers used to a more "pythonic" development ecosystem.

PR #117, opened 2015-08-13
build-sys: Avoid self-dependency in setup.py

cgwalters:
The earlier 58a6e0e4eff379578d497d67002cbef1438f8ab2 and later
455fc8499fee1b970eafd7f80942405ef6bef17f both break if the atomic
package is not actually installed on the system - as will be the case
with `mock/pbuilder` style builds.

It looks like OpenStack uses https://pypi.python.org/pypi/pbr which
presumably solves this problem.  I however am not super interested
right now in diving into that...I just want the thing to build.

So let's just hardcode the version number in two places for now.
Anyone who wants to improve this should probably look at PBR or
something other projects do.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,13 @@
 #!/usr/bin/env python
 
 # Author: Dan Walsh <dwalsh@redhat.com>
+import sys
+import os
 from setuptools import setup
 import pkg_resources
 
-__version__ = pkg_resources.require('Atomic')[0].version
+sys.path.insert(0, os.path.abspath(os.getcwd()))
+from Atomic import __version__
 
 with open('requirements.txt') as f:
     requirements = f.read().splitlines()


rhatdan:
@rtnpro thoughts.

rhatdan:
@cgwalters change to use 1.2 and then I will merge for now.

cgwalters:
:arrow_up: reworks the patch to do what Ansible does.

cgwalters:
Until this lands I have frozen my internal CI/CD builds at 918609772bb8538cec923e30e75fa2acfa485d27

setup.py
```diff
@@ -1,10 +1,13 @@
 #!/usr/bin/env python
 
 # Author: Dan Walsh <dwalsh@redhat.com>
+import sys
+import os
 from setuptools import setup
 import pkg_resources
```
rtnpro: @cgwalters Please remove the unused import: ``pkg_resources``

rtnpro:
@rhatdan @cgwalters Looks good to me, apart from the unused import of ``pkg_resources``. Once this is fixed, it is good to merge, :+1:

rhatdan:
I will remove the pkg_resources.

PR #121, opened 2015-08-17
    Atomic/mount.py

baude:
* Add mnt_mkdir keyword arg (bool) to the DockerMount class
          to indicate we want to use the mountpoint as a basedir and
          create temporary dirs there for mounting.  This fits the
          API model so things like the image-scanner and friends
          can leverage this code.

        * Small changes to DockerMount.mount and cleanup to account
          for this change.  Also return the new path when the mount
          function is called.

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -165,11 +165,16 @@ class DockerMount(Mount):
     """
     A class which can be used to mount and unmount docker containers and
     images on a filesystem location.
+
+    mnt_mkdir = Create temporary directories based on the cid at mountpoint
+                for mounting containers
     """
 
-    def __init__(self, mountpoint, live=False):
+    def __init__(self, mountpoint, live=False, mnt_mkdir=False):
         Mount.__init__(self, mountpoint, live)
         self.client = docker.Client()
+        self.mnt_mkdir = mnt_mkdir
+
 
     def _create_temp_container(self, iid):
         """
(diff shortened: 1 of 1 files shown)


rhatdan:
@willmtemple

willmtemple:
@rhatdan No, mountpoint doesn't have to end in a '/', but os.path.join() will handle that.

I have strange test problems that I'm not convinced aren't because I broke my Fedora install. Do you all have passing tests running this pull request on Fedora's docker package? @baude

baude:
Updated changing error names to e

rhatdan:
@baude are the tests passing for you?

baude:
@rhatdan No, but they are also NOT passing in master either.  Lets talk on IRC today and see if we can figure out what is going on.

willmtemple:
@baude @rhatdan Tests are passing on my CentOS 7 server on devicemapper docker 1.6.2.

I will try updating docker to 1.7 on CentOS and report back with results. They are not working on my F22 laptop, but again, I'm unsure as to whether or not that's been caused by me messing with the docker installation.

rhatdan:
@baude What is the state of this now?

baude:
@rhatdan It looks like the test failures were injected prior to this PR.  If you like, I can look at the test failures as part of this PR.  I'll need to touch base with you prior and get some info.  Shall we discuss tomorrow?

rhatdan:
Sure, although if we feel confident on this pull request we could merge and fix test problems separately.

rhatdan:
Since test failures were fixed elsewhere merging.

PR #122, opened 2015-08-20
Removes --name \${NAME} from atomic install and uninstall man pages

aveshagarwal:
Because it is error prone to use it this way. Using the same NAME in the
environment variable -e NAME=\${NAME} and in container name (--name) for
install and uninstall seems incorrect because both of them serve
different purposes. Also if NAME is being used to create a container in
the install script, which happens most of the time, it leads to failure
because the same NAME is used for the ephemeral container to process
LABEL INSTALL and due to that, the actual container can not be created
and it gives following error:
"Error response from daemon: Conflict. The name "etcd1" is
already in use by container b50ea8bf1d40."

Anyway assigning names (using -name NAME) to ephemeral containers during
install and uninstall does not make much sense in my point of view.

### DIFF
--- a/docs/atomic-install.1.md
+++ b/docs/atomic-install.1.md
@@ -20,7 +20,7 @@ IMAGE, if this field does not exist, `atomic install` will install the IMAGE
 
 If the container image has a LABEL INSTALL instruction like the following:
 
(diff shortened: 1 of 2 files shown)


rhatdan:
Makes sense.

PR #125, opened 2015-08-24
Atomic.util.image_by_name only matches some possible matches

ibotty:
e.g. the following image (in docker)

    docker.io/user/image:tag

is not matched by the following queries, because the first `/` might not separate the registry from the user-image but the user from the image. 

 * user/image
 * image

The fix for that case is straightforward, but the catch is, that non-prefixed containers (e.g. docker.io/fedora:latest) show an ambiguity: It could be user `docker.io` or registry `docker.io`.

Non-user supplied queries to `_decompose` might also be incorrectly matched, because (at least on one of my machines) there were images without registry: `user/image:latest` in docker's images output.

The patch fixes it by not decomposing the query (and docker's images), but directly matching four times.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -20,35 +20,26 @@ def image_by_name(img_name):
     """
     Returns a list of image data for images which match img_name.
     """
-    def _decompose(compound_name):
-        """ '[reg/]repo[:tag]' -> (reg, repo, tag) """
-        reg, repo, tag = '', compound_name, ''
-        if '/' in repo:
-            reg, repo = repo.split('/', 1)
-        if ':' in repo:
-            repo, tag = repo.rsplit(':', 1)
-        return reg, repo, tag
+    def _repo_tag_matches(tag, name):
+        """
+        Returns whether a tag matches an image, e.g.
+        "registry/user/image:tag" will be matched by "image", "user/image",
+        "registry/user/image", with or without appended ":tag".
+        """
+        return any([matches(tag, '*/'+name+':*'),
+                    matches(tag, '*/'+name),
+                    matches(tag, name+':*'),
+                    matches(tag, name)])
+
+    def _repo_tags_match(tags, name):
+        return any([_repo_tag_matches(tag, name) for tag in tags])
 
     c = docker.Client()
 
-    i_reg, i_rep, i_tag = _decompose(img_name)
-    # Correct for bash-style matching expressions.
-    if not i_reg:
-        i_reg = '*'
-    if not i_tag:
-        i_tag = '*'
-
     images = c.images(all=False)
     valid_images = []
-    for i in images:
-        for t in i['RepoTags']:
-            reg, rep, tag = _decompose(t)
-            if matches(reg, i_reg) \
-                    and matches(rep, i_rep) \
-                    and matches(tag, i_tag):
-                valid_images.append(i)
-                break
-    return valid_images
+
+    return [i for i in images if _repo_tags_match(i['RepoTags'], img_name)]
 
 
 def subp(cmd):


rhatdan:
LGTM

@baude @cgwalters @aweiteka PTAL

rhatdan:
@bexelbie @baude @cgwalters @aweiteka @miminar PTAL

miminar:
I fear this would produce a lot of ambiguity. `image_by_name` is used inside functions that expect single image to match and will raise otherwise.

And to prevent user confustion, I'd vote to use same image matching mechanisms as Docker itself. Upstream's `docker images` wouldn't match `registry.ltd/user/image` with just `*/image` either.

bexelbie:
I don't think a user/image should match without the query containing "user/image"

Could we improve the repo detection logic by checking for a '.' in the item preceding the slash? It appears that usernames are not allowed to have non-alphanumerics.

rhatdan:
@ibotty Could you address the issues above?

ibotty:
Oh, I missed the last comments. Sorry for not doing anything about it.

I am not entirely sure what should be matched by what. Maybe a single
function won't suffice.

I don't think we should mimick `docker images` behavior. In the list
below, it only matches the last query. That can't be what we want.

So, given an image registry/user/image:tag, which of the following queries
should succeed?

 * image
 * image:tag
 * user/image
 * user/image:tag
 * registry/user/image:tag
 * registry/user/image

if user is indeed [a-zA-Z0-9], and repos always contain a `.`, it should
be easy to do.

I might get to it later that week.

jlebon:
bot, test pull request

jlebon:
bot, retest this please

rhatdan:
@ibotty Any update on this?

ibotty:
I am still not sure about the semantics we want. Can someone please comment on my comment above? projectatomic/atomic wrote:@ibotty Any update on this?

rhatdan:
@miminar @bexelbie PTAL

rhatdan:
@baude @runcom PTAL

rhatdan:
@ibotty Have you thought more on this?  Or just dropped it.

ibotty:
I am happy  to implement any semantics you want to have. Having my comment above in mind, I would say the image should be matched by image/user and everything below. What do you say?

ibotty:
typo: user/image

rhatdan:
Sounds good to me.

rhatdan:
@ibotty Have you had time to work on this?

ibotty:
Sorry, I did not. No ETA, because I can't do it in work time. It's no priority anymore unfortunately.

When is 1.10 due? Maybe I can dedicate some time today in the middle European evening to get it in in time.

ibotty:
Please have a look at it now. The fix was made way easier by @bexelbie's comment above, that registries always contain a '.'.

rhatdan:
LGTM

@baude @runcom PTAL

rhatdan:
@cevitch Can you fix the jenkens or know who can?

rhatdan:
@cgwalters Do you know who can fix the jenkins?

cgwalters:
bot, retest this please

jlebon:
@rhatdan 

There are two unit tests failing:

```
UNIT TESTS:
.........FF
FAIL: test_image_by_name_registry_match (test_util.TestAtomicUtil)
  File "/home/jenkins/workspace/atomic-upstream-test/atomic/tests/unit/test_util.py", line 26, in test_image_by_name_registry_match
    self.assertTrue(len(matches) == 1)
AssertionError: False is not true

FAIL: test_image_by_name_tag_glob (test_util.TestAtomicUtil)
  File "/home/jenkins/workspace/atomic-upstream-test/atomic/tests/unit/test_util.py", line 22, in test_image_by_name_tag_glob
    self.assertTrue(len(matches) == 1)
AssertionError: False is not true

```

I haven't tried reproducing this locally yet, though it does seem strange since the tester does clear out all images before each test.

ibotty:
I am pretty sure it's the leading `/` in the image to be matched. I did not test it with it. One could simply drop a potential leading slash in that case.

ibotty:
I think I fixed the test failure. If someone can retest it please.

Atomic/util.py
```diff
@@ -26,9 +26,12 @@
 
 def _decompose(compound_name):
     """ '[reg/]repo[:tag]' -> (reg, repo, tag) """
-    reg, repo, tag = '', compound_name, ''
+    reg, repo, tag = '', lstrip(compound_name, '/'), ''
```
jlebon: Do you mean `compound_name.lstrip('/')` ?

ibotty: I am spending too much time in functional programming languages. I certainly do.

rhatdan:
@ibotty Rebase please and then we can merge if tests pass.

rhatdan:
@baude could you check if this is something we need at this point, or have we already implemented it?

baude:
@rhatdan i dont have any context on this anymore.

@ibotty any chance you can provide some examples I could use to test master and then your code?  examples could just be images names, etc

ibotty:
@baude I lost context as well. But afair, just e.g. docker.io/fedora/apache:latest with queries fedora/apache should exhibit the problem/the fix.

rhatdan:
I am closing this since I think it no longer applies.

Issue #124, opened 2015-08-24
atomic mount fails consistently


ibotty:
using atomic-1.1-1.git5f631c8.fc22.x86_64 but the changes since that revision seem insignificant re that issue. 

```
$ atomic mount doesn_t_matter /path
Traceback (most recent call last):
  File "/bin/atomic", line 306, in <module>
    sys.exit(args.func())
  File "/usr/lib/python2.7/site-packages/Atomic/atomic.py", line 705, in mount
    self.args.live).mount(self.args.image, options)
  File "/usr/lib/python2.7/site-packages/Atomic/mount.py", line 268, in mount
    driver_mount_fn(identifier, options)
  File "/usr/lib/python2.7/site-packages/Atomic/mount.py", line 302, in _mount_devicemapper
    cid = self._identifier_as_cid(identifier)
  File "/usr/lib/python2.7/site-packages/Atomic/mount.py", line 214, in _identifier_as_cid
    if (__cname_matches(c, identifier) or
  File "/usr/lib/python2.7/site-packages/Atomic/mount.py", line 209, in __cname_matches
    return any([n for n in container['Names']
TypeError: 'NoneType' object is not iterable
```

The code in question does not check for a (docker's) container to have `None` as `Names` key, which apparently `Dead` container have:

```
>>> [c['Status'] for c in cli.containers(all=True) if c['Names'] is None ]
[u'Dead']
```

The code in question should maybe just skip these containers, i.e. (maybe non-idiomatic python following)

```
diff --git a/Atomic/mount.py b/Atomic/mount.py
index cf0ff07..7f1446e 100644
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -214,7 +214,7 @@ class DockerMount(Mount):
         container and return its uuid.
         """
         def __cname_matches(container, identifier):
-            return any([n for n in container['Names']
+            return any([n for n in (container['Names'] or [])
                         if matches(n, '/' + identifier)])
 
         # Determine if identifier is a container
```

rhatdan:
@baude @willmtemple Does this make sense to you guys?

ibotty:
Am I being unclear? Maybe I explained it confusingly.

rhatdan:
@ibotty Not it is clear, I just want to confirm with @baude and @willmtemple that your fix is good before merging.  You should probably open a pull request.

ibotty:
Already did: issue #126

baude:
[...]

PR #126, opened 2015-08-24
don't fail when dead images are around

ibotty:
fixes issue #124.

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -214,7 +214,7 @@ class DockerMount(Mount):
         container and return its uuid.
         """
         def __cname_matches(container, identifier):
-            return any([n for n in container['Names']
+            return any([n for n in (container['Names'] or [])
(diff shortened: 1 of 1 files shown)


rhatdan:
@baude @willmtemple Please check this fix?

baude:
@rhatdan  @ibotty I am on vacation but can take a look.  Can either of you remind me how to create a dead container so I can replicate?

ibotty:
I can't replicate them. AFAICT you should be able to get them when nsentering a container and then trying to docker rm it. I never created them on purpose, but sometimes they just emerge and don't go back by itself. Most dead containers (but not all of them) were privileged btw.

rhatdan:
LGTM

rhatdan:
You can create a dead image by creating one image, then create a container based on the image.  Update the image.  Now rebuild the container.  I think this leaves around a dangling image.

PR #127, opened 2015-08-24
Fixing a whitespace bug in satellite.py

jramseye:
I caught a bug in satellite.py where there was extra whitespace in one of the upload calls, which broke the call.


rhatdan:
LGTM

jramseye:
Thanks!

PR #130, opened 2015-08-25
Resolve version info from setup.py using pkg_resources.

rtnpro:
This fixes version info resolution during using setup.py with deps installed and version resolution from ``Atomic/__init__.py`` too. This is in response to https://github.com/projectatomic/atomic/issues/112#issuecomment-134492154

### DIFF
--- a/Atomic/__init__.py
+++ b/Atomic/__init__.py
@@ -1,8 +1,9 @@
 import sys
 import os
+from pkg_resources import get_distribution
 from .pulp import PulpServer, PulpConfig
 from .satellite import SatelliteServer, SatelliteConfig
 from .atomic import Atomic
 from .util import writeOut
 
-__version__ = "1.2"
+__version__ = get_distribution('atomic').version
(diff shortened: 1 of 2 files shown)


rtnpro:
After these changes, ``python setup.py develop`` and ``python setup.py install`` ran seamlessly and installed all the needed python dependencies from ``requirements.txt``

rhatdan:
LGTM 
@cgwalters PTAL

rhatdan:
@rtnpro Please update pull request.

rtnpro:
@rhatdan, I rebased branch on top of current master and resolved the conflicts :)

rhatdan:
LGTM

cgwalters:
This looks like it introduces a hard runtime dependency on setuptools which is a bit unfortunate - e.g. on Atomic Host it's the only thing pulling in setuptools.

rhatdan:
Ok we don't want that.

PR #132, opened 2015-08-26
Fixed CLI so that if the user forgets to specify a URL in the config …

jramseye:
…file or via command line, they will be prompted to input one instead of just failing to upload.  Also refactored so code is pep8 compatible

Fixed bash completion for upload as well.


jramseye:
This changes it so that, if uploading to pulp, we are also prompted for the URL if not specified.

jramseye:
Removed the extra whitespace

rhatdan:
LGTM

jramseye:
Thanks!

PR #138, opened 2015-08-28
Fix issue with commands that don't have opt* args. Closes #137

### LINKED ISSUE #137: atomic stop errors out because opt* args are not there for stop
In the `cmd_env(self)` function the call to self.args.opt1 gives an error because stop doesn't have an `opt1` arg. 

```
[root@f22 foo]$ atomic  stop delloapache                                                                                                                                                                                                    
  File "/usr/bin/atomic", line 306, in <module>
    sys.exit(args.func())
  File "/usr/lib/python2.7/site-packages/Atomic/atomic.py", line 401, in stop
    self.display(cmd)
  File "/usr/lib/python2.7/site-packages/Atomic/atomic.py", line 759, in display
    "/usr/bin/echo \"" + cmd + "\"", env=self.cmd_env, shell=True)
  File "/usr/lib/python2.7/site-packages/Atomic/atomic.py", line 472, in cmd_env
    if self.args.opt1:
AttributeError: 'Namespace' object has no attribute 'opt1'
```

dustymabe:


rhatdan:
Not crazy about the way this function looks, but this is definitely a bug and can't think of a better fix right now.

PR #140, opened 2015-08-28
echo fix for debian/ubuntu

charliedrage:
more cross-platform support for debian/ubuntu

errors due to echo being located at `/bin/echo` rather than `/usr/bin/echo` on fedora/rhel


rhatdan:
Could we just do /bin/echo and it will work on all platforms.

Issue #139, opened 2015-08-28
echo errors on atomic run /bin/sh: 1: /usr/bin/echo: not found


charliedrage:
```
/bin/sh: 1: /usr/bin/echo: not found
```

on debian/ubuntu echo is located at /bin/echo

is there any reason that echo is /usr/bin/ech
```python
 def display(self, cmd):
subprocess.check_call(
"/usr/bin/echo \"" + cmd + "\"", env=self.cmd_env, shell=True)
```

rather than
```python
 def display(self, cmd):
subprocess.check_call(
"echo \"" + cmd + "\"", env=self.cmd_env, shell=True)
```

in `atomic.py`?

rhatdan:
We could make it /bin/echo and then it will work on both.  I often worry about non pathed commands being effected by user environment.  Although I am not sure we are consistent in the code.

rhatdan:
Fixed.

PR #143, opened 2015-08-31
make echo cross-distro

charliedrage:
changes echo location from `/usr/bin/echo` to `/bin/echo`

fixes issue #139


rhatdan:
LGTM

PR #144, opened 2015-08-31
FIX test failures

baude:
Fix test failures that have crept into the atomic master.

tests/integration/test_display.sh
    Fix failure related to defined UID/GID's in the test itself.  I now
    use sed to remove the --user UID:GID to make the comparison equal
    and remove potential for dynamic failures.

    I also had to add a conditional if the labels (in python) are of
    type None to prevent a traceback.

tests/integration/test_info.sh
    There seemed to be two related failures in the test case that relate to
    running 'atomic info --remote'.  If that command is run with a docker
    daemon that is not capable of --remote, the test will failure.  I added
    a variable based on the return of that command and put it under the
    set +e section.

    I then added a conditional for that test (based on the return of above),
    to only run the test if the docker daemon is capable of the --remote
    function.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -595,8 +595,9 @@ class Atomic(object):
             # Some images may not have a 'Labels' key.
             raise ValueError('{} has no label information.'
                              ''.format(self.args.image))
-        for label in labels:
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

PR #145, opened 2015-09-01
Atomic: delete duplicate command from atomic help document

chuanchang:
It's introduced by formatting codes then forget to delete old codes,
the atomic help looks like this.

    version             display image 'Name Version Release' label
    version             display image 'Name Version Release' label


rhatdan:
LGTM

PR #148, opened 2015-09-03
Remove automatic setting of CONFDIR, DATADIR and LOGDIR

rhatdan:
Users can do this themselves by using the ${NAME} field.

Hard coding behaviour of these three variables was a mistake, and this
is a lot more flexible.  I could add back the variables and don't document them
if people think someone has used them

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -61,9 +61,9 @@ class Atomic(object):
                     "-v", "${CONFDIR}:/etc/${NAME}",
                     "-v", "${LOGDIR}:/var/log/${NAME}",
                     "-v", "${DATADIR}:/var/lib/${NAME}",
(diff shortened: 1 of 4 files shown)


rhatdan:
@baude @bexelbie @cgwalters What do you think?

bexelbie:
Perhaps the variables and the corresponding volume mounts should be removed and then be added back via something similar to --spc.  I think this is a pattern that may be worth keeping, but not forcing.  Overall though LGTM.

rhatdan:
We actually don't implement much of anything currently other then passing the ENVIRONMENT into the INSTALL command.  I think it would be better to document this idea rather then write it into gode.  I have another issue on atomic, where these fields are causing confusion.

rhatdan:
Replacing with new pull request.

PR #161, opened 2015-09-14
Remove automatic setting of CONFDIR, DATADIR and LOGDIR

rhatdan:
Users can do this themselves by using the ${NAME} field.

Hard coding behaviour of these three variables was a mistake, and this
is a lot more flexible.  I could add back the variables and don't document them
if people think someone has used them


rhatdan:
@bexelbie @vpavlin @cgwalters @baude PTAL

vpavlin:
I never liked these, so I am all for replacing them.

bexelbie:
+1 for removing.  If we want hard coded behavior it should be part of a package of stuff like --spc

rhatdan:
Ok merging.  Hopefully we don't break anything.

PR #164, opened 2015-09-15
Verify inputs are valid

baude:
Most sub-commands in atomic require the user to enter
    some sort of valid input name.  This can typically be:

        * container name or container ID
        * an image ID
        * some portion of the name of the image as derived
          by the repo-tags

    We know have a common function that can verify if the input
    is valid.  It will return the container ID or image ID if
    the image is valid.  This leverages the current
    util.image_by_name function as well as three additional
    functions that help derive container and image IDs.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -97,6 +97,8 @@ class Atomic(object):
         self.inspect = None
         self.force = False
         self._images = []
+        self.containers = False
+        self.images = False
 
     def writeOut(self, output, lf="\n"):
         sys.stdout.flush()
@@ -116,7 +118,7 @@ class Atomic(object):
             image = self.image
             if self.image.find(":") == -1:
                 image += ":latest"
-            for c in self.d.containers(all=True):
+            for c in self.get_containers():
                 if c["Image"] == image:
                     self.d.remove_container(c["Id"], force=True)
 
@@ -433,6 +435,8 @@ class Atomic(object):
         self.ping()
         BUS_NAME = "org.OpenSCAP.daemon"
         OBJECT_PATH = "/OpenSCAP/daemon"
+        INTERFACE = "org.OpenSCAP.daemon.Interface"
+
         if self.args.images:
             scan_list = self._get_all_image_ids()
         elif self.args.containers:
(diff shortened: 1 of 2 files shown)


baude:
One comment on the code at ->  https://github.com/baude/atomic/commit/6f0f4147085f2be7aa211bb3e9e50d6a5a28a648#diff-555a8731c137f41d8ec2a5b76216bdecR926


I assumed that if we received multiple images back from util.image_by_name that we should list those image names in the ValueError we raise.  I believe my code to do so is OK but I am not dead sure it covers all possibilities.  The information is nice to have, but if deemed risky, I would say we omit the names and make the error message more generic.

rhatdan:
Why is is_image and is_container returning anything other then true/false  We are not using the values.

rhatdan:
Should we be storing list of images and containers?  In case is_container/Image is called multiple times?

baude:
@rhatdan  i do use the returns from is_image and is_container.  I use the image or container ids as the inputs for scanning.  This can be seen -> https://github.com/projectatomic/atomic/pull/164/files#diff-555a8731c137f41d8ec2a5b76216bdecR449


As for the storing of images and containers, flat answer is yes.  Throughout the atomic code, there are instances of calls out to get a list of images or containers.  I'm thinking one of the classes should just go ahead and do it upfront so everyone has it.

rhatdan:
@baude Then change the names to get_image and get_container, and throw exception if it is not a container or image.  I don't think we should return two pieces of data

Yes we should just grab the list of images/containers on first use.  Add a  function like

```
def get_containers(self):
      if self.containers:
            return self.containers
     self.containers = self.d.get_containers()
    return self.containers
```

rhatdan:
Then call it everywhere we call self.d.get_containers()

baude:
@rhatdan  Ok updated as discussed.

rhatdan:
Any update on this?

baude:
@rhatdan I dont see those whitespace issues on my side so I am not sure what is going on there.  The code from 7 days ago should have addressed your review comments.  If not, we can huddle tomorrow.

rhatdan:
Ok I will do a different patch to clean whitespace.

LGTM

rhatdan:
White space is now clean.

PR #167, opened 2015-09-23
Add missing ostree options

rhatdan:


rhatdan:
@cgwalters @dustymabe PTAL

Should I also add the --version?

dustymabe:
This may be a bad idea but is it possible to "collect" all options that atomic cli doesn't know about and simply pass them through to rpm-ostree. If it's a bad option rpm-ostree will complain about it. Would this be better than having to maintain the options in two places? 

If this is a good idea then it is also something we could implement down the road and accept this PR the way it is for now.

cgwalters:
I would lean a bit towards maybe only supporting the "common" options like `--allow-downgrade` - people who are using `--sysroot` are doing pretty advanced things.

That said, no objections to the patch as is too.

(Long term, I think we could support scraping the options dynamically, or just use rpm-ostree as a shared library)

rhatdan:
I took a quick look to see if there was a way to handle options that the command did not know about, but I found nothing.

We could add something like example something like this.

atomic host upgrade -- --foobar

And then pass the additional arguments to the command?

This would not show in atomic host upgrade --help 
though.

rhatdan:
Ok I am going to merge this patch.  Should handle both cases.

Issue #128, opened 2015-08-24
atomic uninstall does not run what is in UNINSTALL label


hhorak:
My intention is to install a container from image and then uninstall just that one container (without removing image itself).
```
#> atomic install -n myc1 --opt2='-e MYSQL_USER=myuser -e MYSQL_PASSWORD=mypass -e MYSQL_DATABASE=mydb' 60bfde38982e
#> atomic install -n myc1 60bfde38982e
```
First problem I have is that `-n` is not documented.

Second and more important is that using the above the command specified in `UNINSTALL` label is not run (in this context the systemd service file installed during `atomic install` is not removed).

Maybe I just don't understand how the `atomic uninstall` is intended to work, but I suspect the `return` at https://github.com/projectatomic/atomic/blob/master/Atomic/atomic.py#L482 is causing the issue.

aveshagarwal:
I have seen this issue too and it worked when I rerun the command again.

hhorak:
Reading it again, my example should have included `uninstall` on the second line:
```
#> atomic install -n myc1 ... 60bfde38982e
#> atomic uninstall -n myc1 60bfde38982e
```

hhorak:
Full example available at https://github.com/openshift/mysql/pull/83.

rhatdan:
Does this work correctly with the current atomic command?

hhorak:
No, it is still the same. Either I use it wrong or it just still doesn't work. Looking at the code, the line https://github.com/projectatomic/atomic/blob/master/Atomic/atomic.py#L518 checks `self.name != self.image`, which is my case and then it is obvious it never get's to UNINSTALL label, since there is return on line 523.

rhatdan:
Does https://github.com/projectatomic/atomic/pull/168 fix the problem?

hhorak:
Almost, it runs the uninstall script, but then it tries to remove the image itself, which is weird:
```
#> atomic install -n myc1 --opt2='-e MYSQL_USER=myuser -e MYSQL_PASSWORD=mypass -e MYSQL_DATABASE=mydb' 5a21ccbafb95
[...]

PR #168, opened 2015-09-23
Run UNINSTALL command even if container name is not same as image

rhatdan:


rhatdan:
@hhorak could you try out this patch?

Should fix https://github.com/projectatomic/atomic/issues/128

rhatdan:
@hroncok verified this fixed his issue.  Merging.

hroncok:
not sure I ever know about this. did you mean @hhorak?

rhatdan:
Oops sorry.

PR #172, opened 2015-09-24
Fix default install line to only pass in suggested directories for CO…

rhatdan:
…NFDIR, LOGDIR and DATADIR


rhatdan:
This now matches what is defined in the man pages.

jlebon:
LGTM!

PR #173, opened 2015-09-24
Make test_display.sh pass again

jlebon:
This patch makes test_display.sh pass. They are mostly tweaks to account
for the following:
- Account for commit 7858957, which removes {CONF,LOG,DATA}DIR from the
  target environment.
- Account for commit ae931d0, which removes those default bind mounts.
- Fix a few mismatches between the labels in the Dockerfile vs the
  expected output in the test script.

And with that, all tests (should) pass!


rhatdan:
Thanks, sorry I missed this.  LGTM

PR #175, opened 2015-09-29
Make pylint scan all the python files and fix pylint issues

jlebon:
The `pylint` invocation in the `Makefile` was only scanning the `atomic` file. We now make it scan all the python files. The following commits fix all the new issues that `pylint` detected.

The final commit fixes the way `rpm-ostree` args are passed to avoid a TypeError during runtime.


cgwalters:
Minor: Can you order the pylint scan *after* the fixes to preserve bisectability?

jlebon:
Good point, will do!

jlebon:
Huh, weird. Seems like GH doesn't detect the re-order. Let me close it and re-open it.

PR #178, opened 2015-10-01
Atomic.images(): handle 'localhost:5000/test-image:latest'

jpopelka:
Previously it failed with:
`too many values to unpack (expected 2)`


rhatdan:
LGTM

rhatdan:
@cgwalters @baude @bexelbie PTAL

I don't believe this would break anything.

baude:
Agree

PR #180, opened 2015-10-05
Drop use of setuptools as it injects a runtime dependency

cgwalters:
This is a re-commit of 106878815f5a5dd61b0e527c502fb6eeaf0c9c67
because the later fix in #112 added a *runtime* dependency on
setuptools, which we don't want.

It looks like OpenStack uses https://pypi.python.org/pypi/pbr which
presumably solves this problem.  I however am not super interested
right now in diving into that...I just want the thing to build.

Ansible just inserts the current dir in `sys.path`, which seems like a
simple hack, so let's do that.

Conflicts:
	setup.py

### DIFF
--- a/Atomic/__init__.py
+++ b/Atomic/__init__.py
@@ -1,11 +1,8 @@
 import sys
 import os
-from pkg_resources import get_distribution
 from .pulp import PulpServer, PulpConfig
 from .satellite import SatelliteServer, SatelliteConfig
 from .atomic import Atomic
 from .util import writeOut
 
-#https://bitbucket.org/logilab/pylint/issues/36/
-#pylint: disable=no-member
-__version__ = get_distribution('atomic').version
+__version__ = '1.5'
--- a/setup.py
+++ b/setup.py
@@ -3,18 +3,17 @@
 # Author: Dan Walsh <dwalsh@redhat.com>
 import sys
 import os
-from setuptools import setup
+from distutils.core import setup
 
-with open('requirements.txt') as f:
-    requirements = f.read().splitlines()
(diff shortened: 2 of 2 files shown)


rhatdan:
This is the most hacked up simple fix...  LGTM

@rtnpro what do you think?

cgwalters:
This doesn't work yet because apparently setuptools is generating the `/usr/bin/atomic` entrypoint script with a runtime dep on `pkg_resouces`.

cgwalters:
Ok, the above commit completely expunges setuptools, going back to distutils.

rtnpro:
As there's no cyclic dependency in imports anymore, I am OK with it :+1: 

However, it was handy to have the ``develop`` option of ``setup.py`` provided by ``setuptools``.

rhatdan:
@cgwalters Do you need this rebuilt for next release? 7.1.6?  7.2?

cgwalters:
Okay, so this caused later regressions because switching between setuptools -> distutils apparently breaks RPM upgrades as the `.egg-info` goes from a directory to a file, and RPM doesn't support that.

So I think my vote here is to revert this, and just eat the runtime dependency on setuptools until we figure out how to do  this really right.

rhatdan:
I reverted the change.  So no update needed for 7.2

cgwalters:
http://pkgs.fedoraproject.org/cgit/atomic.git/commit/?id=1e3d3e184d06c14ef230e95bb5085365fdaead34

rhatdan:
@lsm5 Could you add the above patch to atomic.spec?

lsm5:
rhel 7.2 has `.egg-info` as a file, so we might need to change that, not sure when our next window is to make this change.

PR #183, opened 2015-10-07
Allow GO_MD2MAN path to be configurable via env variable

mpreisler:
Also added a touch invocation to allow users to do:
`GO_MD2MAN=/bin/true make`
This skips building the manuals which helps on systems where getting
the go-md2man tool is difficult. This is similar to the trick that can be
used for PYLINT:
`PYLINT=/bin/true GO_MD2MAN=/bin/true make` now works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ SYSCONFDIR ?= $(DESTDIR)/etc/sysconfig
 PROFILEDIR ?= $(DESTDIR)/etc/profile.d
 PYTHON ?= /usr/bin/python
 PYLINT ?= /usr/bin/pylint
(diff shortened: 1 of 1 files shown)


jlebon:
That makes sense. Additionally, maybe we should split the `pylint` call from the `python-build` target and add it to `all` instead. That way, if we want to skip linting and docs building, we can just do `make python-build`, rather than using that `PYLINT=/bin/true GO_MD2MAN=/bin/true` trick.

rhatdan:
LGTM

rhatdan:
I am fine with splitting that up.

PR #184, opened 2015-10-08
Add --only-cache option for atomic scan

baude:
* For customers who cannot or do not want to fetch the latest OVAL
    CVE input files for atomic scan, they can now pass the --only-cache
    option which disables the ability to fetch new.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -437,7 +437,6 @@ class Atomic(object):
         OBJECT_PATH = "/OpenSCAP/daemon"
         INTERFACE = "org.OpenSCAP.daemon.Interface"
         input_resolve = {}
-
         if self.args.images:
             scan_list = self._get_all_image_ids()
         elif self.args.containers:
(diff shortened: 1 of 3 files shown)


mpreisler:
@rhatdan It changes very often. Every day is quite common.

rhatdan:
What changes?  New CVEs show up everyday?  I am just looking at the limited set of data that we will be scanning a container for.  Just because their is a huge set of data that someone might at somepoint scan for, we should not need to update the scanner.  

Can't we just have a human that mitigates updates of the scan image when a new CVE shows up, else the container image gets updated at the 6 week cadence that we are currently shipping with?

baude:
Dan, the files being fetched are for CVEs only.  As Martin said, it does seem to change close to daily.  If we ship the image preloaded with a cached set of CVE files, does that not satisfy the requirement?  I'm thinking in my head here about other things like yum, which fetch new data before running as well.

rhatdan:
We should open this for bigger discussion on aos-devel list. 

yum talks to satellite in the case I am worried about, and satellite does not get updated that often.  Perhaps if this tool is just talking to satellite to gets its CVE data, I would be less concerned.

baude:
I believe that is the eventual plan but it isn't in production yet.

@rhatdan time for an /etc/atomic.conf ?

mpreisler:
@rhatdan Is it acceptable to phone home in the first version and allow to change the CVE OVAL source in future versions? That way we have some time to add support for this to Satellite and fetch from there.

My data about the RHSA OVALs changing so often comes from the all-inclusive huge XML. It is possible that the split XMLs change less often.

baude:
@rhatdan ill switch the option if you feel strongly.  what do you want the option to be called?

--no-cache, --fetch-upstream, --fetch ?

it would also be nice to be able to set options like this in an /etc/atomic/atomic.conf (in configfile or yaml format).  I could make an additional PR for that if you like.

mpreisler:
@baude I think just `--fetch` is fine, if it's well documented.

The oscapd SPC Dockerfiles now `wget` the RHSA OVALs in preparation for this change, see https://github.com/OpenSCAP/openscap-daemon/blob/master/atomic/rhel7_spc/Dockerfile#L40

I have checked the oscap-docker sources in OpenSCAP and we don't need to change them to support custom URLs. Instantiating the getCVEInput class and setting its `url` member is enough. https://github.com/OpenSCAP/openscap/blob/maint-1.2/utils/oscap_docker_python/get_cve_input.py#L40

So the question is, do we need custom URLs in the first version?

rhatdan:
I think this is a product management decision.  @sghosh151

baude:
@rhatdan assuming the checks pass, this should be ready for merge.

rhatdan:
Could you merge the two requests together

PR #187, opened 2015-10-15
Atomic_scan: add option to override fetching of new CVE data

baude:
* The openscap-daemon can now optionally take a switch
      to override the behavior defined in its configuration
      file as to whether to fetch new CVE input data or
      not.

        --fetch_cves is now a bool (True|False) where true
            means to pull new, false means do not attempt
            to pull aka "phone home"

    * Added timeout value of 99999 to override the default dbus
      connection timeout of 25 seconds.  This allows longer
      scans to complete and not throw an Exception.

    * Added a small def to detect a bool value based on True,
      true, yes, y, 1 and so on

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -437,7 +437,6 @@ class Atomic(object):
         OBJECT_PATH = "/OpenSCAP/daemon"
         INTERFACE = "org.OpenSCAP.daemon.Interface"
         input_resolve = {}
-
         if self.args.images:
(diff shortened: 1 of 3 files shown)


baude:
Replaces https://github.com/projectatomic/atomic/pull/184/ and squashed into one commit.

rhatdan:
@baude You did not need to close the old, you could have merged the two together and then done a --force push.

rhatdan:
LGTM

Issue #186, opened 2015-10-15
`atomic unmount path` not working as expected


bexelbie:
`atomic unmount path` complains that the mount point is not a docker container.  It removes one of the two mount entries and leaves what looks like the layered fs references in place.

Fedora 22
atomic-1.1-1.git5f631c8.fc22.x86_64

---

    bexelbie@bexelbie:~$ mkdir bex_mount
    bexelbie@bexelbie:~$ sudo atomic mount mysql bex_mount
    bexelbie@bexelbie:~$ mount | grep bex_mount
    /dev/mapper/docker-253:2-1451009-444fb76932aeb58f9f47946c1c380364353db3a000b9aea0322b19ae70812c08 on /home/bexelbie/bex_mount type xfs (ro,nosuid,nodev,relatime,context="system_u:object_r:svirt_sandbox_file_t:s0:c4,c384",attr2,inode64,sunit=1024,swidth=1024,noquota)
    /dev/mapper/docker-253:2-1451009-444fb76932aeb58f9f47946c1c380364353db3a000b9aea0322b19ae70812c08 on /home/bexelbie/bex_mount type xfs (ro,nosuid,nodev,relatime,context="system_u:object_r:svirt_sandbox_file_t:s0:c4,c384",attr2,inode64,sunit=1024,swidth=1024,noquota)
    bexelbie@bexelbie:~$ ls bex_mount
    bin   dev                         entrypoint.sh  home  lib64  mnt  proc  run   selinux  sys  usr
    boot  docker-entrypoint-initdb.d  etc            lib   media  opt  root  sbin  srv      tmp  var
    bexelbie@bexelbie:~$ sudo atomic unmount bex_mount
    Device mounted at bex_mount is not a docker container.
    bexelbie@bexelbie:~$ mount | grep bex_mount
    /dev/mapper/docker-253:2-1451009-444fb76932aeb58f9f47946c1c380364353db3a000b9aea0322b19ae70812c08 on /home/bexelbie/bex_mount type xfs (ro,nosuid,nodev,relatime,context="system_u:object_r:svirt_sandbox_file_t:s0:c4,c384",attr2,inode64,sunit=1024,swidth=1024,noquota)
    bexelbie@bexelbie:~$ ls bex_mount/
    id  rootfs
    bexelbie@bexelbie:~$ sudo atomic unmount bex_mount
    Device mounted at bex_mount is not a docker container.
    bexelbie@bexelbie:~$ mount | grep bex_mount
[...]

PR #188, opened 2015-10-15
BUG: umount not working when pool and dev names dis-similiar

baude:
Atomic/mount.py
        Corrects https://github.com/projectatomic/atomic/issues/186

        Previously the devicemapper unmount function did
        some comparisons between the pool name and the device
        name to determine if the mount point being passed
        to atomic was in fact a container that was mounted.  This
        comparison did not work under certain conditions with
        thinpool/devicemapper mounts when there was a somewhat
        non-default configuration for docker-storage.

        Because the atomic mount function always adds the container
        ID to the device name, we can now reliably use that
        and check the list of existing container IDs for a match.

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -274,6 +274,19 @@ class DockerMount(Mount):
         Mounts a container or image referred to by identifier to
         the host filesystem.
         """
+        try:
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM

PR #190, opened 2015-10-16
Fixed up atomic-scan.1.md to be consistent with the fetch_cves arg

mpreisler:
Previously the old "--no-cache" was mentioned and "fetch-cves" was used
instead of "fetch_cves".

Using --fetch-cves resulted in:
```
atomic: unrecognized arguments: --fetch-cves=False
```


rhatdan:
LGTM

PR #191, opened 2015-10-21
Do not build atomic with an egg file

baude:
setup.py: egg files are hard to debug and working with,
        espicially on a platform like Atomic.

### DIFF
--- a/Atomic/__init__.py
+++ b/Atomic/__init__.py
@@ -1,6 +1,3 @@
-import sys
-import os
-from pkg_resources import get_distribution
 from .pulp import PulpServer, PulpConfig
 from .satellite import SatelliteServer, SatelliteConfig
 from .atomic import Atomic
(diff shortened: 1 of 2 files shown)


rhatdan:
@baude tests are blowing up with the change?

baude:
@rhatdan I believe this is correct now.  The result is that instead of getting all of the atomic "inside" the egg-file, we now get an egg-file for info but atomic ends up in its own directory in /usr/lib/python2/site-packages/ ... this is handy for doing debugging.

rhatdan:
Arrggg we have gone back and forth on this many times.  On where the version should be stored.

https://github.com/projectatomic/atomic/pull/130

@cgwalters @rtnpro

baude:
Well, the main change is for the egg builds.  The rest of the changes is to deal with the get_distribution call that was being used in __init__.  If you prefer the version, author, author_email, etc to be stored in setup.py I can rework.

rhatdan:
Yes it should stay in setup.py.

rhatdan:
I am merging this, if it is a problem.  @cgwalters and you need to get in a cage match.

PR #192, opened 2015-10-22
XFS needs 'nouuid' set

baude:
For systems that use XFS as the docker FS, 'nouuid' needs to
    be set to avoid duplicate UUID warnings when mounting
    the same container or image multiple times.  XFS is the default
    for RHEL7.

    This also fixes the test_mount.sh failure commonly
    observed on RHEL 7.


rhatdan:
LGTM

rhatdan:
Nice work.

PR #195, opened 2015-10-23
BUGFIX: Atomic throws exception when no scan_targets are passed

baude:
Martin Preisle has found a bug that when you run atomic scan
    without any values, it produces an exception.  This small
    patch verifies that if neither --all, --images, --containers
    is passed, then scan_targets must have at least one
    argument.


rhatdan:
LGTM

Issue #194, opened 2015-10-23
Debian support borked.


charliedrage:
Seems that
`$(PYLINT) -E --additional-builtins=_ *.py atomic Atomic tests/unit/*.py`

Added to `Makefile`

Has caused an issue when using
`make all`
on Debian.

Removing the file, or reverting back to:
`$(PYLINT) -E --additional-builtins _ atomic`

Fixes it.

Here is the error with using the current `make all` method:

```
▶ make all
/usr/bin/python setup.py build
running build
running build_py
running build_scripts
/usr/bin/pylint -E --additional-builtins=_ *.py atomic Atomic tests/unit/*.py
No config file found, using default configuration
************* Module atomic_dbus
E: 20, 4: Use of super on an old style class (super-on-old-class)
Makefile:18: recipe for target 'python-build' failed
make: *** [python-build] Error 2
```

pinging fellow canadian @jlebon

jlebon:
Hmm, interesting. (Also, howdy!)
What version of `pylint` are you running?

charliedrage:
@jlebon (howdy from waterlooooo)

```
▶ pylint --version
No config file found, using default configuration
pylint 1.4.4, 
astroid 1.3.8, common 1.0.2
Python 2.7.9 (default, Mar  1 2015, 12:57:24) 
[GCC 4.9.2]

~                                                                                                                                                                                                                                                             
▶ 
```
```

**Comment by rhatdan** [comment]
Seems like it does not like

```
    def __init__(self, *p, **k):
        super(atomic_dbus, self).__init__(*p, **k)
        self.atomic = Atomic.Atomic()
```

Any python guru out there to tell me a better way to do this?

**Comment by jlebon** [comment]
Weird, got the exact same version here:

```
$ pylint --version
No config file found, using default configuration
pylint 1.4.4, 
astroid 1.3.8, common 1.0.2
Python 2.7.8 (default, Apr 15 2015, 09:26:43) 
[GCC 4.9.2 20150212 (Red Hat 4.9.2-6)]
```

And yet:

```
$ pylint -E --additional-builtins=_ *.py atomic Atomic tests/unit/*.py                                                                                                                                                                  
No config file found, using default configuration
$ echo $?
0
```

**Comment by charliedrage** [comment]
Hmmm. After performing `apt-get update && apt-get install pylint` I'm now getting a different repository from Debian 8 official repos.

```
~/atomic  master ✔                                                                                                                                                                                                                                        7m  
[...]

PR #196, opened 2015-10-23
Update Debian.md

charliedrage:
Fixes issue #194

Note: no need for `make all` as there is an alias for it within `Makefile`


jlebon:
LGTM!

PR #198, opened 2015-10-28
Atomic migrate

shishir-a412ed:
atomic migrate allows the user to easily migrate images, volumes, and containers from one version of Docker to another. With this command, users can quickly save all their data from the current docker instance, change the docker storage backend, and then import all their old data to the new system.


shishir-a412ed:
ping @rhatdan

shishir-a412ed:
@rhatdan `import` is a reserved keyword. Can't use `import`.

rhatdan:
Maybe capitalize it.

shishir-a412ed:
@rhatdan PTAL.

jlebon:
bot, test pull request

rhatdan:
Jenkins is failing?

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+
+def export_docker(graph, export_location):
+    """
+    This is a wrapper function for exporting docker images, containers
+    and volumes.
+    """
+
+    if os.geteuid() != 0:
+        exit("You need to have root privileges to run atomic export."
+             "\nPlease try again, this time using 'sudo'. Exiting.")
```
rhatdan: Remove second line.
"\nPlease try again, this time using 'sudo'. Exiting.")

We don't care how they become root.

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+
+    if not os.path.isdir(export_location):
+        os.mkdir(export_location)
```
rhatdan: Should be in try block.  use os.makedirs() in case export_location includes directories that do not exist.

os.mkdir("/foobar/does/not/exist") will blow up.

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+    try:
+	#Save the docker storage driver
+	storage_driver = subprocess.check_output("docker info|grep \"Storage Driver\"|cut -d\" \" -f 3", shell=True)
+	file = open(export_location+"/dockerInfo.txt", "w")
```
rhatdan: d = docker.Client()
Use self.d defined in Atomic to use docker.py
d.info()["Driver"]
u'devicemapper'

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+
+    util.writeOut("atomic export completed successfully")
+
+def export_images(export_location):
+    """
+    Method for exporting docker images into a filesystem directory.
+    """
+    if not os.path.isdir(export_location + "/images"):
+        os.mkdir(export_location + "/images")
+    images = subprocess.check_output("docker images|awk 'NR!=1{print $1 \":\" $2}'", shell=True)
```
rhatdan: d.images()
[{u'Created': 1444779592, u'Labels': None, u'VirtualSize': 186507296, u'ParentId': u'a83f22c39f43b9ae80b89251410ea6c5dabe716032917bb5b02510689cdec172', u'RepoTags': [u'fedora:latest', u'docker.io/fedora:latest'], u'RepoDigests': [], u'Id': u'8ee526697d01cda6a017affd97b31bdc98bfe245ac5971c9ee3ed7428b5e74e8', u'Size': 186507296}]

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+        error = sys.exc_info()[0]
+        sys.exit(error)
+
+    util.writeOut("atomic export completed successfully")
+
+def export_images(export_location):
+    """
+    Method for exporting docker images into a filesystem directory.
+    """
+    if not os.path.isdir(export_location + "/images"):
```
rhatdan: makedirs as above.  Could throw an exception.

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+                    images, export_location, id), shell=True)
+
+def export_containers(graph, export_location):
+    """
+    Method for exporting docker containers into a filesystem directory.
+    """
+    if not os.path.isdir(export_location + "/containers"):
+        os.mkdir(export_location + "/containers")
+
+    containers = subprocess.check_output("docker ps -aq", shell=True)
```
rhatdan: d.containers()
[]

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+                              " --graph={1} --export-location={2}"
+                              .format(split_containers[i], graph, export_location),
+                              shell=True)
+
+def export_volumes(graph, export_location):
+    """
+    Method for exporting docker volumes into a filesystem directory.
+    """
+    if not os.path.isdir(export_location + "/volumes"):
+        os.mkdir(export_location + "/volumes")
```
rhatdan: makedirs, can throw exception.

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+    util.writeOut("Exporting Volumes")
+    subprocess.check_call("tar --selinux -zcvf {0}/volumes/volumeData.tar.gz"
```
rhatdan: path tar.  /usr/bin/tar

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+                          " -C {1}/volumes . > /dev/null"
```
rhatdan: Don't redirect to > /dev/null
Use this instead
                return subprocess.check_call(cmd, stderr=DEVNULL)

Atomic/Export.py
```diff
@@ -0,0 +1,107 @@
...
+                          .format(export_location, graph), shell=True)
+    if os.path.isdir(graph + "/vfs"):
+        subprocess.check_call("tar --selinux -zcvf {0}/volumes/vfsData.tar.gz"
+                              " -C {1}/vfs . > /dev/null"
```
rhatdan: return subprocess.check_call(cmd, stderr=DEVNULL)

Atomic/Import.py
```diff
@@ -0,0 +1,81 @@
...
+    and volumes.
+    """
+
+    if os.geteuid() != 0:
+        exit("You need to have root privileges to run atomic import."
+             "\nPlease try again, this time using 'sudo'. Exiting.")
+
+    if not os.path.isdir(import_location):
+        sys.exit("Specified directory {0} does not exist".format(import_location))
+    try:
```
rhatdan: Why are you inconsistent with exit calls?

Atomic/Import.py
```diff
@@ -0,0 +1,81 @@
...
+        subprocess.check_call("/usr/libexec/dockermigrate/migrate.sh import --container-id={0}"
+                              " --graph={1} --import-location={2}"
+                              .format(i[8:], graph, import_location), shell=True)
+
+def import_volumes(graph, import_location):
+    """
+    Method for importing docker volumes from a filesystem directory.
+    """
+    util.writeOut("Importing Volumes")
+    subprocess.check_call("tar --selinux -xzvf {0}/volumes/volumeData.tar.gz"
```
rhatdan: fully path

Atomic/Import.py
```diff
@@ -0,0 +1,81 @@
...
+                          " -C {1}/volumes > /dev/null"
+                          .format(import_location, graph), shell=True)
+    if os.path.isdir(graph + "/vfs"):
+        subprocess.check_call("tar --selinux -xzvf {0}/volumes/vfsData.tar.gz"
+                              " -C {1}/vfs > /dev/null"
```
rhatdan: fully path

Makefile
```diff
@@ -38,6 +38,9 @@ install: all
 	install -d -m 0755 $(DESTDIR)/usr/bin
 	ln -fs ../share/atomic/atomic $(DESTDIR)/usr/bin/atomic
 
+	install -d -m 0755 $(DESTDIR)/usr/libexec/dockermigrate
+	install -m 0755 migrate.sh gotar $(DESTDIR)/usr/libexec/dockermigrate
```
rhatdan: Lets call this migrate and put it in /usr/libexec/atomic/migrate and /usr/lib/exec/atomic/gotar

atomic
```diff
@@ -213,6 +213,37 @@ if __name__ == '__main__':
...
 
+    # atomic migrate
+    migratep = subparser.add_parser("migrate", help=_("migrate docker from one backend storage "
+                                                    "to another"))
+    migrate_subparser = migratep.add_subparsers(help=_("migrate commands"))
+
+   # atomic export
+    exportp = migrate_subparser.add_parser(
+        "export", help=_("export a docker instance into a filesystem directory"),
+        epilog="export a docker instance. "
```
rhatdan: export all containers/images/volumes.

Lets try to eliminate the word docker wherever possible.  There is potential this tool could be used in the future for non docker container tools like rocket or systemd-nspawn.

atomic
```diff
@@ -213,6 +213,37 @@ if __name__ == '__main__':
...
+        "The export command would export docker images, "
+        "containers and volumes into a filesystem directory.")
+    exportp.set_defaults(func=atomic.Export)
+    exportp.add_argument("--graph", dest="graph", default="/var/lib/docker",
+                        help=_("Root of the Docker runtime (Default: /var/lib/docker)"))
+    exportp.add_argument("--dir", dest="export_location",
+                         default="/var/lib/docker-migrate",
+                         help=_("Path for exporting docker (Default: /var/lib/docker-migrate)"))
+
```
rhatdan: Not consistant on Docker versus docker, probably better to just use containers.

Should migrate to /var/lib/atomic not /var/lib/docker-migrate.

Atomic/Export.py
shishir-a412ed: Fixed.

shishir-a412ed:
@rhatdan PTAL.

atomic
```diff
@@ -232,6 +232,37 @@ if __name__ == '__main__':
...
+        epilog="export container's content. "
+        "The export command would export images, "
+        "containers and volumes into a filesystem directory.")
+    exportp.set_defaults(func=atomic.Export)
+    exportp.add_argument("--graph", dest="graph", default="/var/lib/docker",
+                        help=_("Root of the Docker runtime (Default: /var/lib/docker)"))
+    exportp.add_argument("--dir", dest="export_location",
+                         default="/var/lib/atomic-migrate",
+                         help=_("Path for exporting container's content (Default: /var/lib/atomic-migrate)"))
+
```
rhatdan: Should be /var/lib/atomic/migrate

Atomic/Export.py
```diff
@@ -0,0 +1,118 @@
...
+        sys.exit("You need to have root privileges to run atomic export.")
+
+    if not os.path.isdir(export_location):
+        os.makedirs(export_location)
+
+    try:
+	#Save the docker storage driver
+        storage_driver = DOCKER_CLIENT.info()["Driver"]
+        filed = open(export_location+"/dockerInfo.txt", "w")
+        filed.write(storage_driver)
```
rhatdan: Why dockerinfo.txt?  Just info.txt

Atomic/Import.py
```diff
@@ -0,0 +1,85 @@
...
+    util.writeOut("Would you like to cleanup (rm -rf {0}) the temporary directory [y/N]"
+          .format(import_location))
+    choice = sys.stdin.read(1)
+    if (choice == 'y') or (choice == 'Y'):
+        util.writeOut("Deleting {0}".format(import_location))
+        subprocess.check_call("rm -rf {0}".format(import_location), shell=True)
+    else:
+        util.writeOut("Cleanup operation aborted")
+    util.writeOut("Please restart docker daemon for the changes to take effect")
+
```
rhatdan: systemctl restart docker

Atomic/Import.py
```diff
@@ -0,0 +1,85 @@
...
+    for i in split_tarballs:
+	util.writeOut("Importing image with id: {0}".format(i[:-4]))
+        subprocess.check_call("docker load < {0}/images/{1}".format(import_location, i), shell=True)
+
+def import_containers(graph, import_location):
+    """
+    Method for importing docker containers from a filesystem directory.
+    """
+    if not os.path.isdir(import_location + "/containers"):
+        sys.exit("Specified directory {0} does not exist.No containers to import."
```
rhatdan: does not exist. No containers to import

docs/atomic-migrate.1.md
```diff
@@ -0,0 +1,57 @@
...
+containers from one version of atomic to another. With this command, 
+users can quickly save all their data from the current atomic instance, 
+change the container's content storage backend, and then import all their old data 
+to the new system.
+
+# COMMANDS
+**export**
+
+export command will export all the current images, volumes, and containers
+to the specified directory, in the /images, /volumes, /containers subdirectories.
```
rhatdan: Tell me the default. /var/lib/atomic/migrate

shishir-a412ed: The defaults are mentioned in `export` and `import` `--dir` option.

rhatdan: Are you worried about space.  Please add it here also.  specified directory, /var/lib/atomic/migrate by default.  Content will be placed in the /images, /volumes, /containers subdirectories.

shishir-a412ed: Fixed.

rhatdan:
This LGTM for now, I would like it to be moved to a migrate class when you get back , or while you are on your trip.

shishir-a412ed:
@rhatdan oh wait :)
I was working on `systemctl restart docker` fix and there is an issue in test.
I ll push fix patches to atomic later.

ISSUE IN JENKY (LOCAL):
Any idea why the daemon is trying to start a container in the import phase. `atomic import` is just about creating a new container and updating the metadata. Getting this error.

Error response from daemon: Cannot start container b839d3341bad8243000c1f187be197f2ad5f14dc429290faaac1274812c0d10c: [8] System error: mkdir /var/lib/docker/overlay/b839d3341bad8243000c1f187be197f2ad5f14dc429290faaac1274812c0d10c/merged/run/secrets: cannot allocate memory

rhatdan:
Looks like you are having a problem with the secrets patch.  What is atomic import?

shishir-a412ed:
@rhatdan I meant `atomic migrate import`

rhatdan:
You create a new container when you do an atomic migrate import.  No idea why it would be doing a start.

shishir-a412ed:
@rhatdan I just checked. We do a `docker run` to create (and run) a new container and apply the diff on it (at the entry point).
https://github.com/projectatomic/atomic/blob/master/migrate.sh#L172

Is there a case with secrets patch, where you can only do `docker create` and NOT `docker run` ?
Also, has the secrets patch been tested on overlayfs ? 

System error: mkdir /var/lib/docker/overlay/b839d3341bad8243000c1f187be197f2ad5f14dc429290faaac1274812c0d10c/merged/run/secrets: cannot allocate memory

rhatdan:
Do we actually need to run the container, or could we just do a docker create?

I am not sure on secrets on overlay.

shishir-a412ed:
@rhatdan I remember `docker create` was failing for this. So we had to do `docker run` to apply the diff.
But I might be wrong. so let me double check.

shishir-a412ed:
@rhatdan I tested the above problem. We need `docker run` to apply the diff.
I changed it to `docker create` in the code and after importing the container, I didn't see the diff.
With `docker run` I see the diff applied properly.

rhatdan:
Ok so we need to figure out why the secrets patch is causing problems, and how come I am not seeing it.  Is this just failing in test?  You need to work on the other problem for Vivek if this is only a test issue.

jlebon:
@shishir-a412ed -- As @baude pointed out to me, it seems like the `test_migrate.sh` test is failing because it's trying to create an image that already exists from the previous test run:

```
+ /usr/bin/docker create --name test-migrate-1 -v /tmp fedora /bin/bash
Error response from daemon: Conflict. The name "test-migrate-1" is already in use by container acd8665d2141. You have to delete (or rename) that container to be able to reuse that name.
+ cleanup
```

Could you modify the tests so that any image of the same name, if it exists, is removed before creating? Just like how it's done in `test.sh`:

```bash
# Remove the old image... Though there may not be one.
set +e
${DOCKER} rmi ${iname} &>> ${LOG}
set -e
```

The alternative is that I make the tests clear out all the images it has, but that would also wipe out the larger images such as `docker.io/fedora` and `docker.io/centos`, which would then make the tests take longer to execute. I'd rather not keep a whitelist, which might rot down the road.

shishir-a412ed:
@jlebon you mean "the container" :)
I have a couple more issues along with this. I will send a fixup PR soon.

jlebon:
@shishir-a412ed Ahh, apologies. Yes, didn't read closely enough. :)
I don't mind wiping containers in the tester actually. That sounds like a good cleanup I should do anyway.

shishir-a412ed:
@jlebon no worries. The containers (along with the associated volumes) are removed in my tests.
https://github.com/shishir-a412ed/atomic/blob/master/tests/integration/test_migrate.sh#L50

However, to solve this problem, I have use IDs instead of names. as container name is conflicting with an existing container name.

jlebon:
@shishir-a412ed I think another issue also is that in your `cleanup()` routine, you might not end up executing everything since we have `set -e` turned on. For example:

```bash
#!/bin/bash
set -euo pipefail

function cleanup {
        echo "In cleanup"
        false
        echo "After false"
}

trap cleanup EXIT

false
```

And running the above:

```
$ bash tmp.sh
In cleanup
$ 
```

Notice how it doesn't reach the "After false" because `false` made it fail. Similarly, in your `cleanup()`, if e.g. the `umount` command fails, it will never run the following statements. I'm thinking you can either do a `set +e` at the beginning of the function, or check before running each command that they're necessary (probably a better option).

PR #201, opened 2015-11-06
Proxy new `rpm-ostree deploy` command

mbarnes:
I just added a new `deploy` command to `rpm-ostree` in https://github.com/projectatomic/rpm-ostree/pull/178.

Here's the changes necessary to proxy it through `atomic host`.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -549,6 +549,16 @@ class Atomic(object):
             argv.append("--os=" % self.args.os )
         self._rpmostree(argv)
 
+    def host_deploy(self):
+        argv = ["deploy", self.args.revision]
+        if self.args.reboot:
+            argv.append("--reboot")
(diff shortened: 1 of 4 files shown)


mbarnes:
Man page updated.

I noticed the options section isn't quite right since different subcommands have different options, but wasn't sure how much you care since we don't ship man pages on Atomic.

rhatdan:
We ship the man pages inside of the *-tools containers, RHEL-tools for example.

We should fix the man pages but that should not block this request.

mbarnes:
Fixed the bash completion for the `deploy` command.  This can be squashed with the first commit.

I also went through the other `atomic host` commands and updated their bash completion.

rhatdan:
LGTM squash the commits, and I will merge.

mbarnes:
Squashed.

PR #202, opened 2015-11-09
BUGFIXES: fix critical atomic scan bugs related to Mount

baude:
During QE of 'atomic scan', they discovered that when scanning
    of containers occurred, we frequently were left with <none>
    type images.  This was because when containers are mounted, the
    docker mount code actually copies (_clone, docker commit) into
    a new image from which a temporary container is then made.
    However, the temporary image was never removed post-processing.

    While fixing this bug, I discovered two addition bugs that
    needed attention.  When doing a scan, the docker mount backend
    would throw an exception when trying to make a temporary
    directory that already existed.

    And finally, the docker mount backend would throw an exception
    when it tried to work with containers that were already
    mounted.  After speaking with vgoyal and given the inability to
    guess what might be going on there programatically, I have
    injected a quick check prior to handing off to the openscap-
    daemon where if we find that one of the docker objected to be
    scanned is already mounted, we error out there and display
    a nice message to the user explaining that we cannot work
    with busy docker objects.  Per vgoyal, this also follows
    one of docker's golden rules.

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -176,6 +176,7 @@ class DockerMount(Mount):
         Mount.__init__(self, mountpoint, live)
         self.client = docker.Client()
         self.mnt_mkdir = mnt_mkdir
+        self.tmp_image = None
 
     def _create_temp_container(self, iid):
         """
@@ -210,7 +211,12 @@ class DockerMount(Mount):
             )['Id']
         except docker.errors.APIError as ex:
             raise MountError(str(ex))
+        self.tmp_image = iid
         return self._create_temp_container(iid)
+
+    def _is_container_running(self, cid):
+        cinfo = self.client.inspect_container(cid)
+        return cinfo['State']['Running']
 
     def _identifier_as_cid(self, identifier):
         """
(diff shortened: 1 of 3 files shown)


Atomic/mount.py
```diff
@@ -449,6 +460,12 @@ def _cleanup_container(self, cinfo):
             except Exception as e:
                 raise MountError(e)
 
+    def _clean_tmp_image(self):
+        # If a temporary image is created with commit,
+        # clean up that too
+        if self.tmp_image is not None:
+            self.client.remove_image(self.tmp_image)
+
```
rhatdan: Shouldn't we add  
self.tmp_image = None

baude: That is done in init by default -> https://github.com/projectatomic/atomic/pull/202/files#diff-ce5473cd96a720a53d0834698c9b2eaaR179

rhatdan: Yes, but in this case some one could theoretically call _clean_tmp_image() a second time.  If they did, you would have a failure, if you reset the value, the second call would not fail.

rhatdan:
LGTM

PR #206, opened 2015-11-11
Return the same message when we fail to connect to the docker daemon

rhatdan:
Docker client returns

"Cannot connect to the Docker daemon. Is the docker daemon running on this host?"

atomic info should also trigger this error.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -331,8 +331,8 @@ class Atomic(object):
         except docker.errors.APIError:
             pass
         except requests.exceptions.ConnectionError as e:
-            raise IOError("Unable to communicate with docker daemon: %s\n" %
-                          str(e))
+            raise IOError("Cannot connect to the Docker daemon. Is the docker daemon running on this host?")
+
         return None
 
     def _inspect_container(self, name=None):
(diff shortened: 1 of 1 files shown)


Atomic/atomic.py
```diff
@@ -648,7 +648,7 @@ def _no_such_image():
         inspection = None
         if not self.args.force_remote_info:
             try:
-                inspection = self.d.inspect_image(self.args.image)
+                inspection = self._inspect_image(self.args.image)
```
dustymabe: Ahh.. So the original error message was good enough, but there was a bug right here where we weren't actually calling our internal version of `_inspect_image()`. Cool.

Atomic/atomic.py
```diff
@@ -331,8 +331,8 @@ def _inspect_image(self, image=None):
         except docker.errors.APIError:
             pass
         except requests.exceptions.ConnectionError as e:
-            raise IOError("Unable to communicate with docker daemon: %s\n" %
-                          str(e))
+            raise IOError("Cannot connect to the Docker daemon. Is the docker daemon running on this host?")
+
```
dustymabe: The original message here was good enough. If you want to leave the question in there that is fine but might still be useful to have the `str(e)` in there. 

Either way works. Make the change if you want or leave it like is.

dustymabe:
LGTM.. Feel free to take my suggestions or just merge as is.

PR #213, opened 2015-11-16
Should be export_docker not export_containers

rhatdan:


rhatdan:
Not related.  I am merging this code.

PR #217, opened 2015-11-17
Do not continue atomic install without INSTALL label

baude:
When running atomic install on an image that has no
    INSTALL label, atomic continued to running  a new
    container without any install arguments.  For example:

    [bbaude@localhost atomic]$ sudo ./atomic install busybox
    /usr/bin/docker run -t -i --rm --privileged -v /:/host --net=host --ipc=host --pid=host -e HOST=/host -e NAME=busybox -e IMAGE=busybox -e CONFDIR=/host/etc/busybox -e LOGDIR=/host/var/log/busybox -e DATADIR=/host/var/lib/busybox --name busybox busybox
    / #

    This behavior is not desirable.  We would rather have
    a "gentle" exit like:

    [bbaude@localhost atomic]$ sudo ./atomic install busybox
    This image has no INSTALL label to execute.

    [bbaude@localhost atomic]$

    We do not impact the "pull" of the image if it does
    not exist locally.


rhatdan:
I agree, but you need to fix the man pages and remove the INSTALL_ARGS?  Also I think we should just install the image and do nothing, IE Not report it.  

atomic install 

Just becomes 

docker pull

I guess if you specify a NAME we should report an error.

baude:
@rhatdan  updated per our conversation.

PR #218, opened 2015-11-18
atomic verify: remote inspect and more details

baude:
With the remote_inspect function in RH docker, we can inspect
    and image on a remote repository for various information like
    versions (when present).  This allows us to expand atomic
    verify to check a local version against a remote version.

    atomic verify now can take an image as input (as before) and
    provide a greater level of detail when checking each base
    image (defined as a non-intermediate image).  It now iterates
    on those base images looking for update status.

    The output of atomic verify has also been changed slightly to
    include a verbose option.  When invoked, the verify output
    will list each base image with its versioning information.

    The non-verbose output remains largely the same, where only
    base images that have identified updates are put to stdout.
    If verbose is not called for by the user but we find base
    images with no version information, we output a warning
    message and print verbosely anyway.

    Updated man page


baude:
@rhatdan Im going to close this PR and submit a new one once we talk.  Too much has changed in the codebase.

Issue #165, opened 2015-09-22
command-not-found-handle that redirects to the toolsimg does not work with sudo


ibotty:
which is expected, but makes it inconvenient to use. It's possible to use

```
-                atomic run ${TOOLSIMG} "$@"
+                sudo atomic run ${TOOLSIMG} "$@"
```

Is that a good idea?

rhatdan:
You want to fix the test script to use sudo?

rhatdan:
You want atomic.sh to run as non root?  If it works, I have no problem with the change.

ibotty:
I kind of expected `sudo command` to not work (call `sudo atomic run $TOOLSIMG command`). Always calling sudo in the command-not-found-handle will allow people to try to run `command` as regular user, but will call `command` within the tools container _as root_. That's maybe not great either.

Ultimately, I only want `sudo strace` to work. I am fine with getting a permission denied when trying to run something as regular user.

rhatdan:
What happens with sudo strace?

baude:
It is worth mentioning that things like the mount function *do* require sudo; I cannot say if other subcommands are the same or not.

rhatdan:
Currently we are not differentiation atomic, might want to add a check to make sure it is running as root.  Since I don't want people using docker socket as non root, you need to run atomic as root to talk to it.

ibotty:
I think my intention got lost in the conversation. For now when you are not root, you cannot use the command-not-found handle at all, because when calling `sudo not_existing_command` it will not run the handler. When calling `not_existing_command` it will try to run `atomic` as non-root which will fail.

I want 
 * the atomic command-not-found handle be able to be run under sudo (don't know whether that's possible), or (preferably both)
 * the command-not-found handle to run atomic with sudo, so it works when just called without sudo.

PR #222, opened 2015-11-20
Exit atomic if not run as root

rhatdan:
Adding to fix https://github.com/projectatomic/atomic/issues/165

### DIFF
--- a/Atomic/Export.py
+++ b/Atomic/Export.py
@@ -22,9 +22,6 @@ def export_docker(graph, export_location):
     and volumes.
     """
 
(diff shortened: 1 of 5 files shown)


rhatdan:
@baude @cgwalters PTAL

baude:
@rhatdan Your diff looks good to me.  Can I suggest, however, that you also remove similar references in import/export?

https://github.com/projectatomic/atomic/blob/master/Atomic/Export.py#L26

rhatdan:
Ok I removed them all.

Issue #225, opened 2015-11-23
Atomic scan json -- mess on output


ybznek:
When you want to get JSON output e.g. ``` atomic scan TEST_RHEL7 --json ```, 
you will get "Scanning..." before JSON data, so you cannot parse it easily.

Atomic should not prepend "Scanning..." before JSON. 

```
$ sudo atomic scan DOCKER_TEST_RHEL7 --json 2> /dev/null

Scanning...

{
    "allcontainers": false,
...
```

rhatdan:
https://github.com/projectatomic/atomic/pull/227 fixes this issue.

PR #226, opened 2015-11-23
atomic scan: Suppress 'scanning' output with json

ybznek:
Should fix https://github.com/projectatomic/atomic/issues/225


ybznek:
Ok, thanks.

PR #227, opened 2015-11-23
Don't output Scanning, user knows that

rhatdan:


rhatdan:
Replaces https://github.com/projectatomic/atomic/pull/226

PR #229, opened 2015-11-24
diff: Quiet pylint warning

cgwalters:
It doesn't understand native code bindings.


rhatdan:
Taking yours instead of mine.

PR #230, opened 2015-11-24
Fix bogus pylint error

rhatdan:


cgwalters:
This is a dup of https://github.com/projectatomic/atomic/pull/229 no?  I believe mine is better as well, as it avoids turning off warnings for the whole file.

cgwalters:
Is there a reason this one didn't wait for review?

rhatdan:
Just wanted to get it fixed.   As soon as I saw yours I reverted.

PR #231, opened 2015-11-24
I merged Colins patch instead of mine

rhatdan:
This reverses my previous patch


cgwalters:
I would have used `git revert` for this BTW.

rhatdan:
I will try that next time.

PR #232, opened 2015-11-24
Mount

rhatdan:
atomic mount is broken on python3 with XFS.


rhatdan:
@baude @cgwalters PTAL

baude:
@rhatdan looks good. nice catch.

PR #233, opened 2015-11-24
Fix output of diff to work on python3

rhatdan:
rpm returns bytes for python3, need to fix encoding to printout correctly.

### DIFF
--- a/Atomic/diff.py
+++ b/Atomic/diff.py
@@ -158,16 +158,19 @@ class RpmDiff(object):
         ts = rpm.TransactionSet(chroot_os)
         ts.setVSFlags((rpm._RPMVSF_NOSIGNATURES | rpm._RPMVSF_NODIGESTS))
         image_rpms = []
+        enc=sys.getdefaultencoding()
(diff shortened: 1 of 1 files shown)


rhatdan:
@baude @cgwalters PTAL

Atomic/diff.py
```diff
@@ -158,16 +158,18 @@ def _get_rpms(self, chroot_os):
         ts = rpm.TransactionSet(chroot_os)
         ts.setVSFlags((rpm._RPMVSF_NOSIGNATURES | rpm._RPMVSF_NODIGESTS))
         image_rpms = []
+        enc=sys.getdefaultencoding()
         for hdr in ts.dbMatch():  # No sorting  # pylint: disable=no-member
             if hdr['name'] == 'gpg-pubkey':
```
baude: Does this need to be decoded to compare properly? similar to the mount issue ?

rhatdan: Yes good catch

rhatdan: Fixed.

baude:
Looks great @rhatdan

PR #235, opened 2015-11-25
Make the tests pass

jlebon:
This patch series makes some fixes so that the tests pass again.

PR #198 was merged with the tests failing. This caused the PRs that followed to also be marked as failures, thus potentially hiding any new failures. On the plus side, these failures have also helped uncover a few issues with the testing infrastructure, which should now all be resolved.

Hopefully, the auto-tester will like this PR. Once master passes again, PRs that are rebased should get a clearer picture of whether things passed or failed and why.

### DIFF
--- a/Atomic/Export.py
+++ b/Atomic/Export.py
@@ -28,7 +28,7 @@ def export_docker(graph, export_location):
         os.makedirs(export_location)
 
     try:
-	#Save the docker storage driver
+    #Save the docker storage driver
         storage_driver = DOCKER_CLIENT.info()["Driver"]
         filed = open(export_location+"/info.txt", "w")
         filed.write(storage_driver)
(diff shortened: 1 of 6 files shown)


jlebon:
@rhatdan @shishir-a412ed PTAL

jlebon:
Nice, thanks @rh-atomic-bot! :punch:

jlebon:
Updated PR. Addressed @cgwalters's comments and also decided to give a whack at replacing those `subprocess` calls to not use the shell.

cgwalters:
Looks pretty good to me!  Thanks a lot for cleaning this up.  :+1:

rhatdan:
BTW @jlebon great work, thanks for cleaning this up.

jlebon:
Thanks @cgwalters and @rhatdan!

PR #236, opened 2015-11-30
Proposal: Move lengthy subfunctions to classes

baude:
Proposal PR to set a pattern where if an atomic subfunction
    is lengthy, then we can break that out into its own class.
    This reduces the overall size of the Atomic class and allows
    new development and debug to be cleaner.

    Instead of using the func function of argparser, we pass
    the Class and Function that should be run as variables. For
    example, the Atomic verify subfunction would now be:

        _class=Atomic.Atomic, func='verify'

    The defined class is instantiated after parseargs is run.  We
    then perform the set_args and call the function.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -19,7 +19,6 @@ import dbus
 import docker
 import requests
 
-from . import diff
 from . import mount
 from . import util
 from . import satellite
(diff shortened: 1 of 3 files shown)


cgwalters:
Also can you drop `Proposal: ` from the commit subject?  After it's committed it will no longer be a proposal =)  And before things are committed, they're all basically proposals.

baude:
Updated to default to the atomic class per @rhatdan 's comment.  Dropped proposal from commit subject per @cgwalters .

rhatdan:
LGTM

PR #238, opened 2015-11-30
Use from . import methods

rhatdan:


rhatdan:
@baude @cgwalters PTAL

cgwalters:
Yes.  Incidentally the current PR tester won't catch this as it's not using a clean slate, but rpmdistro-gitoverlay would.

PR #239, opened 2015-12-02
BUGFIX: Atomic diff leaks FD (BZ # 1285006)

baude:
Under the condition that you diff two images and one or both
    of the images are not RPM based, no cleanup of the mounts and
    dm devices was done. This was reported via BZ #1285006.

    We now handle the exit and cleanup differently which counts
    for this condition.


baude:
@rhatdan Reworked to not query rpms unless *both* inputs are RPM-based.

rhatdan:
LGTM.

PR #243, opened 2015-12-04
Add noprune to image deletion

baude:
Correcting a bug observed by the openscap team where under
    certain conditions, the deletion of a temporary image would
    cause a traceback when noprune=False.

    Added a loop with timeout for unmounting by path to prevent
    race failures between openscap, docker, the filesystem, and/or
    dms.

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -24,7 +24,7 @@ import os
 import sys
 import json
 from fnmatch import fnmatch as matches
-
+import time
 import docker
 
 from . import util
@@ -156,13 +156,25 @@ class Mount:
         return stdout.replace('SOURCE\n', '').strip().split('\n')[-1]
 
     @staticmethod
-    def unmount_path(path):
+    def unmount_path(path, timeout=10):
         """
         Unmounts the directory specified by path.
         """
-        r = util.subp(['umount', path])
-        if r.return_code != 0:
-            raise ValueError(r.stderr)
+
+        # Added this timeout loop because it seems openscap/openscap-daemon
+        # still has a left over process running that causes the mount path
+        # to be busy and therefore causes the umount to fail.
+        #
+        # When that is fixed, this can revert to a simple command executed
+        # by subp.
+        for x in range(0, timeout, 1):
+            rc, result_stdout, result_stderr = util.subp(['umount', path])
+            if rc == 0:
+                return rc, result_stdout, result_stderr
+            sys.stderr.write("Warning: {}\nRetrying {}/{} to unmount {}\n"
+                             .format(result_stderr, x+1, timeout, path))
+            time.sleep(1)
+        raise ValueError("Unable to unmount {0} due to {1}".format(path, result_stderr))
 
 
 class DockerMount(Mount):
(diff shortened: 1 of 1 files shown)


ybznek:
@baude Thanks, it fixes random issues which we have in oscapd https://github.com/OpenSCAP/openscap-daemon/issues/34

cgwalters:
Commit message is indented again.

baude:
Lazy cannot be used because the next call will be to delete the CM device.  So all we would be doing is pushing the issue to dmsetup.  It needs to be done here unless someone else has a better idea.

Fixed indents.

cgwalters:
Can't we use `dmsetup remove --deferred` then?

rhatdan:
@rhvgoyal PTAL

cgwalters:
I'm not strongly objecting to this patch, but I'd at least like to be sure we've tried to not poll.  Polling costs power, increases latency etc.

rhvgoyal:
I agree. That we can use lazy unmount followed by deferred device removal. But the question I have is that why mount point is buy and device is busy? Can that be added as part of commit message.

baude:
The fact is that we dont entirely know.  We suspect that either it is a race condition of some sort (typically lsof doesnt show any open files) or there is some leftover openscap process still wrapping up. Members of the openscap team have suggested as such.

So bottom line, while we have been unable to pin it down, i'd rather not have the scan fail and oscapd crash.

If folks are ok with lazy and deferred, i'll code it up as such.  I will say the proposed patch has been tested and works.

cgwalters:
It looks like deferred remove on the docker side requires adding:

```
--storage-opt dm.use_deferred_removal=true
```

to the daemon options.  Can you test that?

cgwalters:
@rhvgoyal in https://github.com/docker/docker/commit/51e059e7e90f37848596a0b6ec83f7835e6e4131 you say "By default feature will be turned off."  But why is that?

rhvgoyal:
@cgwalters Just being safe. It was such a new feature and there has been push back from lvm team about the idea of deferred remove and deferred delete in production. So I did not make these new features as docker deafault. Instead made them fedora/rhel defaults by modifying docker-storage-setup and passing them as options to docker daemon.

rhvgoyal:
@cgwalters I think problem @baude is facing not because of docker but because atomic is trying to unmount and remove a device. I suspect that this device is somehow used by docker an that's why it fails. Not sure though in what cases it is used by docker.

baude:
I believe this is up to everyones specs.  I added a route through for lazy/deferred when it detects if the options are set in dockerd.

rhatdan:
@rhvgoyal @cgwalters PTAL

baude:
Reworked per @cgwalters  comments

jlebon:
Seems like one of the unit tests is failing. Could you look into that?

```
UNIT TESTS:
...F.......
FAIL: test_mount_excepts_unknown_backend (test_mount.TestAtomicMount)
  File "/home/jenkins/workspace/atomic-upstream-test/atomic/tests/unit/test_mount.py", line 14, in test_mount_excepts_unknown_backend
    self.assertRaisesRegexp(mount.MountError, exp, m.mount, 'fedora:22')
AssertionError: "Atomic mount is not supported on the foobardriver docker storage backend." does not match "fedora:22 did not match any image or container."
```

Let me know if you're not able to reproduce this locally.

jlebon:
Thanks @baude for fixing the test. :+1:

baude:
@rhatdan Ok reworked per our conversation yesterday and it passes the tests.

baude:
Adding message as @rhatdan recommends.

rhatdan:
LGTM

PR #244, opened 2015-12-07
pep8: Convention updates.

ashcrow:
This PR includes pep8 fixes and standardized docstring quotes (```"""```)

### DIFF
--- a/Atomic/Export.py
+++ b/Atomic/Export.py
@@ -9,10 +9,10 @@ import docker
 
 from . import util
 
-
 DOCKER_CLIENT = docker.Client()
 
 ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
+
 
 def export_docker(graph, export_location):
     """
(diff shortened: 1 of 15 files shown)


ashcrow:
cc @cgwalters

cgwalters:
Personally I find the PEP8 line length limit to be overall worse.  Yes, sometimes long lines are better broken up.  But a lot of the time it just makes things worse, and this has a lot of the latter.  I find 100 to be a sane limit personally.

Anyways, I'll defer to @rhatdan for this.

My only minor comment; the commit message shouldn't have a period `.` at the end, see the rest of `git log` here and e.g. https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/log/

ashcrow:
@cgwalters I can revert some of it back and just leave in removing the whitespaces and pushing the code towards commonality if that would be better.

cgwalters:
FWIW in another project I help maintain we do:

https://git.gnome.org/browse/gobject-introspection/tree/Makefile.am#n69

cgwalters:
Are you using a tool to reindent this btw?  If it's by hand let's let @rhatdan weigh in before you redo it too many times =)

One other comment: If we want this to pass, let's add it to `make check` in the `Makefile`.

ashcrow:
@cgwalters it's by hand :smile:. I'll add a ```make check``` for it and wait to hear back. Thanks!

rhatdan:
I am not crazy about the line length changes.

ashcrow:
@rhatdan / @cgwalters is length 99 instead of 79 acceptable (based off the example from gobject-introspection)?

ashcrow:
@rhatdan updated with the latest comments. Does as 99 line length maximum make sense or do you think it should be any length?

rhatdan:
99 length is probably pretty good, I think we would need to look at the lines that are longer and figure out if they look good shorter.

rhatdan:
Maybe allow them unlimited, and do one patch, then do a second patch at 99. so we could decide then.

rhatdan:
@ashcrow Are you still interested in this?

ashcrow:
@rhatdan I am but we can close this. If I am able to get back to looking I'll open a new PR.

rhatdan:
Great thanks.

PR #245, opened 2015-12-09
Atomic TOP: Top-like display for container processes

baude:
Adding a new atomic sub-command that behaves like GNU top
but for processes being run for containers.  It currently
displays the container id,  container name,
pid, cpu% (as reported by docker
top), mem% (as reported by docker top), and the command.

You can optionally pass in -o ppid, stime, time to collect
more data on the processes themselves.

While in the interactive display, you can also sort on
the columns to re-organize the data as needed.

You can define an interval for refreshing the process
information.

atomic top can be run without any additional
parameters.  If that is the case, it will by default
show processes for all active containers.  You can also
add one or more container_ids for exclusive process
monitoring by container.

Also added an AtomicDocker class to atomic.py which
allows for custom docker, python-api calls without
having to re-invent the wheel.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -1054,12 +1054,12 @@ class Atomic(object):
 
         return self.containers
 
-    def get_active_containers(self):
+    def get_active_containers(self, refresh=False):
         '''
         Wrapper function for obtaining active containers.  Should be used
         instead of direct queries to docker
         '''
-        if not self.active_containers:
+        if not self.active_containers or refresh:
             self.active_containers = self.d.containers(all=False)
 
         return self.active_containers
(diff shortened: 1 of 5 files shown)


baude:
@rhatdan Mind checking my bash completion syntax? This isnt something I am very skilled at.

jlebon:
@baude Could you add some tests for this?

This isn't related to your PR only, and I'm not sure what others think (@cgwalters, @rhatdan?) , but it'd be nice if in the future every new features (and maybe even bug fixes) also came with some tests. I'm not one to fly by coverage numbers, but it *is* pretty low, which means we probably wouldn't even notice if something broke/regressed.

```
Coverage report:
Name                  Stmts   Miss Branch BrPart  Cover
-------------------------------------------------------
Atomic/Export.py         54      5     20      7    84%
Atomic/Import.py         47      6     12      4    83%
Atomic/__init__.py        7      0      0      0   100%
Atomic/atomic.py        713    470    334     24    28%
Atomic/diff.py          208    171     84      0    13%
Atomic/mount.py         267     98     82     20    59%
Atomic/pulp.py          157    130     44      0    14%
Atomic/satellite.py     213    186     92      0    10%
Atomic/top.py            98     81     50      0    13%
Atomic/util.py          145     86     68      8    38%
-------------------------------------------------------
TOTAL                  1909   1233    786     63    30%
```

baude:
I'm not opposed to writing tests but I would prefer we merge this first as I have a number of PRs I need to work in that are dependent on code from this one.  I will create issues for tests for top and diff to remind myself.

jlebon:
@baude Awesome, that sounds good!

PR #248, opened 2015-12-10
In docker-1.10 the device/imageIDs are not == to container ID

rhatdan:
This is causing atomic unmount to fail. This searches for the DeviceName
in the list of containers to umount.

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -84,7 +84,7 @@ class Mount:
         DM device id 'dm_id' in the docker pool.
         """
         table = '0 %d thin /dev/mapper/%s %s' %  (int(size)/512, pool, dm_id)
-        
+
         cmd = ['dmsetup', 'create', name, '--table', table]
         r = util.subp(cmd)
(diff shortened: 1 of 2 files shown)


runcom:
There currently a regression in latest docker which causes `DeviceName` to be the one of the image the container run from https://github.com/docker/docker/issues/18614

rhatdan:
Yes we are hung up on this issue right now,  the rest of this patch set is needed but until docker-1.10 is fixed atomic mount is broken.

runcom:
@rhatdan I merged https://github.com/docker/docker/issues/18655 which is fixing this, latest docker 1.10 master will have the correct `DeviceName` of the container
This is no more blocked

rhatdan:
You are awesome.

rhatdan:
Oops there are lots of other fixes here.  I will test this with the latest docker to see if everything works.

rhatdan:
https://github.com/projectatomic/atomic/pull/255 replaces this.

PR #249, opened 2015-12-11
CLeanups for running atomic top when no containers are running

rhatdan:


rhatdan:
@baude PTAL

baude:
@rhatdan This looks good and make sense.

PR #250, opened 2015-12-11
Install Check:

baude:
At present, when 'atomic run' is called, there is no check to see
if an atomic install has occurred (if needed).  This could result
in the an imaging faling to run or running incorrectly.

We introduce a file (/var/lib/atomic/install) that keeps a record
of which images have been installed.

The file looks something like:

```
{
    "install-test": {
        "image_name": "install_test",
        "installed": true,
        "instances": [
            {
                "date": "2015-12-11 14:09:27.957807",
                "iid": "56c6bd1160a15dbb2905779173cb3df7a0b1445f24378dc039fc7b76f08a0077"
            }
        ]
    }
}

```

atomic run will now check if the image in question has an INSTALL
label, and if so, it will see if the image has been installed.  The
check now is simply looking for the presence of the image name.  In
the future, the checking can be more complex, for example, by
combining the image name and image id.

atomic run also has a --nocheck option which allows users to override
the install check.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -24,7 +24,7 @@ from . import satellite
 from . import pulp
 from .Export import export_docker
 from .Import import import_docker
-
+from datetime import datetime
 
 IMAGES = []
 
@@ -89,6 +89,8 @@ class Atomic(object):
                 "-i",
                 "--name", "${NAME}",
                 "${IMAGE}"]
+
+    INSTALL_FILE = "/var/lib/atomic/install"
 
     def __init__(self):
         self.d = AtomicDocker()
@@ -387,7 +389,22 @@ class Atomic(object):
         command += self.image
         return command
 
+    def return_image_id_name(self):
+        iid = self.inspect['Id']
+        if 'Name' in self.inspect.get('Config', {}).get('Labels', {}):
+            name = self.inspect['Config']['Labels']['Name']
+        else:
+            infos = self.get_image_infos()
+            _name = (x['Tag'] for x in infos if x['Id'] == iid).next()
+            trim = _name.rfind(':')
+            name = _name[:trim]
+            if name.startswith("/"):
+                name = name[1:]
+
+        return (iid, name)
+
     def run(self):
+
         missing_RUN = False
         self.inspect = self._inspect_container()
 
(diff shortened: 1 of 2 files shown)


rhatdan:
@sctweedie @cgwalters @riekrh @aweiteka 

Lets discuss.

rhatdan:
The problem I see here is what happens on update.  How do we handle the fact that rsyslog container was priviously installed and atomic run works, now when atomic gets updated, atomic run fails saying that rsyslog has not been installed.

One thing we could do here is check to see if a container exists with this name, then we know if atomic install has happened, else we check for the existence of container name in /var/lib/atomic/install.

If neither, we tell user that install is necessary before run.

baude:
This is the flow for the data structure and various conditions. Consider image 'foobar' has an INSTALL label.

```
[bbaude@localhost atomic (atomic_install_check)]$ sudo ./atomic run foobar

Unable to verify if foobar has been installed before. Run 'atomic install foobar' or 'atomic run --nocheck foobar' to override this check
```

The "check" is being done by the name of the image.  We offer an override type function. Otherwise an atomic install would look like:

```
[bbaude@localhost atomic (atomic_install_check)]$ sudo ./atomic install foobar
docker run --rm --privileged -v /:/host -e HOST=/host -e IMAGE=foobar -e NAME=foobar foobar /bin/true
```

The history file looks like:

```
{
    "foobar": {
        "image_name": "foobar",
        "installed": true,
        "instances": [
            {
                "date": "2015-12-14 16:15:42.093573",
                "iid": "56c6bd1160a15dbb2905779173cb3df7a0b1445f24378dc039fc7b76f08a0077"
            }
        ]
    }
}
```

Note how the history is kept by the name of the image.  If the user runs another install (right now), then you will get another listing under "instances" with a new date.  If the image ID differs, it will be note.

Where we are lacking covering the use case of when a new image becomes available.  Re-running atomic install will overwrite (but hopefully back up the conf files that are exposed to the host).  On the other hand, the image might truly require the user to run an install  to propagate new files to the host (for example).

So ideally, I think when atomic run is executed, it should check if the image name is in the history AND if the iid is the same or not.  If the image iid differs, then we have to know what to do...

aweiteka:
> We introduce a file (/var/lib/atomic/install) that keeps a record
> of which images have been installed.

My concern would be around `/var/lib/atomic/install` file management. We know people won't use uninstall very much. Does it just keep growing over the life of the machine? Is there any risk in having a bunch of old references? Do we prune it during `docker rm`, `docker rmi`?

The install action has always felt a bit "magical" to me (what does it do?) so I don't use it frequently. This adds to the complexity, even if it's a very simple mechanism. However I think these are valid use cases that make working with containers more elegant and useful.

I'm not opposed to the patch, just cautious.

rhatdan:
Brent lets discuss this one some time this week.

rhatdan:
Closing this because of inactivity and no agreement on functionality.  We can revisit this in the future.

PR #252, opened 2015-12-15
Atomic Diff: RPMS arent being collected

baude:
Due to very small bug, the function that collects the
operating system version and RPMs was not being called.


rhatdan:
LGTM

PR #254, opened 2015-12-15
Atomic Top: fix repeated process information

baude:
Fixed an indention issue which caused the process
information to be repeated several times.


rhatdan:
lgtm

PR #255, opened 2015-12-16
 In docker-1.10 the device/imageIDs are not == to container ID

rhatdan:
This is causing atomic unmount to fail. This searches for the DeviceName
in the list of containers to umount.

Since I was constantly typing `atomic umount` rather then `atomic unmount`,
I thought we should add an alias.


rhatdan:
@baude @cgwalters @bexelbie PTAL

baude:
@rhatdan  LGTM, merge and i can hunt down any more verify issues

PR #257, opened 2015-12-16
Add tests for Atomic diff and top

baude:
Basic tests for atomic diff and top which should catch
basic code regressions.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -1031,19 +1031,6 @@ class AtomicDocker(docker.Client):
     A class based on the docker client class with custom APIs specifically for
     atomic
     """
-    def atomic_top(self, container, ps_args=None):
-        """
-        Same as the docker top API but allows passing of ps args
-        :param container: container ID
-        :param ps_args:  custom ps args
-        :return:
-        """
-        u = self._url("/containers/{0}/top".format(container))
-        params = {}
-        if ps_args is not None:
-            params['ps_args'] = ps_args
-        return self._result(self._get(u, params=params), True)
-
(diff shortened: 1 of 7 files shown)


jlebon:
Thanks, that's a good start!

One good future improvement on this is doing some data validation on the output (e.g. not just checking that `atomic top` quits nicely). Maybe those would belong more in new unit tests, or maybe they could be fit into these ones.

jlebon:
```
+ /usr/bin/coverage run --source=./Atomic/ --branch --append atomic top --quit
  File "atomic", line 501, in <module>
    sys.exit(_func())
  File "/home/jenkins/workspace/atomic-upstream-test/atomic/Atomic/top.py", line 92, in atomic_top
    old_settings = termios.tcgetattr(fd)
error: (25, 'Inappropriate ioctl for device')
```

I'm guessing this is probably due to Jenkins not allocating a TTY.

jlebon:
Hmm, it passed, but it seems like it didn't output anything. Is that expected?

```
                            ATOMIC TOP
                            
[7mCONTA(I)NER* (N)AME     (P)ID      (C)PU  (M)EM  CMD       [0m
```

jlebon:
bot, retest this please

baude:
@rh-atomic-bot  Ready for your review again ....

baude:
@rhatdan Good to go now.  Please consider and merge.  Feedback welcome.

jlebon:
@baude Nice! Would you consider adding a quick grep of the container ID in the `atomic top` output just so something like this doesn't slip by in the future?

rhatdan:
LGTM

PR #260, opened 2015-12-18
Atomic.atomic: improve atomic images output

chuanchang:
When list of images are empty, atomic images will hit
a python error 'max() arg is an empty sequence', which
is not friendly for users, with the patch, we can see
an expect result like 'There are no images available.'

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -715,6 +715,10 @@ class Atomic(object):
             return
 
         _images = self.get_images()
+        if not _images:
+            self.writeOut("There are no images available.")
+            return
+
         _max_repo, _max_tag = get_col_lengths(_images)
         col_out = "{0:" + str(_max_repo) + "} {1:" + str(_max_tag) + \
                   "} {2:12} {3:19} {4:10}"


jlebon:
bot, retest this please

chuanchang:
@rhatdan, yeah, the list of image should be empty , but subsequent programming hasn't check it , the result is a python error 'max() arg is an empty sequence' is hit, it is not friendly for users I think, an expected error is 'There are no images available.'

baude:
@chuanchang Can you provide an example command you are running?

chuanchang:
@baude, the cmdline as follows.

```shell
# atomic images
max() arg is an empty sequence
```

# rpm -q python-docker-py docker atomic
python-docker-py-1.4.0-118.el7.noarch
docker-1.9.1-9.el7.x86_64
atomic-1.7-3.git95b28a9.el7.x86_64

baude:
@chuanchang and @rhatdan 

How would you feel about something like this?

```
index 1e5bc55..a4c1ce4 100644
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -716,7 +716,10 @@ class Atomic(object):
             return
 
         _images = self.get_images()
-        _max_repo, _max_tag = get_col_lengths(_images)
+        try:
+            _max_repo, _max_tag = get_col_lengths(_images)
+        except ValueError as e:
+            raise ValueError("There are no images available")
         col_out = "{0:" + str(_max_repo) + "} {1:" + str(_max_tag) + \
                   "} {2:12} {3:19} {4:10}"
         self.writeOut(col_out.format("REPOSITORY", "TAG", "IMAGE ID",
```

The upside of this is the return code will then be 1.

rhatdan:
This works for me.

chuanchang:
@baude, I think the empty list of images is not a error or exception, for the empty images, in docker, the return code of docker images is 0.

chuanchang:
@baude and @rhatdan, Merry  Christmas :)

rhatdan:
Changing my mind on this.

I guess I would prefer to get_images() to return [] and this be checked.  I agree with @chuanchang this is not an error situation.

rhatdan:
I think other pull requests fixed this.

PR #261, opened 2015-12-18
F23 tweaks

jlebon:
In preparation for moving the tester to F23, some fixes are needed to get the current master to pass cleanly.


jlebon:
Updated to use `/run` instead.

jlebon:
Updated to put pylint override on the same line.

rhatdan:
LGTM

Issue #259, opened 2015-12-18
Allow users to define a specific test for make test


baude:
It would be preferable when designing or debugging atomic tests if you could specific a specific test to run.  The makefile and tests.sh should be modified to take a test name as a parameter and run just that test .... else run them all.

I have discussed with @jlebon and he had some ideas on implementation.

rhatdan:
I totally agree.  As I have been testing the `tests`, I would love this feature.

PR #262, opened 2015-12-18
test.sh: add TEST_INTEGRATION support

jlebon:
This allows users to skip all/certain tests.

```
$ sudo make test
	tests all integration tests

$ sudo make test TEST_INTEGRATION=
	skip all integration tests

$ sudo make test TEST_INTEGRATION='display pass'
	test only the 'display' and 'pass' integration tests
```

Resolves: #259

---

This is a first step at least for integration tests, which take the longest. A `TEST_UNIT` variable could be added at a later time for similarly filtering unit tests.

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -134,6 +134,15 @@ export ATOMIC="${COVERAGE}
 atomic"
 
 for tf in `find ./tests/integration/ -name test_*.sh`; do
+
+    if [ -n "${TEST_INTEGRATION+ }" ]; then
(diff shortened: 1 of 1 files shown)


jlebon:
Actually, this will run unwanted tests if the test name also happens to be a substring of another test name. Will tweak.

jlebon:
Updated.

rhatdan:
LGTM

PR #264, opened 2015-12-19
Correct blank version names

baude:
Due to a bug in atomic verify, sometimes the local and remote
versions ended up being blank.


rhatdan:
LGTM

PR #265, opened 2015-12-21
Add -e to docker top to correct TTY issues (BZ 1293063)

baude:
Depending on if a container has an allocated tty, you can get
different results from docker top unless you pass the -e
argument.  Adding the -e option will give correct results
regardless.

This is documented in https://bugzilla.redhat.com/show_bug.cgi?id=1293063.


rhatdan:
Makes sense.

LGTM

PR #266, opened 2015-12-22
Atomic: top: add optional uid,gid,user,group to ps_args

runcom:
`docker run -ti -u 1042 fedora /bin/bash` will run `/bin/bash` as user id `1042` for instance
`docker run -ti fedora /bin/bash` instead run as user id `0`

add an option to `ps_args` in `atomic top` which optionally display `UID` for the container

@baude I hope I've covered everything

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

### DIFF
--- a/Atomic/top.py
+++ b/Atomic/top.py
@@ -58,7 +58,15 @@ class Top(Atomic):
             {'shortname': 'STIME', 'column': '(S)TIME', 'character': 's', 'sort': True, 'index': 7,
              'field': '{:10}', 'active': False, 'ps_opt': 'stime', 'sort_order': True},
             {'shortname': 'PPID', 'column': 'PPI(D)', 'character': 'd', 'sort': True, 'index': 8,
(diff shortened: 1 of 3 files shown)


cgwalters:
bot, add author to whitelist

rhatdan:
If we are going to add this, shouldn't we also add user, gid and group?

runcom:
@rhatdan we could, btw, I've added `UID` because the default `ps -ef` will have that column

rhatdan:
Rewrite your patch with those added and I will merge.  Also make sure you update your man page.  It might be useful if these were listed in the help for -o option also.

runcom:
will do, I've already updated the man page for `UID`, I'll add `user,gid,group`

baude:
The code changes look fine.  Agree with others if you'd like to add user,gid,group

runcom:
updated patch to have optional `uid,gid,user,group`
@rhatdan @baude updated man as well, PTAL

rhatdan:
LGTM

PR #269, opened 2016-01-06
--prune is not working with python3 due to encoding

rhatdan:


rhatdan:
@baude @cgwalters @bexelbie PTAL

baude:
lgtm

cgwalters:
Can we use Linux-kernel (and systemd, and ostree, etc.) style commit messages please?  In this case:

 - Use a prefix header
 - Be imperative

Like:

```
prune: Fix encoding in python3
```

rhatdan:
Fine with me.

PR #270, opened 2016-01-07
Atomic/util.py: Hush benign pylint error

baude:
With CentOS and Debian, 'makes' were failing due to a
maybe-no-member pylint error.  The error is a false
positive.  Added #pylint: disable=maybe-no-member to hush
it.


cgwalters:
:+1:   (I'm curious why this is `maybe-no-member` but the other case is `no-member`.  But I'll assume this is right)

PR #272, opened 2016-01-08
Use the async API from openscap-daemon to perform CVE scans if possible

mpreisler:
The async API has several advantages, namely:
- it doesn't take over the dbus message queue of openscap-daemon,
  other users can still query it and interact with it
- it allows better scheduling in the daemon itself, daemon will
  limit the amount of concurrent work
- in the future it will be possible to cancel the work (CTRL+C),
  the change in atomic already does everything necessary, this
  only needs to be fixed in the daemon

In case the user has an older version of openscap-daemon we fallback to
the old synchronous API with huge timeout.

You need openscap-daemon from the git repository or 0.1.1 when it gets released to use the async API.

**I haven't tested this as much as I normally would, am in a hurry this Friday. Hopefully everything works, the implementation is very similar to `oscapd-cli scan` which has been tested a bit more.**

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -435,7 +435,33 @@ class Atomic(object):
                 fetch = 1
             else:
                 fetch = 0
-            scan_return = json.loads(oscap_i.scan_list(scan_list, 4, fetch, timeout=99999))
+
+            json_result = ""
+            try:
+                token = oscap_i.CVEScanListAsync(
+                    scan_list, 4, fetch
+                )
+                try:
+                    while True:
+                        success, json_result = oscap_i.GetCVEScanListAsyncResults(token)
+                        if success:
+                            break
+                        time.sleep(1)
+                except:
+                    oscap_i.CancelCVEScanListAsync(token)
+                    raise
+
+            except dbus.exceptions.DBusException as e:
+                if e.get_dbus_name() == \
+                        "org.freedesktop.DBus.Error.UnknownMethod":
+                    # Maybe we are using an old version of openscap-daemon and the
+                    # async API is not supported. Perform the scan with the old
+                    # synchronized API.
+                    json_result = oscap_i.scan_list(scan_list, 4, fetch, timeout=99999)
+                else:
+                    raise
+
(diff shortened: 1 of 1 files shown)


rhatdan:
bot, add author to whitelist

rhatdan:
LGTM but I don't know the calls. PTAL @baude

mpreisler:
Rebased to fix conflicts.

mpreisler:
Rebased once again to fix conflicts.

jlebon:
bot, retest this please

rhatdan:
@baude PTAL

Atomic/atomic.py
```diff
@@ -435,7 +435,33 @@ def scan(self):
...
+                )
+                try:
+                    while True:
+                        success, json_result = oscap_i.GetCVEScanListAsyncResults(token)
+                        if success:
+                            break
+                        time.sleep(1)
+                except:
+                    oscap_i.CancelCVEScanListAsync(token)
+                    raise
```
baude: Should there be a real exception raised here like ValueError with a real message?

Atomic/atomic.py
```diff
@@ -435,7 +435,33 @@ def scan(self):
...
+
+            except dbus.exceptions.DBusException as e:
+                if e.get_dbus_name() == \
+                        "org.freedesktop.DBus.Error.UnknownMethod":
+                    # Maybe we are using an old version of openscap-daemon and the
+                    # async API is not supported. Perform the scan with the old
+                    # synchronized API.
+                    json_result = oscap_i.scan_list(scan_list, 4, fetch, timeout=99999)
+                else:
+                    raise
```
baude: Same as above

Atomic/atomic.py
mpreisler: Hi,
this "empty" raise reraises the exception that was caught. The only thing this does is issue a cancel command and then reraise so something else handles the real exception. The best reference I found for this are the Python docs - https://docs.python.org/2/tutorial/errors.html - search for "re-raise"

baude:
LGTM.

PR #273, opened 2016-01-09
Atomic/help.py: Display man-like help for an image

baude:
Images or containers can now have an associated
man-like help page to help users understand more
about the image.  Typical information included
are things like a longer description, if the image
needs to be installed, security implications, steps
to upgrade, etc.

help.py extracts help page from the image/container
and then displays it in a pager.  The default path
for the help is /image_help.1 but this can be over-
ridden with the HELP label and fully-qualified
path including the file name.

### DIFF
--- a/Atomic/help.py
+++ b/Atomic/help.py
@@ -1,0 +1,95 @@
+from . import Atomic
+import subprocess
+from pydoc import pager
+import os
+from . import mount
+import sys
+from . import util
+
+
+class AtomicHelp(Atomic):
+
+    def __init__(self):
+        super(AtomicHelp, self).__init__()
+        self.mount_location = '/run/atomic'
+        self.help_file_name = 'help.1'
+        self.docker_object = None
+        self.is_container = True
+        self.use_pager = True
+        self.alt_help_cmd = None
+
+    def help(self):
+        """
+        Displays help text for a container.
+        :return: None
+        """
+        self.docker_object = self.args.image
+        docker_id = self.get_input_id(self.docker_object)
+        self.inspect = self._inspect_container(docker_id)
+        if self.inspect is None:  # docker_id is an image
+            self.inspect = self._inspect_image(docker_id)
+            self.is_container = False
+        else:
+            # The docker object is a container, need to set
+            # its image
+            self.image = self.inspect['Image']
+
+        # Check if an alternate help command is provided
+        labels = self._get_labels()
+        self.alt_help_cmd = None if len(labels) == 0 else labels.get('HELP')
+
+        if self.alt_help_cmd is not None:
+            self.display_alt_help()
+        else:
+            self.display_man_help(docker_id)
+
+    def display_man_help(self, docker_id):
+        """
+        Display the help for a container or image using the default
+        method of displaying a man formatted page
+        :param docker_id: docker object to get help for
+        :return: None
+        """
+        if not os.path.exists(self.mount_location):
+            os.mkdir(self.mount_location)
+        # Set the pager to less -R
+        enc = sys.getdefaultencoding()
+        if sys.stdout.isatty():
+            os.environ['PAGER'] = '/usr/bin/less -R'
+        else:
+            # There is no tty
+            self.use_pager = False
+        dm = mount.DockerMount(self.mount_location, mnt_mkdir=True)
+        mnt_path = dm.mount(docker_id)
+        try:
+            help_file = open(os.path.join(mnt_path, self.help_file_name))
+        except IOError:
+            pass
+        try:
+            help_file = open(os.path.join(mnt_path, 'rootfs', self.help_file_name))
+        except IOError:
+            dm.unmount(path=mnt_path)
+            raise ValueError("Unable to find help file for {}".format(self.docker_object))
+
+        cmd2 = ['groff', '-man', '-Tascii']
+        c2 = subprocess.Popen(cmd2, stdin=help_file, stdout=subprocess.PIPE)
+        result = c2.communicate()[0].decode(enc)
+        help_file.close()
+        if not self.use_pager:
+            util.writeOut("\n{}\n".format(result))
+        else:
+            # Call the pager
+            pager(result)
+
+        # Clean up
+        dm.unmount(path=mnt_path)
+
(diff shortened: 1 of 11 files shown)


baude:
Hold off on this one.  I see some issues I need to fix.

cgwalters:
Troff is a really ugly format, and most people generate man pages from something else that's slightly more sane, like Docbook or markdown.  What about using existing metadata labels in markdown format?

For anything longer, linking to online documentation seems sanest.

This would also be pulling `groff` into every host.

baude:
@cgwalters I checked an groff is in the cloud images, base RHEL, and atomic host.  I talked to man-db guys and they said man basically uses groff to display its content.

Atomic/help.py
```diff
@@ -0,0 +1,41 @@
...
+
+class AtomicHelp(Atomic):
+    def help(self):
+        """
+        Displays the image/containe help file in a pager
+        :return: None
+        """
+        # Set the pager to less -R
+        dm = mount.DockerMount("/tmp")
+        enc=sys.getdefaultencoding()
```
rhatdan: Probably should not be using /tmp.  Perhaps something under /run?

cgwalters:
In other words, this seems like it overlaps a bit with https://github.com/projectatomic/ContainerApplicationGenericLabels

That said I do applaud the idea.  Are you trying to encourage a manpage-style doc per container? That does seem like a good goal, I guess ContainerLabels is more like RPM metadata, and this would be more like manpages?

Atomic/help.py
```diff
@@ -0,0 +1,41 @@
...
+        environ['PAGER'] = '/usr/bin/less -R'
+        docker_id = self.get_input_id(self.args.docker_object)
+        self.inspect = self._inspect_image(docker_id)
```
rhatdan: Why do we need docker_id and help_id, shouldn't those be the same.

Atomic/help.py
baude: the DockerMount class requires a mount point for mounting images/containers.  The mount point is not used until something is actually mounted, so the passing of "/tmp" is quite benign.  I'm happy to change to '/run' if you prefer.

Atomic/help.py
```diff
@@ -0,0 +1,41 @@
...
+        help_id = docker_id
+        if self.inspect is None: # docker_id is a container
+            self.inspect = self._inspect_container(docker_id)
+            help_id = dm._clone(docker_id, image_only=True)
+        # Check if the help path has been overridden
+        alt_help_path = self._get_labels().get('HELP')
+        help_file = alt_help_path if alt_help_path is not None else '/image_help.1'
```
rhatdan: I think the spec says the HELP label is a command to execute.  Rather then a path to the help.

I think we should execute the HELP command so that the tool could present this data in a different way.

Default file to help.1 rather then image_help.1 since it could be a container.

Atomic/help.py
```diff
@@ -0,0 +1,41 @@
...
+        cmd1 = ['docker', 'run', '--rm', help_id, 'cat', help_file]
+        cmd2 = ['groff', '-man', '-Tascii']
```
rhatdan: What happens if we don't have cat?  Couldn't we just atomic mount the image and then use local tools like groff, rather then docker run.

Atomic/help.py
baude: To make sure i understand you correctly, if there is a HELP label, we execute than only and do not do anything else right?

Atomic/help.py
baude: atomic mounts have more overhead and opportunity for failure (IMHO), so I took this route as the most straight-forward.  Given that I misunderstood the HELP label prior, I think it would be more consistent to stick with a 'run' rather than a mount as well.  Do you foresee not having cat on RHEL, cloud-images, or atomic?

Atomic/help.py
baude: When the docker_id is a container, I am using DM._clone() which generates a temporary image from the container given and then run a temporary container.  The temporary container and image is cleaned up after execution.  This very much follows other mount operations.  In the case of using a container, I use help_id to differentiate between the docker_id and the ID of the temporary container.

Atomic/help.py
rhatdan: I foresee images that are not RHEL and do not contain Linux userspace.

Atomic/help.py
rhatdan: Yes that is what is described in the ContainerGenericLabel page.
https://github.com/projectatomic/ContainerApplicationGenericLabels

Atomic/help.py
rhatdan: I think we should just mount it and read the file if it exists rather then cloning and executing code within the container.

rhatdan:
@cgwalters We were going for a least common format that would look decent (Better then ASCII) but would have tools available on the host OS.  We figured groff was the best we had.  Packagers could use markdown and then convert to man page format.  I agree that if we had a tools that would display markdown in a terminal window, that would be better.

baude:
Ok, redone based on inputs and also fixed tests @rhatdan ready for re-review.

mfojtik:
a question: wouldn't be better to just set an LABEL on image that points users to a README.md (which I assume everyone that wants to have image on dockerhub do) and then just fetch that README.md, format it nicely for the console and display it? In this way you don't need to keep README.md content and man page in sync ;-)

baude:
@mfojtik Not all images are hosted in public git hubs so I don't think that will work.  Moreover, that would also require external web access to view help.  But ! That is where the HELP LABEL can come in.  You could use that to display a link to the README.md for reference or you could convert your README.md to a man format and display any which way you like.

rhatdan:
Right

HELP firefox http://myserver/myapp.html

Would work.

rhatdan:
@baude Could you add three tests for this.  One with help1, one with a container with a HELP label, and lastly a container with neither.

baude:
@rhatdan tests added

rhatdan:
LGTM except for comment on builtin HELP Function.

baude:
@rhatdan  better?

rhatdan:
LGTM

PR #274, opened 2016-01-11
Use `/usr/bin/env docker` instead of hardcoded path

bexelbie:
docker-machine and possibly others use /usr/local/bin/docker
not /usr/bin/docker

This fix should generalize atomic for those environments


rhatdan:
I have never been crazy about this type of change.  The problem here is users could get unexpected results.  IE Using an old version of docker when they think they are using the latest.  Also doing
sudo atomic  

versus 

sudo sh -c atomic

can act differently.  I am also concerned about the security vulnerablities from using environment variables.

bexelbie:
Some systems default install docker to a different location.  How would you suggest I modify the patch to allow for what I believe are the two use cases:

1 - I want to use a  docker other than /usr/bin/docker with atomic
2 - I want to build for a system that uses a location other than /usr/bin/docker

My initial thought on #1 was to use an environment variable, however that makes #2 ugly

rhatdan:
Could you create a link from /usr/bin/docker to /usr/local/bin/docker?

Is there a package that is installing docker to /usr/local?  

If you are using autotools you should do something like --prefix=/usr

bexelbie:
@rhatdan the default install of docker in some Linux variants is in /usr/local - to create a link would mean adding directions to users on those platforms to indicate the need to take this action.

I understand your concerns, but I'd like to find a way to support systems like docker-machine out of the box.

rhatdan:
We could change the tool to look in those two locations.  Rather then allow general use case.

Search /usr/bin/docker followed by /usr/local/bin/docker.  Or vice versa.

bexelbie:
What about just eliminating the path all together? so that calls to docker initiated directly by atomic would use the same executable as non-path specified labels (which seem to just use the first "docker" in the path).  I realize that this goes against your idea of specificity.

rhatdan:
We can do that.  Still would want to cleanse the Environment for PATH.

rhatdan:
https://github.com/projectatomic/atomic/pull/274 fixes this issue, I believe

rhatdan:
@bexelbie @baude @cgwalters PTAL

jlebon:
@rhatdan You mean #282, right?

rhatdan:
Yes/

baude:
LGTM.

rhatdan:
Merged alternative patch

PR #275, opened 2016-01-11
Use DOCKER_* environment variables

bexelbie:
Utilize the docker environment variables for non-local
daemons

### DIFF
--- a/Atomic/Export.py
+++ b/Atomic/Export.py
@@ -8,9 +8,10 @@ import subprocess
 import docker
 
 from . import util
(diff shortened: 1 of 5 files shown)


rhatdan:
These LGTM,  @baude @cgwalters PTAL

cgwalters:
Looks reasonable, and should make us consistent given we also use the client binary (which will honor the variables).

I have some longer term concern about the fact that some of the commands are inherently local (e.g. `mount`), but...eh.

PR #276, opened 2016-01-11
`atomic images` now handles the zero image case

bexelbie:
Prior to this fix, get_col_lengths() returned an error
if repo_tags was empty.  Now it returns minimal data


bexelbie:
I had not noticed #260 @rhatdan - if you'd prefer that I close this PR and wait on that one to complete, let me know.  This one does not return [], in returns 1,1 to minimize the rewrite for spacing code below.

rhatdan:
I would like to get that one fixed, but this LGTM, for a safety measure.

PR #277, opened 2016-01-11
Make git_images always return a list.

rhatdan:
I would prefer this pull request to https://github.com/projectatomic/atomic/pull/260

Basically we want get_images to always return a list, and if the list is empty to display nothing on an

atomic images list 

command.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -99,7 +99,7 @@ class Atomic(object):
         self.force = False
         self._images = []
         self.containers = False
-        self.images_cache = None
+        self.images_cache = []
         self.active_containers = []
 
     def writeOut(self, output, lf="\n"):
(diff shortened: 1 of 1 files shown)


rhatdan:
@bexelbie https://github.com/chuanchang PTAL

bexelbie:
@rhatdan because `docker images` returns headers in the zero image case, I think we should too.

bexelbie:
Reading the code again, I believe the code in #276 is a situation where there is always a list being passed.  If it isn't, I need a more clear guidepost to where it isn't.  The big change that #276 introduced was to ensure we always had a length for the formatting.  It defaults to 1 which means the empty list just displays headers and exits.

rhatdan:
I actually think docker images returningin NULL is wrong, and we have other places in the code where we do not check the code.

for i in self.get_images() 

For example.

rhatdan:
@baude @cgwalters PTAL

cgwalters:
The test is failing due to a syntax error.

rhatdan:
@cgwalters Fixed.

rhatdan:
@bexelbie @baude @cgwalters PTAL

Atomic migrate is failing, but unrelated to this patch.

baude:
@rhatdan this looks good.  makes sense

PR #283, opened 2016-01-13
Atomic/run.py: Add security implications messages based on RUN label

baude:
Laymen users who are told to run a image may not understand
the docker run switches that have security implications.  We
now look for the following switches:

* --privileged
* --cap-add
* --security-opt label:disable
* --net=host
* --pid=host
* --ipc=host

and output an appropriate security message.

Also, moved def run() from Atomic/atomic.py to Atomic/run.py
to reduce the size and the number of definitions in
Atomic/atomic.py.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -387,59 +387,7 @@ class Atomic(object):
         command += self.image
         return command
 
-    def run(self):
-        missing_RUN = False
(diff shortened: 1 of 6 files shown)


baude:
@rhatdan updated with man page and bash completion.  also added correction for a test failure in master.

rhatdan:
LGTM

PR #284, opened 2016-01-13
Fix atomic migrate test

shishir-a412ed:

### DIFF
--- a/Atomic/Export.py
+++ b/Atomic/Export.py
@@ -24,22 +24,28 @@ def export_docker(graph, export_location):
     if not os.path.isdir(export_location):
         os.makedirs(export_location)
 
-    try:
-    #Save the docker storage driver
-        storage_driver = DOCKER_CLIENT.info()["Driver"]
-        filed = open(export_location+"/info.txt", "w")
-        filed.write(storage_driver)
-        filed.close()
+    dangling_images = DOCKER_CLIENT.images(filters={"dangling":True}, quiet=True)
+    if any(dangling_images):
+        util.writeOut("There are dangling images in your system. Would you like atomic to prune them [y/N]")
+        choice = sys.stdin.read(1)
+        if choice.lower() == 'y':
+            util.writeOut("Deleting dangling images")
+            subprocess.check_call(["docker", "rmi", "-f"]+dangling_images)
+        else:
+            raise ValueError("Please delete dangling images before running atomic migrate export")
 
-        #export docker images
-        export_images(export_location)
-        #export docker containers
-        export_containers(graph, export_location)
-        #export docker volumes
-        export_volumes(graph, export_location)
-    except:
-        error = sys.exc_info()[0]
-        sys.exit(error)
+    #Save the docker storage driver
+    storage_driver = DOCKER_CLIENT.info()["Driver"]
+    filed = open(export_location+"/info.txt", "w")
+    filed.write(storage_driver)
+    filed.close()
+
+    #export docker images
+    export_images(export_location)
+    #export docker containers
+    export_containers(graph, export_location)
+    #export docker volumes
+    export_volumes(graph, export_location)
 
     util.writeOut("atomic export completed successfully")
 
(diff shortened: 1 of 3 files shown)


jlebon:
bot, add author to whitelist

rhatdan:
bot, add author to whitelist

shishir-a412ed:
@rhatdan I have fixed the comments. PTAL.
Also fixed some pylint formatting issues.

Atomic/Export.py
```diff
@@ -38,7 +49,7 @@ def export_docker(graph, export_location):
         #export docker volumes
         export_volumes(graph, export_location)
     except:
-        error = sys.exc_info()[0]
+        error = sys.exc_info()[1]
```
shishir-a412ed: sys.exc_info is a 3 tuple structure {type, value, traceback}. Since earlier I was reading index 0 (type) it was not printing the error message. sys.exc_info()[1] will print the correct error message. This should fix all our error handling problems.

shishir-a412ed:
@rhatdan yes it caches that `ValueError`. I tested that.

Atomic/Export.py
```diff
@@ -25,6 +25,17 @@ def export_docker(graph, export_location):
...
 
     try:
+        dangling_images = DOCKER_CLIENT.images(filters={"dangling":True}, quiet=True)
+        if dangling_images:
+            util.writeOut("There are dangling images in your system. Would you like atomic to prune them [y/N]")
+            choice = sys.stdin.read(1)
+            if choice.lower() == 'y':
+                for key in dangling_images:
+                    util.writeOut("Deleting dangling image with ID: {0}".format(key))
+                    subprocess.check_call(["/usr/bin/docker", "rmi", "-f", key])
```
jlebon: You should drop the `/usr/bin/` prefix now.

Atomic/Export.py
jlebon: Can't we use the usual Python syntax here?

```
except ValueError as e:
     error = e
```

tests/integration/test_migrate.sh
```diff
@@ -72,7 +72,7 @@ trap cleanup EXIT
 # at /var/lib/docker
 atomic_migrate () {
 	setup
-	${ATOMIC} migrate export --dir "$(pwd)/migrate-dir"
+	echo 'y'|${ATOMIC} migrate export --dir "$(pwd)/migrate-dir"
```
jlebon: It seems like there should be a new atomic flag e.g. `-y` or `--yes` so that these commands can be used non-interactively, like in these tests.

shishir-a412ed:
@rhatdan Fixed. PTAL.

tests/integration/test_migrate.sh
rhatdan: We are adding a --quiet in a different pull request.  I am not sure I have ever seen a --yes option on a command.

jlebon: Ahhh, sounds good. I was thinking of the yum/dnf `-y` switch, but apparently the long opt for it is `--assumeyes`, not `--yes` as I thought.

Atomic/Export.py
shishir-a412ed: @rhatdan can we remove `/usr/bin` ?

rhatdan: On 01/14/2016 10:20 AM, Shishir wrote:
>
> In Atomic/Export.py
> <https://github.com/projectatomic/atomic/pull/284#discussion_r49738363>:
>
> > @@ -25,6 +25,17 @@ def export_docker(graph, export_location):
> >          os.makedirs(export_location)
> >  
> >      try:
> > +        dangling_images = DOCKER_CLIENT.images(filters={"dangling":True}, quiet=True)
> > +        if dangling_images:
> > +            util.writeOut("There are dangling images in your system. Would you like atomic to prune them [y/N]")
> > +            choice = sys.stdin.read(1)
> > +            if choice.lower() == 'y':
> > +                for key in dangling_images:
> > +                    util.writeOut("Deleting dangling image with ID: {0}".format(key))
> > +                    subprocess.check_call(["/usr/bin/docker", "rmi", "-f", key])
>
> @rhatdan <https://github.com/rhatdan> can we remove |/usr/bin| ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/projectatomic/atomic/pull/284/files#r49738363>.
>
Yes, we have another patch that went in to remove all /usr/bin/docker

shishir-a412ed:
@rhatdan @jlebon Fixed the comments. PTAL.

rhatdan:
LGTM

PR #288, opened 2016-01-19
atomic: harden shell invocations (1248038)

baude:
Remove instances where shell=True on subprocess calls to
avoid possible exploitation.  Added def chk_call in util.py
which is similar to subp for re-use.

This addresses bugzilla 1248038.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -25,6 +25,7 @@ from . import pulp
 from .Export import export_docker
 from .Import import import_docker
 from docker.utils import kwargs_from_env
+import re
 
 IMAGES = []
 
@@ -470,8 +471,9 @@ class Atomic(object):
         args = self._get_args("STOP")
         if args:
             cmd = self.gen_cmd(args + list(map(pipes.quote, self.args.args)))
+            cmd = self.sub_env_strings(cmd)
             self.display(cmd)
-            subprocess.check_call(cmd, env=self.cmd_env, shell=True)
+            util.check_call(cmd, env=self.cmd_env)
 
         # Container exists
         try:
@@ -545,8 +547,9 @@ class Atomic(object):
         args = self._get_args("UNINSTALL")
         if args:
             cmd = self.gen_cmd(args + list(map(pipes.quote, self.args.args)))
+            cmd = self.sub_env_strings(cmd)
             self.display(cmd)
-            subprocess.check_call(cmd, env=self.cmd_env, shell=True)
+            util.check_call(cmd, env=self.cmd_env)
 
         if self.name == self.image:
             self.writeOut("docker rmi %s" % self.image)
@@ -554,37 +557,28 @@ class Atomic(object):
 
     @property
     def cmd_env(self):
-        env = dict(os.environ)
-        env.update({'NAME': self.name,
-                    'IMAGE': self.image})
+        os.environ['NAME'] = self.name
+        os.environ['IMAGE'] = self.image
 
         if hasattr(self.args, 'opt1') and self.args.opt1:
-            env['OPT1'] = self.args.opt1
+            os.environ['OPT1'] = self.args.opt1
 
         if hasattr(self.args, 'opt2') and self.args.opt2:
-            env['OPT2'] = self.args.opt2
+            os.environ['OPT2'] = self.args.opt2
 
         if hasattr(self.args, 'opt3') and self.args.opt3:
-            env['OPT3'] = self.args.opt3
+            os.environ['OPT3'] = self.args.opt3
 
         default_uid = "0"
         with open("/proc/self/loginuid") as f:
             default_uid = f.readline()
 
-        if "SUDO_UID" in os.environ:
-            env["SUDO_UID"] = os.environ["SUDO_UID"]
-        else:
-            env["SUDO_UID"] = default_uid
+        if "SUDO_UID" not in os.environ:
+            os.environ["SUDO_UID"] = default_uid
 
-        if 'SUDO_GID' in os.environ:
-            env['SUDO_GID'] = os.environ['SUDO_GID']
-        else:
-            try:
-                env['SUDO_GID'] = str(pwd.getpwuid(int(env["SUDO_UID"]))[3])
-            except:
-                env["SUDO_GID"] = default_uid
+        if 'SUDO_GID' not in os.environ:
+            os.environ["SUDO_GID"] = default_uid
 
-        return env
 
     def gen_cmd(self, cargs):
         args = []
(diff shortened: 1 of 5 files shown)


baude:
@cgwalters  can you elaborate?

cgwalters:
Let's say that we're trying to write a wrapper for `rm` which only allows users to delete files in a system-owned cache directory.  Something like allowing a desktop user to `yum clean all` in `/var/cache/yum`.

If you just do:

```
def attempt_at_safe_rm(path):
  if not path.startswith("/var/cache/yum"):
    fatal("Invalid path")
  arg_command = "rm " + path
  argv = arg_command.split(" ")
  subprocess.check_call(argv)
```

If the user provides input like:  "/var/cache/yum/foo.xml /etc/shadow" then `rm` will get both as separate arguments, and delete both files.

cgwalters:
So I think the correct fix to this is to go to the roots which is the labels.  Those should probably be split using e.g. `g_shell_parse_argv()`.  Systemd I think also has a variant of this.

The best case of course is if we supported argv array labels.  Docker introduced an array variant of `RUN ["foo", "--bar"]` for all of the reasons we're talking about here.

baude:
@cgwalters Thanks for the feedback.  One issue I foresee is we have labels that have commands in them like RUN and some which truly are more string related, like NAME.  I don't think we'd want to treat them differently but maybe so.

@rhatdan What are your thoughts?

baude:
@rhatdan updated per our conversation this week.  ready for re-review.

cgwalters:
Expanding the environment variables ourself makes sense, though looks like `os.path.expandvars()` leaves undefined variables alone, but I think we're tied into backwards compatibility into expanding them to nothing.

So I think this is farther along, but the bug won't be fixed until we can replace every instance of `util.chk_call()` back to `subprocess.check_call()` and assert the argument is an array.

baude:
Re-review please

rhatdan:
How come you are not using os.path.expandvars?

baude:
git failure on my part .... retry now.

rhatdan:
Well it LGTM, but this patch has a chance to break stuff.  I will ,merge but I would like to ping atomicapp people to make sure they test it with their code.

rhatdan:
@dustymabe Could you do some testing with new version of atomic.

dustymabe:
@rhatdan I'll try to run it through some tests and see if we notice any failures. Probably will happen tomorrow after FOSDEM is over.

cdrage:
@baude @rhatdan seems that it has failed. env variables seem to not be passed in correctly?

```bash
dev/adb-tests/cdrage-atomicapp-ci  master ✗                                                                                                                                                                                                           5d ⚑ ◒  
▶ sudo ./helloapache.sh run docker

##########
run: TESTING HELLOAPACHE EXAMPLE
##########

Sending build context to Docker daemon  12.8 kB
Step 0 : FROM atomicapp:build
 ---> 9b5c7f459155
Step 1 : MAINTAINER Red Hat, Inc. <container-tools@redhat.com>
 ---> Running in 500be4b14e05
 ---> de7fd8386682
Removing intermediate container 500be4b14e05
Step 2 : LABEL io.projectatomic.nulecule.providers "kubernetes,docker" io.projectatomic.nulecule.specversion "0.0.2"
 ---> Running in ed4f6e6cf6da
 ---> 380506138f74
Removing intermediate container ed4f6e6cf6da
Step 3 : ADD /Nulecule /Dockerfile README.md /application-entity/
 ---> ec6f278cf8af
Removing intermediate container 717c76451ad4
Step 4 : ADD /artifacts /application-entity/artifacts
 ---> 9e62c5592c23
Removing intermediate container 32921e0e69e2
Successfully built 9e62c5592c23
docker run -it --rm --privileged -v `pwd`:/atomicapp -v /run:/run -v /:/host --net=host --name helloapache -e NAME=helloapache -e IMAGE=projectatomic/helloapache projectatomic/helloapache -v run --provider=docker -a answers.conf --destination=build

INFO: --net=host 
      Processes in this container can listen to ports (and possibly rawip traffic) on the host's network.

INFO: --privileged 

2016-01-31 21:53:46,709 - __main__ - INFO - Action/Mode Selected is: run
2016-01-31 21:53:46,710 - __main__ - DEBUG - Setting app_spec based on $IMAGE env var
2016-01-31 21:53:46,710 - __main__ - DEBUG - Final parsed cmdline: run -v --provider=docker -a answers.conf --destination=build
2016-01-31 21:53:46,711 - atomicapp.nulecule.main - DEBUG - NuleculeManager init app_path: build
2016-01-31 21:53:46,711 - atomicapp.nulecule.main - DEBUG - NuleculeManager init    image: projectatomic/helloapache
2016-01-31 21:53:46,711 - __main__ - ERROR - Provided answers file doesn't exist: answers.conf
```

I have tested this against this commit as well as the one before and can confirm that this commit introduces the failure in code.

I will not be able to debug until this *late* Thursday due to traveling  as well as Config Management Camp in Belgium. Just thought to bring it up!

baude:
@cdrage How can I replicate your test? Can you give or point me to something?

cgwalters:
I'll restate as well my 2c: The bug won't be fixed until we can replace every instance of util.chk_call() back to subprocess.check_call() and assert the argument is an array.

cdrage:
@baude 
Running `atomic run projectatomic/helloapache -a answers.conf` will make it fail.

```
~/atomic  master ✔                                                                                                            2d  ⍉
▶ touch answers.conf

~/atomic  master ✗                                                                                                           2d ◒  
▶ sudo atomic run projectatomic/helloapache -a answers.conf
docker run -it --rm --privileged -v `pwd`:/atomicapp -v /run:/run -v /:/host --net=host --name helloapache -e NAME=helloapache -e IMAGE=projectatomic/helloapache projectatomic/helloapache -v run -a answers.conf

INFO: --net=host 
      Processes in this container can listen to ports (and possibly rawip traffic) on the host's network.

INFO: --privileged 

2016-02-03 13:44:49,395 - __main__ - INFO - Action/Mode Selected is: run
2016-02-03 13:44:49,396 - __main__ - DEBUG - Setting app_spec based on $IMAGE env var
2016-02-03 13:44:49,396 - __main__ - DEBUG - Final parsed cmdline: run -v -a answers.conf
2016-02-03 13:44:49,396 - atomicapp.nulecule.main - DEBUG - NuleculeManager init app_path: /host/var/lib/atomicapp/projectatomic-helloapache-d60bd72d7ee2
2016-02-03 13:44:49,396 - atomicapp.nulecule.main - DEBUG - NuleculeManager init    image: projectatomic/helloapache
2016-02-03 13:44:49,397 - __main__ - ERROR - Provided answers file doesn't exist: answers.conf
```

baude:
@cdrage can provide your answer.conf file? Im not able to replicate the same error you are reporting with mine.

dustymabe:
@rhatdan @baude we updated our scripts to not call a subshell in the docker run command. I have tested with the master version of atomic cli and this seems to work fine.

rhatdan:
Great thanks.

PR #289, opened 2016-01-19
test_migrate.sh: validate and don't use overlayfs

jlebon:
This patch does two things:

   1. Add a validation step after migration to make sure that the
      containers and images were indeed migrated.

   2. Don't actually change the docker configuration to use overlayfs,
      which currently causes issues on e.g. Fedora kernels. This has
      been the source of the instability of the PR tester. It can be
      reproduced locally by running:

    ```
# while true; do dd if=/dev/vda of=/dev/null; done &
# make test TEST_INTEGRATION=migrate
    ```

      We can uncomment out that block once the workaround for it has
      made it to the stable kernels. See the comment block for more
      details.

### DIFF
--- a/tests/integration/test_migrate.sh
+++ b/tests/integration/test_migrate.sh
@@ -36,8 +36,8 @@ if [ ! -f /etc/sysconfig/docker-storage ];then
 fi
 
 setup () {
-	${DOCKER} create --name test-migrate-1 -v /tmp fedora /bin/bash
-	${DOCKER} create --name test-migrate-2 -v /tmp busybox /bin/bash
+	CNT1=$(${DOCKER} create --name test-migrate-1 -v /tmp fedora /bin/bash)
+	CNT2=$(${DOCKER} create --name test-migrate-2 -v /tmp busybox /bin/bash)
 }
 
 cleanup () {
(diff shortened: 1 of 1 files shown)


jlebon:
Whoops, missed a `$`.

jlebon:
Thanks @rh-atomic-bot! :metal:

rhatdan:
@shishir-a412ed PTAL

rhatdan:
LGTM.

shishir-a412ed:
@jlebon @rhatdan 
LGTM. 
Just a small comment. Why do we need CNT1 and CNT2 when the containers are already named 
I can just do `docker inspect test-migrate-1 test-migrate-2`.

Shishir

jlebon:
@shishir-a412ed Yeah, that's what I had originally, but it seems like after importing, it just gives the containers new names rather than the same name. I'm guessing names are not part of the metadata that gets exported?

shishir-a412ed:
@jlebon oh yeah, that is correct. Thanks for reminding me, I need to fix that thing !
As of now, we need to stick with the IDs.

Shishir

rhatdan:
Shishir lets create a rhel7 branch for atomic and remove the migrate code so that @lsm5 can build a package without migrate, then we can add it for our atomic-1.9 release.

shishir-a412ed:
@rhatdan or we can keep the code, and just remove the CLI. if the user doesn't see it, he sure cannot use it.
Bringing `atomic migrate` back in `docker 1.10` would be easier that way.

rhatdan:
Well you would need to turn off the atomic migrate CLI and remove the content from the man pages.

I have no problem if you can do it differently on docker-1.9 then docker-1.10, as long as it does not become too hacky.

PR #292, opened 2016-01-25
use docker.AutoVersionClient to avoid API version mismatch

TomasTomecek:
400 Client Error: Bad Request ("client is newer than server (client API version: 1.21, server API version: 1.20)")

### DIFF
--- a/Atomic/Export.py
+++ b/Atomic/Export.py
@@ -11,7 +11,7 @@ from . import util
 from docker.utils import kwargs_from_env
 
 
-DOCKER_CLIENT = docker.Client(**kwargs_from_env())
+DOCKER_CLIENT = docker.AutoVersionClient(**kwargs_from_env())
 
 ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
 
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -1011,7 +1011,7 @@ def SetFunc(function):
     return customAction
 
 
-class AtomicDocker(docker.Client):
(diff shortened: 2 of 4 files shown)


cgwalters:
Going meta about versions: Are there version concerns around docker-py with this again?  We've been burned in the past by that, keep in mind that `atomic` needs to work on older distributions too.

rhatdan:
Currently this is probably not a problem, since the client and server code only work on the same machine.  

What would happen if we started supporting docker client on separate machine the docker server.   When we get authentication and authorization working.

Wouldn't we want to support docker client talking to older servers?

TomasTomecek:
It's in docker-py since 1.1: https://github.com/docker/docker-py/blob/master/docs/change_log.md#110

Logic around features available since version X is very well defined in docker-py: it won't allow you to use a feature if it's not available in docker daemon, e.g.: https://github.com/docker/docker-py/blob/9b3b0967098206efba487f9dbcfb24ac53f5bde6/docker/api/exec_api.py#L12 so atomic command would have to handle such situation when docker-py refuses to do an API call.

> Wouldn't we want to support docker client talking to older servers?

This patch will unlock this for free.

Does this answer your questions?

rhatdan:
Ok, LGTM

rhatdan:
@baude PTAL

baude:
This LGTM.  I'm wondering if you would be willing to expand on the change in the git commit description?

rhatdan:
@TomasTomecek Add some information that you added to this pull request to the checkin and I will merge.

TomasTomecek:
done

rhatdan:
LGTM

PR #293, opened 2016-01-25
Atomic/run.py: rewrap security helper strings

jlebon:
I noticed at least one spacing issue on the latest el7 build in those
outputs. This patch reworks the layout of the strings so that it's
easier to read in the source code and to make sure that there are no
spacing issues.

Text is exactly the same, other than fixing spacing and changing "hosts
network" to "host's network".


cgwalters:
:+1:

PR #294, opened 2016-01-27
atomic migrate: Add support for docker 1.10

shishir-a412ed:

### DIFF
--- a/migrate.sh
+++ b/migrate.sh
@@ -203,8 +203,8 @@ container_import(){
 	oldStorageDriver=$(sed -n '1p' $importPath/info.txt)||exit 1
 	newStorageDriver=$(docker info|grep "Storage Driver"|cut -d" " -f 3)
 
-	sed -i "s|\"Driver\":\"$oldStorageDriver\"|\"Driver\":\"$newStorageDriver\"|g" config.json	
-	sed -i "s|$oldDockerRootDir/containers/$oldNotruncContainerID|$dockerRootDir/containers/$oldNotruncContainerID|g" config.json
+	sed -i "s|\"Driver\":\"$oldStorageDriver\"|\"Driver\":\"$newStorageDriver\"|g" config.v2.json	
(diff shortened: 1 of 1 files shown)


shishir-a412ed:
Since with docker 1.10 the concept of `linkgraph.db` (file based db for managing container's information) has been removed. `Atomic migrate export` and `import` will have consistent container names.

This will also fix this issue: https://github.com/projectatomic/atomic/pull/289#issuecomment-173240231

However we would only be supporting `atomic migrate` with `docker 1.10` and higher versions.

ping @rhatdan . I tested this and it works with `docker 1.10`.

rhatdan:
Ok I think we need to remove atomic-migrate from atomic for docker-1.9 release then.

@lsm5

jlebon:
Can't we version it so that it works regardless of the docker version?

shishir-a412ed:
@jlebon we can create a versioned code, however `atomic migrate` for `< docker 1.10` would not have consistent container names. i.e https://github.com/projectatomic/atomic/pull/289#issuecomment-173240231 issue will exist for `atomic migrate ( < docker 1.10)`

jlebon:
@shishir-a412ed That doesn't sound so bad, especially since it's the status quo anyway. OTOH, taking it out might appear as taking features away for folks stuck on 1.9 for one reason or another. Would it be difficult to support both methods?

shishir-a412ed:
@jlebon it's pretty easy actually. container's `config.json` file is renamed to `config.v2.json` for `docker 1.10`.
Would just have to put a check to support both file names.

@rhatdan WDYT ?

shishir-a412ed:
@rhatdan 
As per our discussion, we will merge this PR which would keep container names consistent in `atomic migrate export` and `import`. I will open another PR to remove `atomic migrate` (CLI and man page changes). N We ll start supporting this from `docker 1.10` onwards.

rhatdan:
I think we should keep the patch separate from github and have @lsm5 carry it.  IE we build a patch to remove atomic migrate from atomic-1.8 and then we ship it without the patch when we ship atomic-1.9 and docker-1.10

rhatdan:
LGTM

PR #295, opened 2016-01-28
Bugzilla: 1284877: fixes run/gotar issue

shishir-a412ed:


shishir-a412ed:
ping @rhatdan

shishir-a412ed:
@rhatdan 
I tested this, It throws an error if we try to delete 'run/gotar' entry when it does not exist in the tarball.
To handle that we need both `2>/dev/null` (redirect the error to /dev/null) and `|| true`.

PTAL.

PR #296, opened 2016-02-01
We are not calling self.cmd_env just passing the function

rhatdan:
This actually calls the function to setup the environment.


rhatdan:
@baude PTAL

baude:
@rhatdan I left those as they were originally defined, which was as a property.  See https://github.com/projectatomic/atomic/blob/1bc38149cd1aa3ac9bab6507ec42401f27ce9df3/Atomic/atomic.py#L500 as an example.

rhatdan:
Well if this is the way it was done, before, I believe it is wrong, since it would be setting the environment to a function pointer.  I think we should merge this and test.

rhatdan:
I am merging this to fix the issues with Environment.

PR #297, opened 2016-02-02
Removal of remote inspection from docker

baude:
We stopped carrying a patch in docker that allows for a RESTFUL
probe of remote registries.  This function is being replaced
by a new package called skopeo.

The remote_inspection def was removed from the Docker.Client
class extension in atomic.py and is replaced by a def called
util.skopeo which returns the same json-based results.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -5,7 +5,6 @@ import collections
 from fnmatch import fnmatch as matches
 from docker.utils import kwargs_from_env
 import os
-
 import docker
 import selinux
 
@@ -254,3 +253,19 @@ def urllib3_disable_warnings():
             # Except only call disable-warnings if it exists
             if hasattr(urllib3, 'disable_warnings'):
                 urllib3.disable_warnings()
+
+
+def skopeo(image):
+    """
+    Performs remote inspection of an image on a registry
+    :param image:  fully qualified name
+    :return: Returns json formatted data
+    """
+
+    cmd = ['/usr/bin/skopeo', image]
+    results = subp(cmd)
+    if results.return_code is not 0:
+        raise ValueError(results.stderr)
+    else:
+        return json.loads(results.stdout)
+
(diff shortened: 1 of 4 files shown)


Atomic/util.py
```diff
@@ -254,3 +253,19 @@ def urllib3_disable_warnings():
             # Except only call disable-warnings if it exists
             if hasattr(urllib3, 'disable_warnings'):
                 urllib3.disable_warnings()
+
+
+def skopeo(image):
```
runcom: how does this translates a provided image called `busybox` to docker's `DefaultRegistries[0-N]/busybox`?

runcom: I mean does this `image` gets fully qualified? if the docker daemon on the host is configured with `--add-registry=myregistry.com:5000` and one does `atomic verify busybox` we should make sure to run `skopeo` with:

`skopeo myregistry.com:5000/busybox` <- if this fails otherwise return result from myregistry.com:5000
`skopeo docker.io/busybox`

runcom: `myregistry.com:5000` and `docker.io` (in this order) are returned by the `/info` endpoint and before calling `skopeo` we should fully qualify the _unqualified_ image got from the cli (if it is unqualified, otherwise let's just call `skopeo` with the `image` name provided by the command line)

runcom: the behavior I'm telling about was _inside_ the docker remote inspection patch, but now that we removed it, we need to move this _loop through default registries_ logic into `atomic`

baude: @runcom  In speaking with @rhatdan , we decided we would replace remote inspection with one PR and then fix verify afterwards.  So I am waiting for this PR to merge before fixing verify wrt images and fully qualified images.

runcom: @baude SGTM

runcom:
LGTM @rhatdan ping to merge

rhatdan:
LGTM

PR #300, opened 2016-02-08
Atomic/atomic.py: Fix pylint error

giuseppe:
cmd_env is not a property, remove the decorator.


rhatdan:
LGTM

rhatdan:
BTW what is a property?

rhatdan:
Reading through the following,  shows that this is not a property
https://docs.python.org/2/library/functions.html#property

Issue #302, opened 2016-02-15
atomic fails to build without dockerd running 


baude:
It appears that the addition of docker.AutoversionClient requires that dockerd be running in order to work correctly.

```
>>> c = docker.Client()
>>> c = docker.AutoVersionClient()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 337, in __init__
    super(AutoVersionClient, self).__init__(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 77, in __init__
    self._version = self._retrieve_server_version()
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 97, in _retrieve_server_version
    'Error while fetching server API version: {0}'.format(e)
docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', error(2, 'No such file or directory'))
```
Because setup.py imports atomic content, unless dockerd is running, the build will fail like so:

```
[bbaude@localhost atomic (verify_name)]$ make
/usr/bin/python setup.py build
Traceback (most recent call last):
  File "setup.py", line 5, in <module>
    import Atomic as _Atomic
  File "/home/bbaude/docker/atomic/Atomic/__init__.py", line 3, in <module>
    from .atomic import Atomic
  File "/home/bbaude/docker/atomic/Atomic/atomic.py", line 25, in <module>
    from .Export import export_docker
  File "/home/bbaude/docker/atomic/Atomic/Export.py", line 14, in <module>
    DOCKER_CLIENT = docker.AutoVersionClient(**kwargs_from_env())
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 337, in __init__
    super(AutoVersionClient, self).__init__(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 77, in __init__
    self._version = self._retrieve_server_version()
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 97, in _retrieve_server_version
    'Error while fetching server API version: {0}'.format(e)
docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', error(2, 'No such file or directory'))
Makefile:20: recipe for target 'python-build' failed
make: *** [python-build] Error 1
```

One idea would be to do a try/except on the autoversion.

[...]

PR #303, opened 2016-02-15
Atomic/client.py: Universal method for docker.Client()

baude:
All uses of docker.Client() (docker-py) now should be using
the DockerClient definition in client.py.  Any changes to the
client instantiation or function can now be changed in a
singular location.

Also, the DockerClient function has a fallback from
docker.AutoVersionClient to docker.Client using a try and
except condition.  This fixes an issue raised in:

https://github.com/projectatomic/atomic/issues/302

where atomic cannot be built due to imports when dockerd is not
running.

### DIFF
--- a/Atomic/client.py
+++ b/Atomic/client.py
@@ -1,0 +1,12 @@
+import docker
+from docker.utils import kwargs_from_env
+
+def get_docker_client():
+    """
+    Universal method to use docker.client()
+    """
+    try:
+        return docker.AutoVersionClient(**kwargs_from_env())
+
+    except docker.errors.DockerException:
+        return docker.Client(**kwargs_from_env())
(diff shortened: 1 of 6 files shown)


baude:
@TomasTomecek  Please take a look.

Atomic/client.py
```diff
@@ -0,0 +1,12 @@
+import docker
+from docker.utils import kwargs_from_env
+
+def DockerClient():
```
TomasTomecek: I would actually follow pep8 here and named it e.g. `get_docker_client`.

baude: roger that

cgwalters:
LGTM

baude:
Updated with the optional name shortening @cgwalters  provided.

cgwalters:
Ok, let's wait on the tests.

jlebon:
bot, retest this please

baude:
Ok, tests complete, please merge.

rhatdan:
LGTM

PR #304, opened 2016-02-16
Atomic/help.py: cmd_env no longer a property

baude:
When the cmd_env defition had its property decorator removed,
a single call to cmd_env in help.py was missed.


jlebon:
LGTM, thanks!

PR #305, opened 2016-02-16
disable test_migrate.sh for atomic migrate command

shishir-a412ed:

### DIFF
--- a/tests/integration/test_migrate.sh
+++ b/tests/integration/test_migrate.sh
@@ -2,6 +2,13 @@
 set -euo pipefail
 IFS=$'\n\t'
 
+#With the inclusion of this PR (https://github.com/projectatomic/atomic/pull/294)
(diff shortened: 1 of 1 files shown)


shishir-a412ed:
ping @baude @jlebon

jlebon:
Thanks @shishir-a412ed.

I guess the issue here is that atomic is so tightly coupled with docker that the tester should probably compile and install docker from scratch instead of being limited to whatever is in the fedora repos currently.

That said, I do wonder what the strategy is for atomic re. versioning? Do we not want the atomic client to work on older docker? Or are they always going to be paired this closely?

rhatdan:
Right now we are tying them together. Mainly because atomic is working with new features we are looking to get merged into docker.  Perhaps in the future we can make them less tied.

PR #306, opened 2016-02-17
Fix verify where an image is tagged more than once

baude:
In the case where an image was tagged more than once, atomic
verify would sometimes resolve the wrong fully qualified name
due to an ordering issue in one of the supporting functions.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -40,11 +40,22 @@ def convert_size(size):
     return '0B'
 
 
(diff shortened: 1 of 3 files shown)


baude:
Rebased to master

baude:
The problem with a test case for verify is it basically relies on a remote registry which will make our test structure much more complicated.

rhatdan:
Ok.  LGTM

PR #307, opened 2016-02-17
Add test for atomic verify

baude:
Added a test case to test for failure when an image has
multiple repotags and is asked to be verified by image
id.  This should result in a failure.

Also, turned verify debug off.


jlebon:
Thanks for fixing that formatting nit. Is there a new `test_verify.sh` script that's supposed to go along with this PR?

PR #311, opened 2016-03-01
Atomic diff man fix

baude:


rhatdan:
You need to remove the  	Unify error messages for no docker daemon (BZ #1300187) patch

rhatdan:
LGTM

Issue #314, opened 2016-03-04
specififying container name/ID to 'atomic top' does not work


miabbott:
On Fedora 23 Atomic Host (23.76); atomic-1.8-3.gitcc5997a.fc23.x86_64

I have an `nginix` container running with the name `my-nginx` as seen in the output of `docker ps` below.

When I try to run `atomic top` and specify the name or ID of the container, the command returns that there are no containers running.  However, running the command with no arguments is successful.

```
# atomic host status
  TIMESTAMP (UTC)         VERSION   ID             OSNAME            REFSPEC                                                        
  2016-03-03 21:29:33     23.81     78b7f5bf76     fedora-atomic     fedora-atomic:fedora-atomic/f23/x86_64/testing/docker-host     
* 2016-03-03 20:07:19     23.76     e52382cdaa     fedora-atomic     fedora-atomic:fedora-atomic/f23/x86_64/docker-host             
# rpm -q atomic
atomic-1.8-3.gitcc5997a.fc23.x86_64
# docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
552a6776cd20        nginx               "nginx -g 'daemon off"   36 minutes ago      Up 36 minutes       80/tcp, 443/tcp     my-nginx
# atomic top my-nginx
No containers running
# atomic top 552a6776cd20
No containers running
# atomic top

                                                  ATOMIC TOP
                                                  
CONTA(I)NER* (N)AME   (P)ID      (C)PU  (M)EM  (U)ID  (G)ID  CMD                                       
552a6776cd20 my-nginx 1964       0.0    0.2    0      0      nginx: master process nginx -g daemon off;
552a6776cd20 my-nginx 1999       0.0    0.1    104    107    nginx: worker process                     
```

rhatdan:
@baude PTAL

baude:
I think https://github.com/baude/atomic/tree/atomic_top_no_containers should fix it.  If @miabbott validates, I will make a PR.

miabbott:
WFM.  Tested with name, ID, and no args.

```
# docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
c0c607b267a6        nginx               "nginx -g 'daemon off"   9 seconds ago       Up 8 seconds        80/tcp, 443/tcp     my-nginx
# ./atomic top my-nginx

                                                  ATOMIC TOP
                                                  
[...]

PR #316, opened 2016-03-04
Atomic/top.py - Fix when specifying container names or ids

baude:
Atomic top was broken when you passed it valid container names
or ids.  This is now fixed plus added a simple check for validate
that the names or ids are correct.

This addresses https://github.com/projectatomic/atomic/issues/314


rhatdan:
LGTM

Issue #315, opened 2016-03-04
'atomic diff' should check the dist tag when comparing RPM versions


miabbott:
On Fedora 23, using git master:

```
# atomic/atomic --version
1.9
# git rev-parse HEAD
9f801bb4787358cf31ecddd7a45ad44f0746a1e4
```

When comparing two RHEL7 base images that have a single package difference, `atomic diff` did not report any RPM differences:

```
# atomic/atomic diff -n -r rhel7:7.2-44 rhel7:7.2-46                                                                                                         

rhel7:7.2-44 and rhel7:7.2-46 have no different RPMs
```

However, a diff of the RPMs says otherwise:

```
# docker run --rm rhel7:7.2-44 sh -c "rpm -qa | sort" > rhel7.2-44.rpms.txt
# docker run --rm rhel7:7.2-46 sh -c "rpm -qa | sort" > rhel7.2-46.rpms.txt
# diff rhel7.2-44.rpms.txt rhel7.2-46.rpms.txt
94c94
< openssl-libs-1.0.1e-51.el7_2.2.x86_64
---
> openssl-libs-1.0.1e-51.el7_2.4.x86_64
```

@baude suspected the diff code was not checking the dist tag (`el7_2.2`)

baude:
Please try https://github.com/baude/atomic/tree/atomic_diff_rpm_versions to verify fix and provide feedback.

miabbott:
That fix works for me.

```
# ./atomic diff -n -r rhel7:7.2-44 rhel7:7.2-46                                                                                                                              

rhel7:7.2-44                            | rhel7:7.2-46                           
--------------------------------------- | ---------------------------------------
Red Hat Enterprise Linux Server relea   | Red Hat Enterprise Linux Server relea  
se 7.2 (Maipo)                          | se 7.2 (Maipo)                         
--------------------------------------- | ---------------------------------------
openssl-libs-1-1.0.1e-51.el7_2.2        |                                        
                                        | openssl-libs-1-1.0.1e-51.el7_2.4       
```

baude:
https://github.com/projectatomic/atomic/pull/317 <-- fixed in

PR #317, opened 2016-03-04
Atomic/diff.py Add release to version, name, and epoch of RPM

baude:
When an RPM diff was performed between two docker objects, we
previously only store the name, version, and epoch of the RPM for
comparision.  It turns out versioning information is also done in
the release portion as well.

This addresses https://github.com/projectatomic/atomic/issues/315


rhatdan:
LGTM

PR #321, opened 2016-03-08
Improve error handling with the docker-py client

baude:
In python-docker-py-1.6, the error handling improved
significantly.  When a docker object cannot be found,
say with client.inspect(), it now throws a specific
exception called NotFound instead of the old, generic
Docker error.  We have now updated some of our functions
to use the specific error because it was 'covering' other
docker client failures such as API compatibilies.

Also added a new custom Error class to handle a common
error message we used repeatidly.  The new class is
called 'DockerObjectNotFound' and will output a
consistent messages and takes the dockerobject as input.
This is most commonly used when we verify the users'
input for validity and is different that the docker
client NotFound error.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -26,7 +26,8 @@ from .Export import export_docker
 from .Import import import_docker
 import re
 from .client import get_docker_client
-from .util import NoDockerDaemon
+from .util import NoDockerDaemon, DockerObjectNotFound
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM, except for checks failing.

jlebon:
Tests currently failing due to internal issues with the Openstack provider. I will manually rerun them once we're back up.

PR #322, opened 2016-03-09
Use fully-qualified image name in info() and verify()

baude:
We now use the fully qualified image name (if not provided)
when dealing with atomic info and verify.  This is because
skopeo requires a fq image name to do remote inspection. As
an upside, it also makes atomic a little more user friendly
as well.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -645,36 +645,65 @@ class Atomic(object):
             args.append(c)
         return " ".join(args)
 
+    def get_fq_name(self, image_info):
+        if len(image_info['RepoTags']) > 1:
+            raise ValueError("\n{} is tagged with multiple repositories. "
+                             "Please use a repository name instead.\n".format(self.image))
+        else:
+            return image_info['RepoTags'][0]
+
+    def is_iid(self, input):
+        for i in self.get_images():
+            if i['Id'].startswith(self.image):
+                return True
+        return False
+
+    def _no_such_image(self):
+        raise ValueError("Could not find any image matching '{}'"
+                         .format(self.args.image))
+
     def info(self):
         """
         Retrieve and print all LABEL information for a given image.
         """
-        def _no_such_image():
-            raise ValueError('Could not find any image matching "{}".'
-                             ''.format(self.args.image))
-
+        def _no_label():
+            raise ValueError("'{}' has no label information."
+                             .format(self.args.image))
+        # Check if the input is an image id associated with more than one
+        # repotag.  If so, error out.
+        if self.is_iid(self.image):
+            self.get_fq_name(self._inspect_image())
+        # The input is not an image id
+        else:
+            try:
+                iid = self._is_image(self.image)
+                self.image = self.get_fq_name(self._inspect_image(iid))
+            except AtomicError:
+                if self.args.force_remote_info:
+                    self.image = self.find_remote_image()
+                if self.image is None:
+                    self._no_such_image()
+        util.writeOut("Image Name: {}".format(self.image))
         inspection = None
         if not self.args.force_remote_info:
-            inspection = self._inspect_image(self.args.image)
+            inspection = self._inspect_image(self.image)
             # No such image locally, but fall back to remote
         if inspection is None:
             # Shut up pylint in case we're on a machine with upstream
             # docker-py, which lacks the remote keyword arg.
             #pylint: disable=unexpected-keyword-arg
-            inspection = util.skopeo(self.args.image)
+            inspection = util.skopeo(self.image)
             # image does not exist on any configured registry
-        if inspection is None:
-            _no_such_image()
-            # By this point, inspection cannot be "None"
         try:
             labels = inspection['Config']['Labels']
         except TypeError:  # pragma: no cover
             # Some images may not have a 'Labels' key.
-            raise ValueError('{} has no label information.'
-                             ''.format(self.args.image))
-        if labels is not None:
+            _no_label()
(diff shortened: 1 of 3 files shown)


jlebon:
bot, retest this please

jlebon:
@baude Should be good now.

Nevermind the one that's still pending, we changed it to something shorter, so GitHub is a bit confused. Future PRs will only have the shorter one.

baude:
@rhatdan this is the PR we discussed with runcom

rhatdan:
@runcom PTAL

Atomic/atomic.py
```diff
@@ -645,36 +645,62 @@ def gen_cmd(self, cargs):
             args.append(c)
         return " ".join(args)
 
+    def get_fq_name(self, image_info):
+        if len(image_info['RepoTags']) > 1:
+            raise ValueError("\n{} is tagged with multiple repositories. "
+                             "Please use a repository name instead.\n".format(self.image))
+        else:
```
runcom: does python panic if `image_info['RepoTags'][0]` cannot be accessed? also, if `get_fq_name` returns an empty string because `image_info['RepoTags'][0]` is it acceptable?

runcom: another question, get_fq_name just returns the tag for the image, why?

runcom:
I'm not sure anymore what this PR is about, the problem with fully qualifying images came from our docker which is using --add-registry=something.com so that if a user does docker pull busybox, the image is downloaded from something.com before trying docker.io.
The fully qualified image stuff is needed if we pass just busybox or fedora and we should talk to docker to gather the additional registries (as I wrote in an email). What am I missing?

Atomic/atomic.py
baude: I think the handling of the list and its index is fine. If you are worried, I can set a condition for that.   get_fq_name returns the repotag so it can be used as a fully qualified name to look up things remotely.

runcom: but how do you really fully qualify the image? I mean the registry part? like busybox and docker.io/busybox

runcom: what I don't understand is what the tag means in *fully qualify* an image; by fully qualify an image I'm referring to prepend to the image name the registry part

baude: I added a debug statement to show you what the return value is.  Hopefully that will clear things up for you.  Note the lines below that start with 'fq_name is...'

```
[bbaude@localhost atomic (skopeo_qualified)]$ sudo ./atomic info busybox
fq_name is docker.io/busybox:latest
'busybox' has no label information.

[bbaude@localhost atomic (skopeo_qualified)]$ sudo ./atomic info fedora
fq_name is docker.io/fedora:latest
'fedora' has no label information.

[bbaude@localhost atomic (skopeo_qualified)]$ sudo ./atomic info rsyslog
fq_name is registry.access.redhat.com/rhel7/rsyslog:latest
build-date: 2016-02-11T13:39:59.109718Z
RUN: docker run -d --privileged --name NAME --net=host --pid=host -v /etc/pki/rsyslog:/etc/pki/rsyslog -v /etc/rsyslog.conf:/etc/rsyslog.conf -v /etc/sysconfig/rsyslog:/etc/sysconfig/rsyslog -v /etc/rsyslog.d:/etc/rsyslog.d -v /var/log:/var/log -v /var/lib/rsyslog:/var/lib/rsyslog -v /run:/run -v /etc/machine-id:/etc/machine-id -v /etc/localtime:/etc/localtime -e IMAGE=IMAGE -e NAME=NAME --restart=always IMAGE /bin/rsyslog.sh
Name: rhel7/rsyslog
Build_Host: rcm-img-docker01.build.eng.bos.redhat.com
architecture: x86_64
vcs-type: git
vcs-ref: cc44dd261743db4185f2a1435e79ce607263f223
Version: 7.2
Architecture: x86_64
INSTALL: docker run --rm --privileged -v /:/host -e HOST=/host -e IMAGE=IMAGE -e NAME=NAME IMAGE /bin/install.sh
Release: 15
Vendor: Red Hat, Inc.
BZComponent: rsyslog-docker
Authoritative_Registry: registry.access.redhat.com
UNINSTALL: docker run --rm --privileged -v /:/host -e HOST=/host -e IMAGE=IMAGE -e NAME=NAME IMAGE /bin/uninstall.sh

```

runcom: great, but:

sudo ./atomic info rsyslog

do you already have the rsyslog image from redhat in your host? what happens if you docker rmi the image and run  the command again? how do you gather the additional registries configured in docker? (for remote inspection ofc)

rhatdan: Yes in this case we should get the names of the registries added to docker and go to them one by one.

baude: ok will do

baude: Just remember, that could take a significantly long time.  Also, in the case of info, you have to pass a switch to check remotely.  @rhatdan are you suggesting that if the image doesnt exist locally, we ignore that switch and look remotely as well?

rhatdan: No only check remotely if the user says to check remote.

runcom: keep in mind the our docker fork stores the image already fully qualified (registry+name+tag). But it could happen to build locally for instance. If info takes a switch I don't believe we should ignore the switch, just fail

baude:
ok, updated.  please review

rhatdan:
@runcom PTAL

runcom:
looks good to me - will keep testing it though

rhatdan:
Ok lets merge and fix if you find any problems.

PR #324, opened 2016-03-10
Fixes for running atomic on python3

rhatdan:
This also should replace https://github.com/projectatomic/atomic/pull/323


rhatdan:
@giuseppe @baude PTAL

baude:
lgtm, sorry for the import error.... my fault

PR #325, opened 2016-03-10
Add new test to make sure printenv passed into container correctly

rhatdan:


rhatdan:
@jlebon Thanks. Fixed

jlebon:
Seems like there's one line difference between the two causing `diff` to return false:

```
+ diff /run/atomic/test/atomic-test-5.1 /run/atomic/test/atomic-test-5.2
25d24
< _=/usr/bin/coverage
```

jlebon:
LGTM, but it needs a rebase.

PR #326, opened 2016-03-10
Set PWD environment if not currently set

rhatdan:


Atomic/atomic.py
```diff
@@ -613,6 +613,9 @@ def cmd_env(self):
         if hasattr(self.args, 'opt3') and self.args.opt3:
             os.environ['OPT3'] = self.args.opt3
 
+        if hasattr(self.args, 'PWD'):
```
baude: if *not*

baude:
With the *not* added, it works.

baude:
lgtm

PR #334, opened 2016-04-01
[RFC] System containers

giuseppe:
these patches are the current status of system containers.  There are a few things missing on Atomic Host before this feature can be used:

- skopeo
- runc
- python3-gobject-base

For an idea of how they work, I prepared a demo here: https://asciinema.org/a/40300

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -284,36 +284,38 @@ class Atomic(object):
             self._check_system_docker_image(repo, "docker://%s" % self.image, True)
         elif self.args.tar:
             temp_dir = tempfile.mkdtemp()
-            with tarfile.open(self.args.image, 'r') as t:
-                t.extractall(temp_dir)
-                manifest = ""
-                with open(os.path.join(temp_dir, "manifest.json"), 'r') as mfile:
-                    manifest = mfile.read()
-                layers = {}
-                next_layer = {}
-                top_layer = None
-                for m in json.loads(manifest):
-                    regloc, image, tag = Atomic._parse_imagename(m["RepoTags"][0])
-                    imagebranch = "ociimage/%s-%s" % (image.replace("sha256:", ""), tag)
-                    for i in m["Layers"]:
-                        layer = i.replace("/layer.tar", "")
-                        layers[layer] = os.path.join(temp_dir, i)
-                        with open(os.path.join(temp_dir, layer, "json"), 'r') as f:
-                            json_layer = json.loads(f.read())
-                            parent = json_layer.get("parent")
-                            if not parent:
-                                top_layer = layer
-                            next_layer[parent] = layer
+            try:
+                with tarfile.open(self.args.image, 'r') as t:
+                    t.extractall(temp_dir)
+                    manifest = ""
+                    with open(os.path.join(temp_dir, "manifest.json"), 'r') as mfile:
+                        manifest = mfile.read()
+                    layers = {}
+                    next_layer = {}
+                    top_layer = None
+                    for m in json.loads(manifest):
+                        regloc, image, tag = Atomic._parse_imagename(m["RepoTags"][0])
+                        imagebranch = "ociimage/%s-%s" % (image.replace("sha256:", ""), tag)
+                        for i in m["Layers"]:
+                            layer = i.replace("/layer.tar", "")
+                            layers[layer] = os.path.join(temp_dir, i)
+                            with open(os.path.join(temp_dir, layer, "json"), 'r') as f:
+                                json_layer = json.loads(f.read())
+                                parent = json_layer.get("parent")
+                                if not parent:
+                                    top_layer = layer
+                                next_layer[parent] = layer
 
-                    layers_ordered = []
-                    it = top_layer
-                    while it:
-                        layers_ordered.append(it)
-                        it = next_layer.get(it)
+                        layers_ordered = []
+                        it = top_layer
+                        while it:
+                            layers_ordered.append(it)
+                            it = next_layer.get(it)
 
-                    manifest = json.dumps({"Layers" : layers_ordered})
-                    Atomic._import_layers_into_ostree(repo, imagebranch, manifest, layers)
-            shutil.rmtree(temp_dir)
+                        manifest = json.dumps({"Layers" : layers_ordered})
+                        Atomic._import_layers_into_ostree(repo, imagebranch, manifest, layers)
+            finally:
+                shutil.rmtree(temp_dir)
         return
 
 
@@ -860,10 +862,14 @@ class Atomic(object):
 
     def _skopeo_get_layers(self, image, layers):
         temp_dir = tempfile.mkdtemp()
-        args = ['skopeo', 'layers', image] + layers
-        r = util.subp(args, cwd=temp_dir)
-        if r.return_code != 0:
-            raise IOError('Failed to fetch the manifest for: %s.' % image)
+        try:
+            args = ['skopeo', 'layers', image] + layers
+            r = util.subp(args, cwd=temp_dir)
+            if r.return_code != 0:
+                raise IOError('Failed to fetch the manifest for: %s.' % image)
+        except Exception as e:
+            shutil.rmtree(temp_dir)
+            raise e
         return temp_dir
 
     @staticmethod
@@ -926,20 +932,20 @@ class Atomic(object):
             return True
 
         layers_dir = self._skopeo_get_layers(img, missing_layers)
+        try:
+            layers = {}
+            for root, _, files in os.walk(layers_dir):
+                for f in files:
+                    if f.endswith(".tar"):
+                        layer_file = os.path.join(root, f)
+                        layer = f.replace(".tar", "")
+                        if layer in missing_layers:
+                            layers[layer] = layer_file
 
-        layers = {}
-        for root, _, files in os.walk(layers_dir):
-            for f in files:
-                if f.endswith(".tar"):
-                    layer_file = os.path.join(root, f)
-                    layer = f.replace(".tar", "")
-                    if layer in missing_layers:
-                        layers[layer] = layer_file
-
-        if (len(layers)):
-            Atomic._import_layers_into_ostree(repo, imagebranch, manifest, layers)
-
-        shutil.rmtree(layers_dir)
+            if (len(layers)):
+                Atomic._import_layers_into_ostree(repo, imagebranch, manifest, layers)
+        finally:
+            shutil.rmtree(layers_dir)
         return True
 
     @staticmethod
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -9,6 +9,8 @@ import getpass
 import argparse
 import subprocess
 import shutil
+import tempfile
+import tarfile
 
 try:
     from subprocess import DEVNULL  # pylint: disable=no-name-in-module
@@ -749,6 +751,46 @@ class Atomic(object):
         if os.path.exists("/var/system/%s.1" % self.image):
             shutil.rmtree("/var/system/%s.1" % self.image)
 
+    def _extract_docker_image(self, image, rootfs):
+        tmpdir = tempfile.mkdtemp()
+
+        with tempfile.NamedTemporaryFile(mode='w') as tmp_image:
+            save_cmd = ["docker", "save", image]
+            cmd = self.sub_env_strings(self.gen_cmd(save_cmd))
+            self.display(cmd)
+            if not self.args.display:
+                util.check_call(cmd, env=self.cmd_env(), stdout=tmp_image)
+            with tarfile.open(tmp_image.name, 'r') as t:
+                t.extractall(tmpdir)
+
+        first = None
+        child = {}
+        for i in os.listdir(tmpdir):
+            d = os.path.join(tmpdir, i)
+            if len(i) == 64 and os.path.isdir(d):
+                with open(os.path.join(d, "json"), 'r') as data:
+                    j = json.loads(data.read())
+                    parent = j.get('parent')
+                    if parent:
+                        child[parent] = i
+                    else:
+                        first = i
+
+        def extract_union(rootfs, layer):
+            with tarfile.open(os.path.join(layer, "layer.tar"), 'r') as t:
+                for it in t:
+                    if os.path.basename(it.name).startswith(".wh."):
+                        destfile = os.path.join(os.path.dirname(it.name), os.path.basename(it.name)[4:])
+                        os.unlink(os.path.join(rootfs, destfile))
+                    elif not it.isdev():
+                        t.extract(it, rootfs)
+        it = first
+        while it:
+            extract_union(rootfs, os.path.join(tmpdir, it))
+            it = child.get(it)
+
+        shutil.rmtree(tmpdir)
+
     def _do_extract_oci(self, deployment, upgrade, skip_restart=False):
         self._check_if_image_present()
 
@@ -763,13 +805,6 @@ class Atomic(object):
             destination = "/var/system/%s.%d" % (self.name, deployment)
             rootfs = os.path.join(destination, "rootfs")
             cmd = "docker export '%s' | tar --directory='%s' -xf -" % (self.image, rootfs)
-            self.display(cmd)
-            if not self.args.display:
-                os.makedirs(rootfs)
-                subprocess.check_call(cmd, shell=True)
-
-            exports = os.path.join(destination, "rootfs/exports")
-
             if not self.args.display:
                 with open(os.path.join(destination, "image"), 'w') as image:
                     image.write(self.image + "\n")
@@ -792,11 +827,31 @@ class Atomic(object):
                 else:
                     self.systemctl_command("start")
 
-        finally:
-            cmd = self.sub_env_strings(self.gen_cmd(["docker", "rm", "-f", self.image]))
-            self.writeOut(cmd)
-            if not self.args.display:
-                util.check_call(cmd, env=self.cmd_env())
+        self._extract_docker_image(self.image, os.path.join(destination, "rootfs"))
+
+        exports = os.path.join(destination, "rootfs/exports")
+
+        if not self.args.display:
+            with open(os.path.join(destination, "image"), 'w') as image:
+                image.write(self.image + "\n")
+            sym = "/var/spc/%s" % (self.name)
+            if os.path.exists(sym):
+                os.unlink(sym)
+            os.symlink(destination, sym)
+
+        shutil.copy2(os.path.join(exports, "config.json"), os.path.join(destination, "config.json"))
+
+        unitfile = os.path.join(exports, "service.template")
+        unitfileout = "/usr/local/lib/systemd/system/%s.service" % (self.name)
+        if os.path.exists(unitfile):
+            with open(unitfile, 'r') as infile, open(unitfileout, "w") as outfile:
+                data = infile.read().replace("$DESTDIR", destination).replace("$NAME", self.name)
+                outfile.write(data)
+            self.systemctl_command("enable")
+            if upgrade:
+                self.systemctl_command("restart")
+            else:
+                self.systemctl_command("start")
 
     def _install_system_container(self):
         if os.path.exists("/var/system/%s.0" % self.name):
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -732,10 +732,10 @@ class Atomic(object):
             util.check_call(cmd, env=self.cmd_env())
 
     def _system_container_exists(self, name):
-        return os.path.exists("/var/system/%s" % name)
+        return os.path.exists("/var/lib/containers/atomic/%s" % name)
 
     def _uninstall_system_container(self):
-        systemdir = os.path.realpath("/var/system/%s" % self.image)
+        systemdir = os.path.realpath("/var/lib/containers/atomic/%s" % self.name)
         service_installed = os.path.exists(os.path.join(systemdir, "rootfs/exports/service.template"))
         self.args.display = False
         if service_installed:
@@ -745,15 +745,10 @@ class Atomic(object):
         if service_installed:
             os.unlink("/usr/local/lib/systemd/system/%s.service" % (self.name))
 
-        exportedfs = os.path.join(systemdir, "rootfs/exports/rootfs/")
-        for root, _, files in os.walk(exportedfs):
-            for f in files:
-                os.unlink("/" + os.path.relpath(os.path.join(root, f), exportedfs))
-
-        os.unlink("/var/system/%s" % self.image)
-        shutil.rmtree("/var/system/%s.0" % self.image)
-        if os.path.exists("/var/system/%s.1" % self.image):
-            shutil.rmtree("/var/system/%s.1" % self.image)
+        os.unlink("/var/lib/containers/atomic/%s" % self.name)
+        shutil.rmtree("/var/lib/containers/atomic/%s.0" % self.name)
+        if os.path.exists("/var/lib/containers/atomic/%s.1" % self.name):
+            shutil.rmtree("/var/lib/containers/atomic/%s.1" % self.name)
 
     def _check_system_docker_image(self, repo, upgrade, image):
 
@@ -828,7 +823,8 @@ class Atomic(object):
         regloc, image, tag = Atomic._parse_imagename(image)
         imagebranch = "ociimage/%s-%s" % (image.replace("sha256:", ""), tag)
 
-        destination = "/var/system/%s.%d" % (name, deployment)
+        destination = "/var/lib/containers/atomic/%s.%d" % (name, deployment)
+
         self.writeOut("Extracting to %s" % destination)
 
         rootfs = os.path.join(destination, "rootfs")
@@ -859,9 +855,10 @@ class Atomic(object):
         exports = os.path.join(destination, "rootfs/exports")
 
         if not self.args.display:
-            with open(os.path.join(destination, "image"), 'w') as imgfile:
-                imgfile.write("%s\n" % image)
-            sym = "/var/system/%s" % (name)
+            with open(os.path.join(destination, "image"), 'w') as image:
+                image.write(self.image + "\n")
+            sym = "/var/lib/containers/atomic/%s" % (name)
+
             if os.path.exists(sym):
                 os.unlink(sym)
             os.symlink(destination, sym)
@@ -890,8 +887,8 @@ class Atomic(object):
 
         self._check_system_docker_image(repo, False, self.image)
 
-        if os.path.exists("/var/system/%s.0" % self.name):
-            self.writeOut("/var/system/%s.0 already present" % self.name)
+        if os.path.exists("/var/lib/containers/atomic/%s.0" % self.name):
+            self.writeOut("/var/lib/containers/atomic/%s.0 already present" % self.name)
             return
 
         return self._checkout_system_container(repo, self.name, self.image, 0, False)
@@ -902,23 +899,22 @@ class Atomic(object):
         repo = OSTree.Repo.new(Gio.File.new_for_path("/ostree/repo"))
         repo.open(None)
 
-        path = os.path.join("/var/oci", name)
+        if not self._check_system_docker_image(repo, True, self.image):
+            return False
+
+        if not self.force:
+            return
+
+        oci = os.path.join("/var/lib/containers/atomic", self.name)
         next_deployment = 0
-        if os.path.realpath(path).endswith(".0"):
+        if os.path.realpath(oci).endswith(".0"):
             next_deployment = 1
-
-        if os.path.exists("/var/oci/%s.%d" % (name, next_deployment)):
-            shutil.rmtree("/var/oci/%s.%d" % (name, next_deployment))
-
-        image = None
-        with open(os.path.join("/var/oci", name, "image"), "r") as i:
-            image = i.readline().rstrip("\n")
 
         if not self._check_system_docker_image(repo, True, self.image):
             return False
 
-        if os.path.exists("/var/system/%s.%d" % (self.name, next_deployment)):
-            shutil.rmtree("/var/system/%s.%d" % (self.name, next_deployment))
+        if os.path.exists("/var/lib/containers/atomic/%s.%d" % (name, next_deployment)):
+            shutil.rmtree("/var/lib/containers/atomic/%s.%d" % (name, next_deployment))
 
         self._checkout_system_container(repo, name, image, next_deployment, True)
 
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -38,6 +38,7 @@ from .util import NoDockerDaemon, DockerObjectNotFound
 from docker.errors import NotFound
 
 IMAGES = []
+ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
 
 def convert_size(size):
     if size > 0:
@@ -306,14 +307,25 @@ class Atomic(object):
                                     top_layer = layer
                                 next_layer[parent] = layer
 
+                        layers_map = {}
+                        enc = sys.getdefaultencoding()
+                        for k, v in layers.items():
+                            out = subprocess.check_output([ATOMIC_LIBEXEC + '/dockertar-sha256-helper',
+                                                           v], stderr=DEVNULL)
+                            layers_map[k] = out.decode(enc).replace("\n", "")
                         layers_ordered = []
+
                         it = top_layer
                         while it:
-                            layers_ordered.append(it)
+                            layers_ordered.append(layers_map[it])
                             it = next_layer.get(it)
 
                         manifest = json.dumps({"Layers" : layers_ordered})
-                        Atomic._import_layers_into_ostree(repo, imagebranch, manifest, layers)
+
+                        layers_to_import = {}
+                        for k, v in layers.items():
+                            layers_to_import[layers_map[k]] = v
+                        Atomic._import_layers_into_ostree(repo, imagebranch, manifest, layers_to_import)
             finally:
                 shutil.rmtree(temp_dir)
         return
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -745,25 +745,22 @@ class Atomic(object):
             self._prune_ostree_images()
             return
 
-        if self.args.system:
-            return self._system_images()
-
         _images = self.get_images()
-        if len(_images) == 0:
-            return
-        _max_repo, _max_tag = get_col_lengths(_images)
-        col_out = "{0:" + str(_max_repo) + "} {1:" + str(_max_tag) + \
-                  "} {2:12} {3:19} {4:10}"
-        self.writeOut(col_out.format("REPOSITORY", "TAG", "IMAGE ID",
-                                     "CREATED", "VIRTUAL SIZE"))
-        for image in self.get_images():
-            repo, tag = image["RepoTags"][0].rsplit(":", 1)
-            self.writeOut(col_out.format(self.dangling(repo) + repo,
-                                         tag, image["Id"][:12],
-                 time.strftime("%F %H:%M",
-                               time.localtime(image["Created"])),
-                 convert_size(image["VirtualSize"])))
+        if len(_images) >= 0:
+            _max_repo, _max_tag = get_col_lengths(_images)
+            col_out = "{0:" + str(_max_repo) + "} {1:" + str(_max_tag) + \
+                      "} {2:12} {3:19} {4:10}"
+            self.writeOut(col_out.format("REPOSITORY", "TAG", "IMAGE ID",
+                                         "CREATED", "VIRTUAL SIZE"))
+            for image in self.get_images():
+                repo, tag = image["RepoTags"][0].rsplit(":", 1)
+                self.writeOut(col_out.format(self.dangling(repo) + repo,
+                                             tag, image["Id"][:12],
+                                             time.strftime("%F %H:%M",
+                                            time.localtime(image["Created"])),
+                                            convert_size(image["VirtualSize"])))
 
+            return self._system_images()
 
     def _check_if_image_present(self):
         self.inspect = self._inspect_image()
@@ -799,6 +796,8 @@ class Atomic(object):
                 and len(x) != len(OSTREE_OCIIMAGE_PREFIX) + 64]
         max_column = max([len(rev) for rev in revs]) + 2
         col_out = "{0:%d} {1:64}" % max_column
+        if len(revs) == 0:
+            return
         self.writeOut(col_out.format("IMAGE", "COMMIT"))
 
         for rev in revs:
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -1167,7 +1167,7 @@ class Atomic(object):
         return True
 
     def _get_ostree_repo(self):
-        repo = OSTree.Repo.new(Gio.File.new_for_path("/ostree/repo"))
+        repo = OSTree.Repo.new(Gio.File.new_for_path(self.get_atomic_config_item(["ostree_repository"]) or "/ostree/repo"))
         repo.open(None)
         return repo
 
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -319,12 +319,14 @@ class Atomic(object):
 
     def pull_image(self):
         repo = self._get_ostree_repo()
-        if self.args.ostree:
-            self._check_system_ostree_image(repo, "ostree://%s" % self.image, True)
-        if self.args.docker:
-            self._check_system_docker_image(repo, "docker://%s" % self.image, True)
-        elif self.args.tar:
-            self._pull_docker_tar(repo, self.args.image)
+        if self.args.image.startswith("ostree://"):
+            self._check_system_ostree_image(repo, self.image, True)
+        elif self.args.image.startswith("docker://"):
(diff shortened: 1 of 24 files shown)


runcom:
Awesome :tada:

giuseppe:
bot, retest this please

atomic
```diff
@@ -273,6 +280,9 @@ if __name__ == '__main__':
     imagesp.set_defaults(func='images')
     imagesp.add_argument("--prune", action="store_true",
                          help=_("prune dangling images"))
+    imagesp.add_argument("--oci", dest="oci",
+                         action='store_true', default=False,
+                         help=_('Show OCI containers'))
```
rhatdan: Wouldn't the default show all container images?  Then we can add a selection criteria to limit the type of images.

--filter oci,docker,rkt?

What is an OCI IMage? Seems a lot like a docker image?

atomic
```diff
@@ -327,6 +337,18 @@ if __name__ == '__main__':
...
+        epilog="pull the latest specified image from a repository.")
+    pullp.add_argument("--tar", dest="tar", action="store_true",
+                       help=_("import from a Docker save'd tarball."))
+    pullp.add_argument("--docker", dest="docker", action="store_true",
+                       help=_("import from a Docker registry."))
+    pullp.add_argument("--ostree", dest="ostree", action="store_true",
+                       help=_("import from an OSTree repository."))
+    pullp.add_argument("image", help=_("image id"))
+    pullp.set_defaults(func='pull_image')
+
```
rhatdan: What is the default?  Is the default something we can set in /etc/atomic.conf?

Atomic/atomic.py
```diff
@@ -254,6 +270,80 @@ def push(self):
...
+                parent = json_layer.get("parent")
+                if not parent:
+                    top_layer = layer
+                next_layer[parent] = layer
+
+        layers_map = {}
+        enc = sys.getdefaultencoding()
+        for k, v in layers.items():
+            out = subprocess.check_output([ATOMIC_LIBEXEC + '/dockertar-sha256-helper',
+           v], stderr=DEVNULL)
```
rhatdan: Should this be overridden via ENV variable or search for it locally as well as in /usr/libexec/atomic, so that we can run tests before installing the package.

Atomic/atomic.py
```diff
@@ -254,6 +270,80 @@ def push(self):
...
+                        imagebranch = "dockerimg/%s-%s" % (image, tag)
+                        input_layers = []
+                        for name in os.listdir(temp_dir):
+                            if name == "repositories":
+                                continue
+                            input_layers.append(name + "/layer.tar")
+                        self._pull_dockertar_layers(repo, imagebranch, temp_dir, input_layers)
+            finally:
+                shutil.rmtree(temp_dir)
+        return
```
rhatdan: Should be in a separate method or probably in a separate PY file.

Atomic/atomic.py
```diff
@@ -771,6 +881,343 @@ def install(self):
...
+
+    def systemctl_command(self, cmd, name):
+        cmd = self.sub_env_strings(self.gen_cmd(["systemctl", cmd, name]))
+        self.display(cmd)
+        if not self.args.display:
+            util.check_call(cmd, env=self.cmd_env())
+
+    def _uninstall_oci_container(self, name):
+        ocidir = os.path.realpath("/var/lib/containers/atomic/%s" % name)
+        self.args.display = False
```
rhatdan: Should this be a constant defined above?

Atomic/atomic.py
```diff
@@ -771,6 +881,343 @@ def install(self):
...
+            shutil.rmtree("/var/lib/containers/atomic/%s.0" % name)
+        if os.path.exists("/var/lib/containers/atomic/%s.1" % name):
+            shutil.rmtree("/var/lib/containers/atomic/%s.1" % name)
+
+    def _prune_ostree_images(self):
+        repo = self._get_ostree_repo()
+        refs = {}
+        app_refs = []
+        prefix = "dockerimg/"
+
```
rhatdan: Where does this prefix come from?

rhatdan:
Man Pages?
Bash Completion?

Atomic/atomic.py
giuseppe: ATOMIC_LIBEXEC is already constructed looking at the environment variable `ATOMIC_LIBEXEC` as:

```
```
It is the same mechanism used for the `migrate.sh` script.  Is it fine like this?

atomic
giuseppe: I am still undecided here about the format.  Should we have --tar/--docker/--ostree or we should rather use a prefix (as we do for install --oci) like:
```
atomic pull tar:///foo/bar.tar
atomic pull ostree://foo:bar
```

runcom: I'm more for the image prefix as a way to standardize those flags

Atomic/atomic.py
giuseppe: yes, will do.

Atomic/atomic.py
giuseppe: it is the prefix for OSTree branches generated from a Docker import.  I will define it in a constant to make it clearer

atomic
giuseppe: I like this idea, just the output has a different format (as most of the Docker fields have no equivalent in an OSTree repository).  Is there anything depending on the format of the `atomic images` output?  How will we print both formats?

This is the output I have right now:

```
bash-4.3# atomic images
REPOSITORY                          TAG    IMAGE ID     CREATED             VIRTUAL SIZE
 docker.io/gscrivano/spc-helloworld latest sha256:aa510 2016-03-22 12:23    216.41 MB 
 docker.io/gscrivano/spc-flannel    latest sha256:9b5dd 2016-03-21 13:44    230.42 MB 
 docker.io/busybox                  latest sha256:47bcc 2016-03-18 18:22    1.11 MB   
 docker.io/alpine                   latest sha256:70c55 2016-03-02 17:16    4.79 MB   
bash-4.3# atomic images --oci
IMAGE                                    COMMIT                                                          
dockerimg/busybox-latest                 ff6b0e9239c9569be3ef48699519d7842975a48c98b0ba560000363006364c5e
dockerimg/gscrivano/spc-flannel-latest   00b316952aee741a2eeb62c98e7dc7a2018c6ce94b1e3cd93ba0ce12955c3cf3
```

giuseppe:
I have not added man pages and bash completion yet.  I will do this as soon as we agree on the interface

atomic
giuseppe: You convinced me.  I added a patch that uses the image prefixes

Atomic/atomic.py
rhatdan: Ok I missed that.

rhatdan:
@runcom @mrunalp @baude @cgwalters PTAL

atomic
rhatdan: I like this, but I would want path to rootfs for --oci option.

rhatdan:
Should the option be --oci or --ostree?  --oci, images should be able to be used in docker or on --ostree.  I believe I should be able to pull an oci image and then install it into docker or ontop of ostree.

giuseppe:
I am really confused on how it should be named.  I thought for a while that `--system` could be the most appropriate name, but in reality this mechanism can be used for any container, not only system containers, while `--oci` is just one of the possible ways that we can use this mechanism for.  Probably `--ostree` is the best option we have.

The format of the image, no matter how it is fetched (via Docker, Docker tarball or OSTree), is made of the rootfs+files under /exports (at the moment: `service.template`, `config.json.template`, `runtime.json.template`) that are distributed within the rootfs itself.  The only part which is related to OCI is that the files `config.json.template` and `runtime.json.template` are automatically copied from '/foo/container/rootfs/exports' to '/foo/container' (after some substitutions are made for values specified with `--set`).  It is assumed though that systemd is present; the entire life cycle of these containers is managed through systemctl.

atomic
giuseppe: there is no rootfs assigned yet for the `--oci` flag.  These are the images available in the OSTree repository, which could have or not been checked-out.  The checkout from the OSTree repository is done by `install --oci`.  There could be different containers using the same image.

rhatdan:
@giuseppe I think we need to have a phone call discussion on this CLI.  Lets bring in @runcom and @cgwalters

dockertar-sha256-helper.go
```diff
@@ -0,0 +1,32 @@
...
+func main() {
+	reader, err := os.Open(os.Args[1])
+	if err != nil {
+		os.Exit(1)
+	}
+	buf := make([]byte, 4096)
+	sha_256 := sha256.New()
+	w := gzip.NewWriter(sha_256)
+
+	for true {
```
mrunalp: This could be just 
```
for {
...
}
```

dockertar-sha256-helper.go
```diff
@@ -0,0 +1,32 @@
...
+		n, err := reader.Read(buf)
+		if err != nil && err != io.EOF {
+			panic(err)
+		}
+		if n <= 0 {
+			break
+		}
+		w.Write(buf[:n])
+	}
+	w.Close()
```
mrunalp: This could be moved up as a defer.

Atomic/atomic.py
```diff
@@ -254,6 +273,54 @@ def push(self):
...
+                json_layer = json.loads(f.read())
+                parent = json_layer.get("parent")
+                if not parent:
+                    top_layer = layer
+                next_layer[parent] = layer
+
+        layers_map = {}
+        enc = sys.getdefaultencoding()
+        for k, v in layers.items():
+            out = subprocess.check_output([ATOMIC_LIBEXEC + '/dockertar-sha256-helper',
```
runcom: can this helper be just plain python instead of embedding another bin?

runcom: or probably do we need gzip from golang?

dockertar-sha256-helper.go
```diff
@@ -0,0 +1,32 @@
...
+	if err != nil {
+		os.Exit(1)
+	}
+	buf := make([]byte, 4096)
+	sha_256 := sha256.New()
+	w := gzip.NewWriter(sha_256)
+
+	for true {
+		n, err := reader.Read(buf)
+		if err != nil && err != io.EOF {
```
mrunalp: nit: Could process the `n > 0` case first as it as suggested in the documentation.
Something like
```
if n > 0 {
  w.Write(buf[:n])
}
if err == io.EOF {
  break
}
if err != nil {
  panic(err)
}

File: dockertar-sha256-helper.go [hunk 11]
```diff
@@ -0,0 +1,32 @@
+package main
```
mrunalp: nit: You could add an empty line after `package main`

Atomic/atomic.py
```diff
@@ -771,6 +858,371 @@ def install(self):
...
+
+        for i in ["config.json", "runtime.json"]:
+            src = os.path.join(exports, i + ".template")
+
+            if os.path.exists(src):
+                with open(src, 'r') as infile, open(os.path.join(destination, i), "w") as outfile:
+                    _write_template(infile.read(), values, outfile)
+
+        unitfile = os.path.join(exports, "service.template")
+        unitfileout = os.path.join("/etc/systemd/system", "%s.service" % name)
```
mrunalp: We could probably store `/etc/systemd/system` as a constant as it is being reused.

Atomic/atomic.py
mrunalp: :+1: to keeping to to python if we can.

giuseppe: it is ugly but we need gzip from golang :(

rhatdan: Why?

giuseppe: so that given a layer in a docker save'd tarball we can compute the same checksum it has on the docker registry (the checksum found in the registry manifest is computed after the .tar file is gzipped).  Other gzip implementations generate a different file, even though the original .tar is the same.

It is the issue discussed here: https://github.com/projectatomic/skopeo/issues/11

dockertar-sha256-helper.go
```diff
@@ -0,0 +1,32 @@
...
+
+	for true {
+		n, err := reader.Read(buf)
+		if err != nil && err != io.EOF {
+			panic(err)
+		}
+		if n <= 0 {
+			break
+		}
+		w.Write(buf[:n])
```
mrunalp: Should check the return code here.

mrunalp: disregard the error check on Write comment as you're not writing to a file.

giuseppe: thanks, I have addressed these in some followup commits

giuseppe:
I went through a major refactoring of the patches, that made impossible to use fixup commits.  I have pushed the updated patches here :arrow_up:

rhatdan:
I would like to start moving this to getting it merged, since we want to ship a new version of atomic

atomic-1.10 should include fixed atomic scan, this patch, docker-latest support, Atomic handling of $PWD, atomic unlock, And a bunch of other fixes.

Atomic/atomic.py
```diff
@@ -30,6 +41,22 @@
 from docker.errors import NotFound
 
 IMAGES = []
+ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
+
+OSTREE_DOCKERIMG_PREFIX = "dockerimg/"
+SYSTEMD_UNIT_FILES_DEST = "/etc/systemd/system"
```
rhatdan: Why is this dockerimage, Could it be ociimage?

Atomic/atomic.py
```diff
@@ -735,15 +735,10 @@ def _uninstall_system_container(self):
...
-                os.unlink("/" + os.path.relpath(os.path.join(root, f), exportedfs))
-
-        os.unlink("/var/system/%s" % self.image)
-        shutil.rmtree("/var/system/%s.0" % self.image)
-        if os.path.exists("/var/system/%s.1" % self.image):
-            shutil.rmtree("/var/system/%s.1" % self.image)
+        os.unlink("/var/lib/containers/atomic/%s" % self.name)
+        shutil.rmtree("/var/lib/containers/atomic/%s.0" % self.name)
+        if os.path.exists("/var/lib/containers/atomic/%s.1" % self.name):
+            shutil.rmtree("/var/lib/containers/atomic/%s.1" % self.name)
```
rhatdan: Can we make this a constant  

CONTAINER_LIB=/var/lib/containers/atomic

runcom:
@giuseppe is this still blocked by me on adding `--insecure` in `skopeo`?

Atomic/atomic.py
```diff
@@ -690,6 +803,405 @@ def install(self):
         if not self.args.display:
             return util.check_call(cmd)
 
+    def _system_images(self):
+        repo = self._get_ostree_repo()
+
+        revs = [x for x in repo.list_refs()[1] if x.startswith(OSTREE_DOCKERIMG_PREFIX) and len(x) != len(OSTREE_DOCKERIMG_PREFIX) + 64]
+        max_column = max([len(rev) for rev in revs]) + 2
```
rhatdan: DOCKERIMG or OCIIMAGE?

atomic
```diff
@@ -501,6 +518,23 @@ if __name__ == '__main__':
                           help=_("filesystem location of image/container to "
                                  "be unmounted"))
 
+    # atomic update-container
+    updatecontainerp = subparser.add_parser(
+        "update-container", help=_("update a container"),
+        epilog="update a container.")
+    updatecontainerp.set_defaults(func='update_container')
```
rhatdan: How is this different then update? Can we just improve atomic update rather then add this command?

Atomic/atomic.py
giuseppe: this is really a Docker layer/image, and it is the prefix of the branch which points to the docker image layer, these OSTree branches will look like:
dockerimg/CHECKSUM (for each layer)
dockerimg/image-name (for the image, which has information on what layers are used)

docs/atomic-install.1.md
```diff
@@ -51,6 +53,12 @@ If --display is not specified the install command will execute.
    Use this name for creating installed content for the container.
 NAME will default to the IMAGENAME if it is not specified.
 
+**--set=NAME=VALUE**
+  Set a value that is going to be used by a system container for its configuration and can be specified multiple times.  It is used only by --system.
+
+**--system**
+  Install a system container.  System containers should be used for services that are required early by the system or that cannot run in a normal container.
+
```
rhatdan: A system container is a container that is executed out of an systemd unit file early in boot, using ocitools and runc.   No need to define this as running early in boot.

docs/atomic-uninstall.1.md
```diff
@@ -49,5 +50,8 @@ Any additional arguments will be appended to the command.
 **-n** **--name**=""
    If name is specified `atomic uninstall` will uninstall the named container from the system, otherwise it will uninstall the container images.
 
+**--system**
+   Uninstall a system container.
```
rhatdan: Is there anyways we can figure out that this is a system container?  Should we leave a bit of information around to tell us that the container is a system container so atomic uninstall will just do the right thing.  If I am running a system container and accidently do not include the --system flag, what happens?

giuseppe:
@runcom I added `--insecure` to the skopeo invocation when :http: is used but I am not sure if we have to consider tihs like a blocking issue.  Maybe I can just remove this part for now?  It is just this line: https://github.com/projectatomic/atomic/pull/334/commits/44fe083e1ec36263a1129ab5e46a5bd97cfcd878#diff-555a8731c137f41d8ec2a5b76216bdecR872

Atomic/atomic.py
rhatdan: Is this hard coded into ostree?  If I explode oci images onto ostree will it do it in dockerimage?  Just want to minimize the use of docker if possible.

Atomic/atomic.py
```diff
@@ -272,6 +360,11 @@ def set_args(self, args):
             self.spc = False
 
         try:
+            self.system = args.system
+        except:
+            self.system = False
```
rhatdan: This is not needed since this is defaulted above, Shouldn't we just pass as we have on other variable assignment.

Atomic/util.py
```diff
@@ -67,22 +67,23 @@ def image_by_name(img_name, images=None):
...
     return ReturnTuple(proc.returncode, stdout=out, stderr=err)
 
 
-def check_call(cmd, env=os.environ, stderr=None, stdout=None):
+def check_call(cmd, env=os.environ, stdin=None, stderr=None, stdout=None):
     # Make sure cmd is a list
     if not isinstance(cmd, list):
         cmd = shlex.split(cmd)
-    return subprocess.check_call(cmd, env=env, stderr=stderr, stdout=stdout)
+    return subprocess.check_call(cmd, env=env, stdin=stdin, stderr=stderr, stdout=stdout)
```
rhatdan: We should just merge this whole section separately.

docs/atomic-uninstall.1.md
giuseppe: I thought these are two different namespaces, in theory it is possible to have a Docker container and a system container with the same name.  Should we not allow to install the system container if there is already a docker container with that name?  Though the user can still create the Docker container in another way after a system container with the same name is running.  How then it will be possible to remove only the system container with `atomic` (or viceversa, the Docker container)?

runcom:
@giuseppe I still don't like that flag in skopeo cause it doesn't make sense for me for docker images.. Im really not sure, that http: sounds incompatible to me with skopeo and docker images. Skopeo does have flags for tls-verify but it's up to the discovery mechanism to tell if the image is being requested with http or https, @rhatdan wdyt?

docs/atomic-uninstall.1.md
rhatdan: I think we should do our best to prevent duplicate named containers.  As  you state we could end up with bad results when we use alternate names.  

If I uninstall a --system container, I would end up with a docker container which I could later uninstall.

If we prevent the installing of a --system container if a docker container exists, then we don't end up in this state unless a user went behind our back.  I think this is better then figuring out what happens if a user tries to remove a system container without specifying --system.

docs/atomic-update-container.1.md
```diff
@@ -0,0 +1,34 @@
+% ATOMIC(1) Atomic Man Pages
+% Giuseppe Scrivano
```
rhatdan: As stated above, I think it would be better to expand on the atomic update command

atomic update --container or something like that.

rhatdan:
>@giuseppe I still don't like that flag in skopeo cause it doesn't make sense for me for docker images.. >Im really not sure, that http: sounds incompatible to me with skopeo and docker images. Skopeo >does have flags for tls-verify but it's up to the discovery mechanism to tell if the image is being >requested with http or https, @rhatdan wdyt? 

I would rather http indicates insecure and https indicates secure and not have any special flags.

runcom:
Right, but right now skopeo don't have any clue about http vs https, skopeo handles download following the discovery logic of docker images where you don't provide an http or https prefix. Everything is auto discovered pinging the endpoint once with https and the with http. Skopeo (and docker) will always use the discovered scheme when interacting with the registry. I don't want to introduce http or https prefix in skopeo cause it's not how docker does this and since skopeo is meant for docker images it seems awkward to tell people to prefix their images. Skopeo does have flags for tls-verify as said though (same as Docker has)

docs/atomic-update-container.1.md
giuseppe: I was in doubt about that, as atomic update accepts an IMAGE as argument, but yes, I will change it to accept the CONTAINER name when --container is used.  It sounds better than another command.  I would still like to make it clear we are updating a system container, is `atomic update --container --system CONTAINER_NAME` fine?

rhatdan: Well as I said on the uninstall, I would prefer it atomic could figure this out for itself.

Atomic/atomic.py
giuseppe: no, this is only into `atomic`.  I kept two different namespaces for images that are imported directly from Docker (either registry or tarball) and those that are created directly into an OSTree commit.
In the first case we want to track every layer used to compose the image, so that it can be reused.  Each layer is mapped to an OSTree branch with the dockerimg/ prefix.

For images that are created directly as OSTree commits (as `ostree commit -b IMAGE_NAME rootfs`) there is no mapping with a Docker layer so there is no `dockerimg/` prefix.

We can "flatten" a Docker imported image by having a commit of the final rootfs, and this can be done at any stage, but doing so will lose all the advantages of tracking the Docker layers.

docs/atomic-install.1.md
giuseppe: thanks, I will change it to "A system container is a container that is executed out of an systemd unit file early in boot, using runc.", as at the moment we are not using ocitools

Atomic/atomic.py
giuseppe: thinking more of it, it should not make a difference if we call it in another way, e.g. `ociimage/`, being a Docker layer is just a detail of how we will handle it, so I am going to change it

Atomic/atomic.py
giuseppe: thanks, I will drop this

Atomic/atomic.py
giuseppe: I'll change it to OCIIMAGE

Atomic/util.py
giuseppe: I will split it into another commit

Atomic/atomic.py
giuseppe: I looked into this, but since I use `self.system` short after, I preferred to keep this as it makes the code cleaner instead of checking if `self.args` has `system` every time.  It is similar to .spc defined just above

rhatdan: No big thing, but I thought we were defaulting self.system=false in the init, so that if this throws the exception, it will still be set to false.  Which is why I suggested to just pass it.

```
          try:
            self.system = args.system
        except:
            pass
```

giuseppe:
I have addressed all the comments reported, now it is possible to do:

```
# atomic install --system gscrivano/spc-etcd
# atomic uninstall spc-etcd-system
```

One thing left to discuss:

The `atomic pull` command doesn't checkout the image to disk, but it is committed to the OSTree repository.  `atomic install --system` will do the checkout, and if the image is a Docker imported image, each layer is checked-out, while for a flat OSTree commit it is basically an `ostree checkout` from the commit.  This makes garbage collection easier as OSTree will take care of it, except a part in `atomic images --prune` for removing unused branches, and we won't have to care what containers are using what image.

giuseppe:
I fixed a few more things :arrow_up:

Atomic/atomic.py
```diff
@@ -379,6 +379,14 @@ def set_args(self, args):
...
 
+        try:
+            self.storage = args.storage
+        except:
+            self.storage = None
+
+        if not self.storage:
+            self.storage = self.get_atomic_config_item(["default_storage"])
+
         if not self.name and self.image is not None:
```
rhatdan: What happens is self.storage is still None?  Will everything still work. or will the system throw a  proper exception?

atomic.conf
```diff
@@ -3,3 +3,4 @@
 default_scanner: openscap
 default_docker: docker
 
+# default_storage: ostree
```
rhatdan: Why not make this the default?  What other choices do I have?  Should we list options in the config file?

Atomic/atomic.py
```diff
@@ -30,6 +41,22 @@
...
+OSTREE_OCIIMAGE_PREFIX = "ociimage/"
+SYSTEMD_UNIT_FILES_DEST = "/etc/systemd/system"
+SYSTEMD_UNIT_FILE_DEFAULT_TEMPLATE = """
+[Unit]
+Description=$NAME
+
+[Service]
+ExecStart=/bin/runc start $NAME
+Restart=on-failure
+WorkingDirectory=$DESTDIR
```
rhatdan: Should this be Restart=on-crash

On-Failure will continue to attempt to restart if their is a config error, while On-Crash will restart is runc crashes for some reason.

atomic
```diff
@@ -277,6 +277,13 @@ if __name__ == '__main__':
...
         help=_("preview the command that %s would execute") % sys.argv[0])
     installp.add_argument("image", help=_("container image"))
+    installp.add_argument("--system", dest="system",
+                          action='store_true', default=False,
+                          help=_('install a system container'))
+    installp.add_argument("--set", dest="setvalues",
+                          action='append',
+                          help=_("Specify a variable in the VARIABLE=VALUE "
+                                 "form for a system container"))
     installp.add_argument("args", nargs=argparse.REMAINDER,
```
rhatdan: If I call --set on a non --system container are we checking for this and exiting out?

bash/atomic
```diff
@@ -94,8 +94,13 @@ __atomic_image_repos_and_tags() {
...
 
 __atomic_image_repos_and_tags_and_ids() {
-	local images="$(__docker_q images -a --no-trunc | awk 'NR>1 { print $3; if ($1 != "<none>") { print $1; print $1":"$2 } }')"
-	COMPREPLY=( $(compgen -W "$images" -- "$cur") )
+	local images
+	if test $_system == "0"; then
+		images="$(__docker_q images -a --no-trunc | awk 'NR>1 { print $3; if ($1 != "<none>") { print $1; print $1":"$2 } }')"
+	else
+		images="$(ls -1 /var/lib/containers/atomic/ | grep -v \.[01]$)"
+	fi
```
rhatdan: This path should come from the config, although I would not hold up the merge to get it.

rhatdan:
@runcom @cgwalters @mrunalp PTAL

We could use some test cases for this also.

atomic.conf
giuseppe: this is the default when nothing is set as well.  Do you want it to be "default" in the sense of being explicit in the configuration file or also when nothing is set?
At the moment only `ostree` storage is supported.

bash/atomic
giuseppe: Ouch true, thanks for spotting it. I fixed it in a fixup commit.

Atomic/atomic.py
giuseppe: if storage is null, it will assume `ostree`

Atomic/atomic.py
giuseppe: Thanks.  Fixed as fixup patch

giuseppe:
fixed :arrow_up: the issues reported.  How do we get to install the missing dependency python3-gobject-base so we can have tests checked on the internal Jenkins?

atomic.conf
rhatdan: Fine leave it commented.

Atomic/atomic.py
rhatdan: Ok great.  Just wanted to make sure you handle this case.

rhatdan:
@giuseppe Looks like you need to refresh your pulls.  Merge some together also, would be helpful.

giuseppe:
I rebased the patches :arrow_up:

rhatdan:
Looks like we need to add ostree to our build environment.  @cgwalters how do we do that?

@mrunalp @runcom @baude @cgwalters PTAL

Atomic/atomic.py
```diff
@@ -8,6 +8,17 @@
 import getpass
 import argparse
 import subprocess
+import base64
+import shutil
+import tempfile
+import json
+import tarfile
+import stat
+import gi
```
mrunalp: This should probably be added to requirements.txt as a new dependency.

Atomic/atomic.py
```diff
@@ -8,6 +8,17 @@
...
+from string import Template
+gi.require_version('OSTree', '1.0')
+from gi.repository import Gio, GLib, OSTree
+from collections import defaultdict
```
mrunalp: I don't see usage of defaultdict, so probably remove this unused import?

Atomic/atomic.py
```diff
@@ -8,6 +8,17 @@
 import getpass
 import argparse
 import subprocess
+import base64
+import shutil
+import tempfile
+import json
```
mrunalp: We have already imported json once above in the file.

Atomic/atomic.py
```diff
@@ -8,6 +8,17 @@
 import getpass
 import argparse
 import subprocess
+import base64
```
mrunalp: unused import

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
         if not self.args.display:
             return util.check_call(cmd)
 
+    def _system_images(self):
+        repo = self._get_ostree_repo()
+
+        revs = [x for x in repo.list_refs()[1] if x.startswith(OSTREE_OCIIMAGE_PREFIX) and len(x) != len(OSTREE_OCIIMAGE_PREFIX) + 64]
```
mrunalp: Might want to spread this across multiple lines. https://www.python.org/dev/peps/pep-0008/#maximum-line-length

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+        max_column = max([len(rev) for rev in revs]) + 2
+        col_out = "{0:%d} {1:64}" % max_column
+        if len(revs) == 0:
+            return
+        self.writeOut(col_out.format("IMAGE", "COMMIT"))
```
mrunalp: We should probably rename writeOut to write_out as it is not pythonic.

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+        for i in repo.list_refs()[1]:
+            if i.startswith(OSTREE_OCIIMAGE_PREFIX):
+                if len(i) == len(OSTREE_OCIIMAGE_PREFIX) + 64:
+                    refs[i] = False
+                else:
+                    app_refs.append(i)
+
+        def visit(rev):
+            commit = repo.resolve_rev(rev, False)[1]
+            manifest = self._get_commit_metadata (repo, commit, "docker.manifest")
```
mrunalp: whitespace before ( needs to bre removed.

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+                        if layer in missing_layers:
+                            layers[layer] = layer_file
+
+            if (len(layers)):
+                self._import_layers_into_ostree(repo, imagebranch, manifest, layers)
+        finally:
+            shutil.rmtree(layers_dir)
+        return True
+
+    def _get_commit_metadata (self, repo, rev, key):
```
mrunalp: Consider using @staticmethod annotation for methods that aren't using `self` like this one. I'll try to spot and add comments for other such methods.

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+            rootfs = os.path.join("/ostree/deploy/", osname, os.path.relpath(rootfs, "/"))
+        except:
+            pass
+
+        os.makedirs(rootfs)
+        revs = []
+
+        rev = repo.resolve_rev(imagebranch, False)[1]
+
+        manifest = self._get_commit_metadata (repo, rev, "docker.manifest")
```
mrunalp: Need to remove whitespace before (

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+        return True
+
+    def _get_commit_metadata (self, repo, rev, key):
+        commit = repo.load_commit(rev)[1]
+        metadata = commit.get_child_value(0)
+        if key not in metadata.keys():
+            return None
+        return metadata[key]
+
+    def _checkout_system_container(self, repo, name, img, deployment, upgrade, values={}):
```
mrunalp: We should probably avoid a mutable default argument here `values={}`
See http://docs.python-guide.org/en/latest/writing/gotchas/

bash/atomic
```diff
@@ -641,6 +664,7 @@ _atomic() {
 		images
 		migrate
 		mount
+                pull
```
mrunalp: indentation looks off

dockertar-sha256-helper.go
```diff
@@ -0,0 +1,36 @@
...
+		if n > 0 {
+			w.Write(buf[:n])
+		}
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			panic(err)
+		}
+		w.Write(buf[:n])
```
mrunalp: Do we need to remove this line since it got moved up?

Atomic/atomic.py
```diff
@@ -261,6 +297,59 @@ def push(self):
 
                 prevstatus = status
 
+
```
mrunalp: don't need this newline

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+            visit(app)
+
+        for k, v in refs.items():
+            if not v:
+                ref = OSTree.parse_refspec(k)
+                self.writeOut("Deleting %s" % k)
+                repo.set_ref_immediate(ref[1], ref[2], None)
+        return
+
+    def _parse_imagename(self, imagename):
```
mrunalp: could be static method

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+            # image prefix
+            reg = ""
+            image = imagename
+        sep = image.find(":")
+        if sep > 0:
+            return reg, image[:sep], image[sep + 1:]
+        else:
+            return reg, image, "latest"
+
+    def _convert_to_skopeo(self, image):
```
mrunalp: could be static method

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+        i = image.replace("oci:", "")
+        if "http:" in i:
+            return ["--insecure", "docker://" + i.replace("http:", "")]
+        else:
+            return ["docker://" + i.replace("https:", "")]
+
+    def _skopeo_get_manifest(self, image):
```
mrunalp: could be static method

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+            args = ['skopeo', 'layers'] + self._convert_to_skopeo(image) + layers
+            r = util.subp(args, cwd=temp_dir)
+            if r.return_code != 0:
+                raise IOError('Failed to fetch the manifest for: %s.' % image)
+        except Exception as e:
+            shutil.rmtree(temp_dir)
+            raise e
+        return temp_dir
+
+    def _get_layers_from_manifest(self, manifest):
```
mrunalp: could be static method

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+        manifest_json = json.loads(manifest)
+        fs_layers = manifest_json.get("fsLayers")
+        if fs_layers:
+            layers = list(i["blobSum"] for i in fs_layers)
+            layers.reverse()
+        else:
+            layers = manifest_json.get("Layers")
+        return layers
+
+    def _import_layers_into_ostree(self, repo, imagebranch, manifest, layers):
```
mrunalp: could be static method

Atomic/atomic.py
```diff
@@ -696,6 +820,407 @@ def install(self):
...
+                        input_layers = []
+                        for name in os.listdir(temp_dir):
+                            if name == "repositories":
+                                continue
+                            input_layers.append(name + "/layer.tar")
+                        self._pull_dockertar_layers(repo, imagebranch, temp_dir, input_layers)
+            finally:
+                shutil.rmtree(temp_dir)
+
+    def _check_system_ostree_image(self, repo, img, upgrade):
```
mrunalp: could be static method

Atomic/atomic.py
giuseppe: I think this must be explicit as some values must be present, as $NAME and $DESTDIR, which are needed for the systemd unit file.

Atomic/atomic.py
giuseppe: adding a new commit to address this

dockertar-sha256-helper.go
giuseppe: ouch, thanks for catching this error.

giuseppe:
@mrunalp, fixed in this new version :arrow_up:

mrunalp:
@giuseppe Thanks! I think that's all the comments I had.

rhatdan:
@runcom PTAL

runcom:
The only thing I'd like to see here is some integration tests (for such a big and important feature) to not regress in the future, code LGTM otherwise, will provision a vm and manually test this either today evening or tomorrow

rhatdan:
I agree tests are critical, but not easy to generate.  If everything looks go to you when you test lets merge and put out atomic-1.10 -rc1  Want to get this out into RHEL  release, which is due tomorrow.

runcom:
:+1:

runcom:
@giuseppe is there anything missing skopeo side?

giuseppe:
@runcom, there is no --insecure yet for oci:http: yet, right?  I think it is fine to keep it as a secret feature for now, and publicize once Skopeo has `--insecure` instead of removing it from the code.  What do you think?

Also, there is no --verify yet, but that is a missing part that we will add later.

runcom:
If @rhatdan is fine merging this w/o --verify I'm fine also. I'll add --insecure to skopeo tomorrow

runcom:
Will ping you tomorrow morning for other questions

rhatdan:
We can add --verify and then just report Not Implemented yet.  Although we need to show something at the Summit.

giuseppe:
do you want me to add a patch with that functionality or it can be done later?

giuseppe:
ok, I have rebased the patches and added `pull --verify` that passes it directly to Skopeo :arrow_up: 

Does it look fine?

runcom:
tested this out, didn't spot anything, LGTM (will work on skopeo now)

giuseppe:
@rhatdan what do you think about these last changes?

Atomic/atomic.py
```diff
@@ -390,6 +390,14 @@ def set_args(self, args):
         if not self.storage:
             self.storage = self.get_atomic_config_item(["default_storage"])
 
+        try:
+            self.verify_image = args.verify
+        except:
+            self.verify_image = None
+
```
rhatdan: Should this be false and set by default? Or should we read config file to set the default.

It should definately be a boolean.

Atomic/atomic.py
```diff
@@ -390,6 +390,14 @@ def set_args(self, args):
...
+        if self.verify_image is None:
+            self.verify_image = self.get_atomic_config_item(["verify_image"])
+
```
rhatdan: Should do this when self.verify_image is initialized.

rhatdan:
Also what are we verifying?  Perhaps this should take a value.

--verify=redhat.com

Might make sense.

TO verify who you want to sign the image.

rhatdan:
Why don't we drop verify for now so we can merge the rest and cut atomic-1.10.  --verify will be in atomic-1.11.

giuseppe:
yes, I agree, I removed the commit which added `--verify`

rhatdan:
@giuseppe @runcom @mrunalp @baude 
Should we merge this and cut atomic-1.10, and continue doing bug fixes?

giuseppe:
@rhatdan I think yes.  Can we also say that it is not yet stable and the interface may break in future?

runcom:
that sounds good to me, but I'd love to stick and not break the interface (as I did with skopeo :()

baude:
@rhatdan can you give me by EOD to get a PR fix for the changes in skopeo prior to cutting the release?  This will include the v1 inspection handling and the changes in the skopeo's JSON structure.

rhatdan:
@baude I will do better then that I will give you to the end of my west coast day.

PR #340, opened 2016-04-07
Atomic/top.py Make sure dockerd is running

baude:
Using self.ping(), we now check to make sure the docker
daemon is running before executing any code to display
atomic top.

This fixes a followup bug reported in:

https://bugzilla.redhat.com/show_bug.cgi?id=1300187#c5


rhatdan:
LGTM

PR #342, opened 2016-04-07
Allow user to specify alternative docker executables.

rhatdan:
This will allow us to support docker-latest

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -15,6 +15,8 @@ ReturnTuple = collections.namedtuple('ReturnTuple',
                                      ['return_code', 'stdout', 'stderr'])
 ATOMIC_CONF = os.environ.get('ATOMIC_CONF', '/etc/atomic.conf')
 ATOMIC_CONFD = os.environ.get('ATOMIC_CONFD', '/etc/atomic.d/')
+_default_docker=None
+_default_docker_lib=None
 
 if sys.version_info[0] < 3:
     input = raw_input
@@ -215,3 +217,16 @@ def get_scanners():
             except AttributeError:
                 pass
     return scanners
+
+def default_docker():
+    global _default_docker
+    if not _default_docker:
+        atomic_config = get_atomic_config()
+        _default_docker = atomic_config.get('default_docker','docker')
+    return _default_docker
+
+def default_docker_lib():
+    global _default_docker_lib
+    if not _default_docker_lib:
+        _default_docker_lib = "/var/lib/%s" % default_docker()
+    return _default_docker_lib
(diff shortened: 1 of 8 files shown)


rhatdan:
@baude,@runcom,@cgwalters PTAL

Atomic/util.py
```diff
@@ -215,3 +215,13 @@ def get_scanners():
             except AttributeError:
                 pass
     return scanners
+
+def default_docker():
+    atomic_config = get_atomic_config()
```
cgwalters: Shouldn't this be cached?  I.e. it's a once-per-process-start thing?

Atomic/util.py
```diff
@@ -215,3 +215,13 @@ def get_scanners():
...
+    default_docker = atomic_config.get('default_docker')
+    if not default_docker:
+        default_docker = "docker"
+    return default_docker
```
runcom: should we somehow validate (like in_array(valid_dockers_array_strings) the `default_docker` string got from the config and have a nicer error? not issue otherwise

cgwalters:
```
    INSTALL_ARGS = [util.default_docker(), "run",
```

is wrong, we shouldn't be trying to load `/etc/atomic.conf` or really doing anything until the program *really* runs.  I think we'll need to move this out of the constant, and prepend it wherever we reference `INSTALL_ARGS` at runtime.

We keep having problems where simply trying to find the version at build time has side effects like talking to the host's docker daemon.

Atomic/util.py
baude: I like the idea of a self.docker_binary being in the init of atomic.

Atomic/util.py
```diff
@@ -215,3 +215,13 @@ def get_scanners():
             except AttributeError:
                 pass
     return scanners
+
+def default_docker():
+    atomic_config = get_atomic_config()
+    default_docker = atomic_config.get('default_docker')
```
baude: return atomic_config.get('default_docker', 'docker')

?

rhatdan:
Added self.docker_binary()
Removed docker being used in ARGS, and only call when needed.
We could add a verify for the docker binary, but we really want to make sure atomic runs even on a system with no docker client.

rhatdan:
I think I answered all of the comments.

rhatdan:
@baude,@runcom,@cgwalters PTAL

Also fixed bash completions to use alternate docker command.

runcom:
LGTM - just that Atomic/Import.py|Export.py aren't Classes and so we use util directly but I guess this is fine

rhatdan:
Yes we should move these to classes @shishir-a412ed

runcom:
we could have Export(Atomic) (and Import) so we'll use the binary exe directly configured in the parent class I guess

baude:
aside from the other comments, looks good to me dan

PR #343, opened 2016-04-12
Atomic/scan.py: Custom docker args

baude:
Added the ability to define custom docker args in the plugin
configuration files for things like bind mounting dirs from
the host to the scanner image.

When parsing the atomic scan json files, we can now handle
'Vulnerabilities' or 'Results' for keys.

Added a --debug switch to the base atomic command to allow for
more specifics when an unwanted exception is raised.

Fixed minor bug where if no image/container is provided, the
scanner would still execute.


rhatdan:
LGTM

PR #344, opened 2016-04-13
Rename COPYING to LICENSING

jzb:
Propose changing to LICENSING to prevent confusion.


rhatdan:
Should we rename or link?

jzb:
I think either is OK.

giuseppe:
why is it confusing?  Not that we follow it, but the GNU coding standards say the GPL license file should be in a file called `COPYING`, and every project that I know of is using this name.

Looking into my ~/src/ directory, I have:

$ find .  -maxdepth 2 -name COPYING | wc -l
64

$ find .  -maxdepth 2 -name LICENSING | wc -l
0

cgwalters:
It's really `LICENSE` that's standard in the Apache community.  For example, that's what it's called [in Kubernetes git](https://github.com/kubernetes/kubernetes/blob/master/LICENSE).

I don't see a reason to rename though - every sane tool should understand both of those.  Introducing a new `LICENSING` (not `LICENSE` seems like a really bad idea).

jzb:
I think a lot of GitHub projects have also taken the pattern of following the LICENSE thing. This was largely in response to @jberkus thinking that several repos had no license because they didn't have a LICENSE file. We could also just drop in a LICENSE file that says "Under GPL: see COPYING."

jlebon:
> We could also just drop in a LICENSE file that says "Under GPL: see COPYING." 

That works, or a symlink even (wonder if GitHub is smart enough to redirect).

> I think a lot of GitHub projects have also taken the pattern of following the LICENSE thing.

Right, that's because the creation page for repos allows you to choose a license from the dropdown, which will create it in LICENSE. I don't think it's a convention in any way.

GitHub itself says this on the [subject](https://help.github.com/articles/open-source-licensing/#where-does-the-license-live-on-my-repository):

> Most people place their license text in a file named LICENSE.txt (or LICENSE.md) in the root of the repository; here's an example from Hubot.
> 
> As long as your intention is clear, you can place your license information wherever you like. Some projects only include a descriptive line in the project's README (e.g., "This project is licensed under the terms of the MIT license."). However, to avoid any potential ambiguity, it's best to include the license file with your project as well.

cgwalters:
I am very, very strongly of the opinion that Github is broken here.  It's also called `COPYING` in [the Linux kernel](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/COPYING) and for that matter [the Git project itself](https://github.com/git/git/blob/master/COPYING).

jzb:
@cgwalters I don't disagree with you, but I also don't believe we win much by going against the tide. Anyway - your call, I am satisfied that we *do* have licensing information, which would have been Very Bad (TM).

cgwalters:
I'm not going oppose a patch to add a symlink `LICENSE -> COPYING` to pacify github.  Anyone could create a PR for that in fewer characters than it's taking me to type this comment.  But *renaming* it to me feels like basically pissing on all of the work that has come from the FSF ecosystem just because github can't be arsed to look in two places.

cgwalters:
https://github.com/projectatomic/atomic/pull/345

cgwalters:
While I can't find a reference in a brief search for this, I believe it's called `COPYING` *for a reason* - because the emphasis is on the fact that you are allowed to copy the software.

jzb:
Hey @cgwalters  - let's do the symlink or stub file then. I don't think there's any harm in having two files and I really don't want to erase COPYING if that's important to folks.

cgwalters:
Done in https://github.com/projectatomic/atomic/pull/345

cgwalters:
So apparently Github's license API only partially understands symlinks:

```
$ curl -s -H "Authorization: token x" https://api.github.com/repos/projectatomic/atomic/license|jq -r '.download_url'
https://raw.githubusercontent.com/projectatomic/atomic/master/LICENSE
$ curl https://raw.githubusercontent.com/projectatomic/atomic/master/LICENSE
COPYING
```
I.e. the download_url is just the raw symlink content.  But the base64 works:

```
$ curl -s -H "Authorization: token x" https://api.github.com/repos/projectatomic/atomic/license|jq -r .content | base64 -d |head
                  GNU LIBRARY GENERAL PUBLIC LICENSE
```

Anyways, Github is still broken here IMO.

PR #346, opened 2016-04-17
Atomic/scan.py: Catch unknown default scanner

baude:
In the case where an unknown scanner is defined as the default in
/etc/atomic.conf, we need to catch that as an error and raise it
to the user.


giuseppe:
LGTM

rhatdan:
LGTM

PR #347, opened 2016-04-18
atomic.d/openscap: Do standard compliance scan without CVEs

baude:
When conducting a compliance scan, we do not want to check CVES
as that is done by the default scan.


rhatdan:
Not sure this is the best way to handle this, Aren't we looking at scanner names per scanner container?

Will the code allow BlackDuck to have a scan called CVE?

baude:
We are looking at defined scanners (i.e. openscap, blackduck, foobar) and scan_types (cve, standards_compliance, etc).  You can have different scanners with the same scan_type.  There is no collision near as I can tell and just tested it.

rhatdan:
LGTM

PR #348, opened 2016-04-20
[merged] Atomic/atomic.py Silence pylinter

baude:
pylint was throwing a false positive error for a no-member problem.
Added the bits to silence it so make will succeed.  Verified with
Giuseppe that the code is correct.


giuseppe:
@cgwalters-bot r+  	b2fdf9d

jlebon:
@cgwalters-bot retry

PR #349, opened 2016-04-20
Atomic/scan.py SELinux prevents bind mounting

baude:
Discovered a bug that revealed SELinux was preventing a proper
bind mount for /scanout.  Added ':rw,Z' to the docker command
used to launch scanners.

Also, altered the DEBUG bool to more appropriately use the
atomic --debug bool.

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -11,7 +11,6 @@ class Scan(Atomic):
     """
     Scan class that can generically work any scanner
     """
-    DEBUG = False
     results = '/var/lib/atomic'
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM.  Only thing I am a little worried about is docker handing :Z when SELinux is disabled.

baude:
@rhatdan do you want me to conditionally check getenforce and alter the docker call accordingly ?

rhatdan:
I think it should just work. without having to hack it out.  I of course do not have an SELinux disable system to test on.

baude:
@rh-atomic-bot I just tested on a permissive system.  Appears ok.

rhatdan:
@baude Need to update the pull request.

PR #350, opened 2016-04-21
Refactor skopeo code

giuseppe:
the same function used by system containers is used by verify (retrieve a manifest through Skopeo), so I refactored this part in common in util.py.

While I was at it, I also changed verify.py to use a newer Skopeo version

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -725,13 +725,15 @@ class Atomic(object):
             # Shut up pylint in case we're on a machine with upstream
             # docker-py, which lacks the remote keyword arg.
             #pylint: disable=unexpected-keyword-arg
-            inspection = util.skopeo_inspect(self.image)
+            inspection = util.skopeo_inspect("docker://" + self.image)
             # image does not exist on any configured registry
-        try:
+        if 'Config' in inspection and 'Labels' in inspection['Config']:
             labels = inspection['Config']['Labels']
-        except TypeError:  # pragma: no cover
-            # Some images may not have a 'Labels' key.
+        elif 'Labels' in inspection:
+            labels = inspection['Labels']
+        else:
             _no_label()
+
         if labels is not None and len(labels) is not 0:
             for label in labels:
                 self.write_out('{0}: {1}'.format(label, labels[label]))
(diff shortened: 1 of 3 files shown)


giuseppe:
@baude do the changes to `verify.py` look correct to you?

baude:
@giuseppe lets discuss in irc. Im working on a PR in a similar area as well.  In fact some of the changes are almost identical.  If we go with yours, you will also need something like below to fix atomic info.

```
index 0a0d666..21b0b61 100644
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -728,10 +728,14 @@ class Atomic(object):
             inspection = util.skopeo(self.image)
             # image does not exist on any configured registry
         try:
-            labels = inspection['Config']['Labels']
-        except TypeError:  # pragma: no cover
+            labels = inspection['Labels']
+        except:
+            try:
+                labels = inspection['Config']['Labels']
+            except TypeError:  # pragma: no cover
             # Some images may not have a 'Labels' key.
-            _no_label()
+                _no_label()
+
         if labels is not None and len(labels) is not 0:
             for label in labels:
                 self.write_out('{0}: {1}'.format(label, labels[label]))

```

giuseppe:
thanks for the review, I missed that part.  Should we add it as a separate patch (and you are the author) or you prefer that I amend it?

Atomic/atomic.py
```diff
@@ -915,42 +917,26 @@ def _parse_imagename(imagename):
     def _convert_to_skopeo(image):
         i = image.replace("oci:", "")
         if "http:" in i:
-            return ["--insecure", "docker://" + i.replace("http:", "")]
+            return ["--insecure"], "docker://" + i.replace("http:", "")
```
runcom: the only code path which leads here is your system containers PR right? there's nothing in atomic 1.10 which goes here right? Asking because I haven't added that flag yet

giuseppe: yes, true, it ends up here only if users specify "http:" and that should not happen, yet.

runcom: alright, all of those new flags/syntax hidden? :smile:

baude:
@giuseppe as you wish

Atomic/atomic.py
giuseppe: yes, it is hidden for now.  I can drop it, but I wanted to make a difference between `https:` (which is the default) and `http:` if specified

giuseppe:
@baude I fixed this case in the new version, and pushed in 68dccc357ab695676782a65a0b343823b888a277

PR #351, opened 2016-04-22
[merged] Improve debug and exception tracebacks

baude:
Reworked the exception handling leveraging a final clause
to print the stacktrace.  One side effect of this is that
sys.exit() under normal considerations is still considered
an Exception and therefore you get a stacktrace for
SystemExit when --debug is enabled.

Reworked DEBUG contants to be variables defined by default
as False in the class' __init__.  Then use a def called
set_debug to set to True if --debug is True.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -143,6 +143,7 @@ class Atomic(object):
         self.active_containers = []
         self.atomic_config = None
         self.docker_cmd = None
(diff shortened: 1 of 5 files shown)


rhatdan:
@baude How about we call self.Exit() which turns off debug?

baude:
@cgwalters-bot r+

PR #353, opened 2016-04-22
[merged] Do not show scanning container stdout

baude:
Early users have suggested the stdout output from the openscap
container is too verbose and confusing.  By default, we now
do not show the container's stdout.  This default can be
overriden with --verbose or --debug.

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -85,8 +85,11 @@ class Scan(Atomic):
         # Show the command being run
         util.write_out(" ".join(scan_cmd))
 
+        # Show stdout from container if --debug or --verbose
+        stdout = None if (self.args.verbose or self.args.debug) else open(os.devnull, 'w')
+
         # do the scan
-        util.check_call(scan_cmd)
+        util.check_call(scan_cmd, stdout=stdout)
 
         # umount all the rootfs
         self._umount_rootfs_in_dir()
--- a/atomic
+++ b/atomic
@@ -438,6 +438,7 @@ if __name__ == '__main__':
     scanp.add_argument("--scanner", choices=[x['scanner_name'] for x in scanners], default=default_scanner, help=_("define the intended scanner"))
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scanner"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
+    scanp.add_argument("--verbose", action='store_false', default=True, help=_("Show more output from scanning container"))
     scan_group.add_argument("--all", default=False, action='store_true', help=_("scan all images (excluding intermediate layers) and containers"))
     scan_group.add_argument("--images", default=False, action='store_true', help=_("scan all images (excluding intermediate layers)"))
     scan_group.add_argument("--containers", default=False, action='store_true', help=_("scan all containers"))


atomic
```diff
@@ -570,18 +571,18 @@ if __name__ == '__main__':
...
+    #    sys.stderr.write("%s\n" % str(e))
+    #    sys.exit(1)
+    #except subprocess.CalledProcessError as e:
+    #    sys.stderr.write("\n")
+    #    sys.exit(e.returncode)
+    #except AttributeError:
+    #    # python3 throws exception on no args to atomic
+    #    parser.print_usage()
+    #    sys.exit(1)
     except Exception as e:
```
rhatdan: Why are these commented out?

Atomic/util.py
```diff
@@ -83,7 +83,10 @@ def subp(cmd, cwd=None):
     return ReturnTuple(proc.returncode, stdout=out, stderr=err)
 
 
-def check_call(cmd, env=os.environ, stdin=None, stderr=None, stdout=None):
+def check_call(cmd, env=os.environ, stdin=None, stderr=None, stdout=None, quiet=True):
+    if quiet:
+        devnull = open(os.devnull, 'w')
+        stdout = devnull
     # Make sure cmd is a list
```
rhatdan: Why not default stdout on check_call to devnull, unless --quiet is set

No need to pass flag.

Atomic/scan.py
```diff
@@ -73,8 +73,11 @@ def get_scan_info(scanner, scan_type):
         # Show the command being run
         util.write_out(" ".join(scan_cmd))
 
+        # Show stdout from container if --debug or --verbose
+        quiet = False if not self.args.verbose or self.args.debug else True
+
         # do the scan
-        util.check_call(scan_cmd)
+        util.check_call(scan_cmd, quiet=quiet)
```
rhatdan: util.check_call(scan_cmd, stdout=stdout)

Atomic/util.py
baude: @rhatdan util.check_call is used by several other functions in atomic.  I didnt want to override their behaviours as well.

rhatdan: I don't see this as an override.  You would keep it to None as the default and only override in the quiet case.

rhatdan: devnull = open(os.devnull, 'w')


 +        # Show stdout from container if --debug or --verbose
+        standard_out = devnull if not self.args.verbose or self.args.debug else None
+
         # do the scan
-        util.check_call(scan_cmd)
+        util.check_call(scan_cmd, stdout=standard_out)

baude:
@cgwalters-bot r+

atomic
```diff
@@ -438,6 +438,7 @@ if __name__ == '__main__':
     scanp.add_argument("--scanner", choices=[x['scanner_name'] for x in scanners], default=default_scanner, help=_("define the intended scanner"))
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scanner"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
+    scanp.add_argument("--verbose", action='store_false', default=True, help=_("Show more output from scanning container"))
     scan_group.add_argument("--all", default=False, action='store_true', help=_("scan all images (excluding intermediate layers) and containers"))
```
rhatdan: Shouldn't this be default=False?

Issue #329, opened 2016-03-28
Does it make sense to have Commissaire Functionality Added to the Atomic command?


ashcrow:
We are currently looking to find the proper place for [Commissaire/Atomic Multi-Host Manager](https://github.com/projectatomic/commissaire)'s CLI to live. Two options that came up recently were:

- Add functionality to the atomic command
- Add a pass through from the atomic command to a Commissaire specific command (similar to https://github.com/projectatomic/atomic/blob/master/atomic#L142)

Personally, I'm a little more interested in the second of the two as it would allow users to use it directly. The additional functionality added by Commissaire would be things such as upgrades, restarts, grouping of hosts (currently called clusters but the name may change since the word is overloaded), etc..

Thoughts?

/cc @rhatdan @devop-mmcgrath @cgwalters @mbarnes

ashcrow:
Spike Doc: https://github.com/projectatomic/commissaire/wiki/SPIKE:-Find-the-best-location-for-the-CLI

jlebon:
I would vote for option two as well (or leaving it as standalone), just to keep the `atomic` codebase lean.

I guess it depends on other factors as well, e.g. could it happen that `commctl` gets rewritten in a different language? I remember mentions of keeping that door open at the beginning.

Also, are there situations where we would have to support nodes which might not have `atomic` installed? If it only requires kubernetes and etcd to work, that might be a nice property to keep.

rhatdan:
Yes I think Option 2 also.

mbarnes:
On 03/28/2016 11:41 AM, Stephen Milner wrote:
> Personally, I'm a little more interested in the second of the two as it
> would allow users to use it directly. The additional functionality added
> by Commissaire would be things such as upgrades, restarts, grouping of
> hosts (currently called clusters but the name may change since the word
> is overloaded), etc..
>
> Thoughts?

Commissaire's CLI tool is just a small amount of Python code [1].  It's 
just wrapping a REST API.  Adding a 1:1 command passthrough to atomic 
like we do for rpm-ostree would be, I think, almost as much code as the 
tool itself.

What if we instead provide a Python module with an API that takes and 
fleshes out an argparse.ArgumentParser in a way that atomic expects?

atomic would just need to import the module and hand it a new subparser.

That would avoid potentially maintaining command syntax in two places, 
and the standalone Commissaire tool (if we decide to keep it) would be 
trivial.

Matthew Barnes


[1] 
[...]

PR #354, opened 2016-04-22
[merged] Add cluster subcommand

mbarnes:
This adds a `cluster` subcommand to integrate with Commissaire, as discussed in https://github.com/projectatomic/atomic/issues/329.

One thing to note about `atomic cluster` is it does **not** require root privilege; it's just talking over HTTP.  I tried to solve that cleanly in followup commits by delaying the root user check until after the command line is parsed and then adding a way for subcommands to opt out of requiring the root user.  This has the added benefit of letting non-root users at least get `--help` output for any `atomic` command.

### DIFF
--- a/atomic
+++ b/atomic
@@ -122,6 +122,15 @@ def create_parser(atomic):
     parser.add_argument('--debug', default=False, action='store_true')
     subparser = parser.add_subparsers(help=_("commands"))
 
+    # atomic cluster (if commctl.cli is available)
+    try:
+        import commctl.cli
+        clusterp = subparser.add_parser(
+            "cluster", help=_("execute cluster management commands"))
+        commctl.cli.add_cluster_commands(clusterp)
+    except ImportError:
+        pass
+
     # atomic diff
     diffp = subparser.add_parser(
     "diff", help=_("Show differences between two container images, file diff or RPMS."),
@@ -392,6 +401,15 @@ def create_parser(atomic):
     #                      default=None,
     #                      dest="org_name",
     #                      help=_("Organization Name"))
+
+    # atomic rhost (if commctl.cli is available)
+    try:
+        import commctl.cli
+        rhostp = subparser.add_parser(
+            "rhost", help=_("execute cluster host management commands"))
+        commctl.cli.add_host_commands(rhostp)
+    except ImportError:
+        pass
 
     # atomic scan
     def bool_detect(myarg):
--- a/bash/atomic
+++ b/bash/atomic
@@ -729,12 +729,138 @@ _atomic_storage_import() {
 
 }
 
+_atomic_no_opts() {
+	case "$cur" in
+		-*)
+			COMPREPLY=( $( compgen -W "$main_options_with_args" -- "$cur" ) )
+			;;
+		*)
+			;;
+	esac
+}
+
+_atomic_cluster_create() {
+	_atomic_no_opts
+}
+
+_atomic_cluster_delete() {
+	_atomic_no_opts
+}
+
+_atomic_cluster_get() {
+	_atomic_no_opts
+}
+
+_atomic_cluster_list() {
+	_atomic_no_opts
+}
+
+_atomic_cluster_deploy_start() {
+	_atomic_no_opts
+}
+
+_atomic_cluster_deploy_status() {
+	_atomic_no_opts
+}
+
+_atomic_cluster_restart_start() {
+	_atomic_no_opts
+}
+
+_atomic_cluster_restart_status() {
+	_atomic_no_opts
+}
+
+_atomic_cluster_upgrade_start() {
+	_atomic_no_opts
+}
+
+_atomic_cluster_upgrade_status() {
+	_atomic_no_opts
+}
+
+_atomic_cluster() {
+	local commands=()
+	local completions_func=_atomic_cluster_${prev}
+
+	case "$prev" in
+		cluster)
+			commands=(create delete get list deploy restart upgrade)
+			;;
+		deploy)
+			commands=(start status)
+			;;
+		restart)
+			commands=(start status)
+			;;
+		upgrade)
+			commands=(start status)
+			;;
+		start|status)
+			if [ $cword -gt 2 ]; then
+				local completions_func=_atomic_cluster_${words[$cword-2]}_${prev}
+			fi
+			;;
+	esac
+
+	case "$cur" in
+		-*)
+			declare -F $completions_func >/dev/null && $completions_func
+			;;
+		*)
+			COMPREPLY=( $( compgen -W "${commands[*]}" -- "$cur" ) )
+			;;
+	esac
+}
+
+_atomic_rhost_create() {
+	case "$cur" in
+		-*)
+			COMPREPLY=( $( compgen -W "-c --cluster $main_options_with_args" -- "$cur" ) )
+			;;
+		*)
+			;;
+	esac
+}
+
+_atomic_rhost_delete() {
+	_atomic_no_opts
+}
+
+_atomic_rhost_get() {
+	_atomic_no_opts
+}
+
+_atomic_rhost_list() {
+	_atomic_no_opts
+}
+
+_atomic_rhost() {
+	local commands=()
+	local completions_func=_atomic_rhost_${prev}
+
+	case "$prev" in
+		rhost)
+			commands=(create delete get list)
+			;;
+	esac
+
+	case "$cur" in
+		-*)
+			declare -F $completions_func >/dev/null && $completions_func
+			;;
+		*)
+			COMPREPLY=( $( compgen -W "${commands[*]}" -- "$cur" ) )
+			;;
+	esac
+}
 
 _atomic() {
 	local previous_extglob_setting=$(shopt -p extglob)
 	shopt -s extglob
 
 	local commands=(
+		cluster
 		host
 		diff
 		info
@@ -743,6 +869,7 @@ _atomic() {
 		storage
 		mount
 		pull
+		rhost
 		run
 		scan
 		stop
--- a/docs/atomic-cluster.1.md
+++ b/docs/atomic-cluster.1.md
@@ -1,0 +1,55 @@
+% ATOMIC(1) Atomic Man Pages
+% Matthew Barnes
+% May 2016
+# NAME
+atomic-cluster - Manage cluster hosts
+
+# SYNOPSIS
+**atomic cluster [OPTIONS] SUBCOMMAND**
+
+**atomic cluster** will issue commands to a Commissaire server over HTTP.
+
+Commissaire allows administrators of a Kubernetes, Atomic Enterprise or
+OpenShift installation to perform administrative tasks without the need
+to write custom scripts or manually intervene on systems.
+
+# NOTE
+The `cluster` subcommand is only available when `commctl` is installed.
+
+# OPTIONS
+**-h** **-help**
+  Print usage statement.
+
+# SUBCOMMANDS
+**create** CLUSTER_NAME
+  Register a new cluster with the specified name.
+
+**delete** CLUSTER_NAME
+  Unregister the specified cluster and disassociate all of its hosts.
+
+**get** CLUSTER_NAME
+  Report status of the specified cluster.
+
+**list**
+  List all available clusters by name.
+
+**deploy start** CLUSTER_NAME VERSION
+  Initiate deployment of a tree image with a version tag matching VERSION on all hosts associated with the specified cluster.
+
+**deploy status** CLUSTER_NAME
+  Report status of a deployment operation on the specified cluster.
+
+**restart start** CLUSTER_NAME
+  Initiate a restart of all hosts associated with the specified cluster.
+
+**restart status** CLUSTER_NAME
+  Report status of a restart operation on the specified cluster.
+
+**upgrade start** CLUSTER_NAME
+  Initiate an upgrade of all hosts associated with the specified cluster.
+
+**upgrade status** CLUSTER_NAME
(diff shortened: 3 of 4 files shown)


mbarnes:
CC @ashcrow: Extra pair of eyeballs wouldn't hurt, if you're interested.

ashcrow:
@mbarnes looks good to me!

bash/atomic
```diff
@@ -651,12 +651,109 @@ _atomic_migrate_import() {
 
 }
 
+_atomic_cluster_no_opts() {
```
rhatdan: Can you create this in Atomic/cluster.py?

atomic
```diff
@@ -119,6 +116,15 @@ if __name__ == '__main__':
     parser.add_argument('--debug', default=False, action='store_true')
     subparser = parser.add_subparsers(help=_("commands"))
 
+    # atomic cluster (if available)
+    try:
+        import commctl.cli
+        clusterp = subparser.add_parser(
```
rhatdan: Not crazy on this, since I think we should just require it.  Having it be optional makes us shipping the man page seem bad.  Remove the try block.

rhatdan:
How about this patch for running atomic as non root.

https://github.com/projectatomic/atomic/pull/355

mbarnes:
> How about this patch for running atomic as non root.
> #355

Looks like it already got merged, but I'm happy with it too.  I'll rebase this branch.

atomic
mbarnes: No problem.  For the record, though, I was following the example of `atomic host` which is also written to be optional.

rhatdan: Yes that is true, but that is more of a case where we are differentiating from normal OS and Atomic Host.

This cluster command will be available on the normal OS correct?  IE RHEL as well as RHEL atomic host?

docs/atomic-cluster.1.md
```diff
@@ -0,0 +1,65 @@
...
+The `cluster` subcommand is only available when the Commissaire client
+module `commctl` is installed.
+
+# OPTIONS
+**-h** **-help**
+  Print usage statement.
+
+# SUBCOMMANDS
+**get cluster** CLUSTER_NAME
+  Report status of the specified cluster.
```
rhatdan: Shouldn't this be just get not get cluster

atomic cluster get CLUSTER_NAME

docs/atomic-cluster.1.md
```diff
@@ -0,0 +1,65 @@
...
+
+**get host** IP_ADDRESS
+  Report status of the host with the specified IP address.
```
rhatdan: DITTO

docs/atomic-cluster.1.md
```diff
@@ -0,0 +1,65 @@
...
+
+**get restart** CLUSTER_NAME
+  Report status of a restart operation on the specified cluster.
```
rhatdan: DITTO

docs/atomic-cluster.1.md
```diff
@@ -0,0 +1,65 @@
...
+
+**get upgrade** CLUSTER_NAME
+  Report status of an upgrade operation on the specified cluster.
```
rhatdan: DITTO

docs/atomic-cluster.1.md
```diff
@@ -0,0 +1,65 @@
...
+
+**create cluster** CLUSTER_NAME
+  Register a new cluster with the specified name.
```
rhatdan: DITTO

docs/atomic-cluster.1.md
```diff
@@ -0,0 +1,65 @@
...
+
+**create host** [**-c**/**--cluster** CLUSTER_NAME] IP_ADDRESS SSH_PRIV_KEY
+  Register a new host with the specified IP address.  SSH_PRIV_KEY is a file path to the host's private SSH key (e.g. id_rsa), used to drive operations on the host.  Use **--cluster** to add the host to an existing cluster.
```
rhatdan: DITTO

bash/atomic
mbarnes: @rhatdan I can, but are you sure?  Nothing else under `Atomic/` appears to be serving bash completion.  Just wanted to double check first.

atomic
mbarnes: Correct, it will be on RHEL and RHEL Atomic Host.

docs/atomic-cluster.1.md
mbarnes: Hmm... I agree that particular case reads better.  But there's some other types of things you can "get" (or "create" or "delete") besides cluster objects and I'm worried it would make parsing tricky -- or even ambiguous if we add more types of things to "get" in the future.

An easier solution is to rename the `atomic cluster` part, though the best I can think of is `atomic multihost`.

@ashcrow: Any thoughts here?

bash/atomic
rhatdan: Never mind, I though this was part of the atomic command not the bash completions.

docs/atomic-cluster.1.md
ashcrow: There are other options to ```commctl```'s```cluster``` command (```get```, ```create```, ```delete```). However, seeing ```cluster get cluster $NAME``` does seem odd.

A few ideas:

* Maybe it would make better sense to use a different direct command than ```cluster```. Maybe ```manage```?
* Maybe we can flatten out the commands we add so that they are at a higher level like:

```
atomic cluster get $NAME
atomic host delete $NAME
atomic status # maybe this one is a special case where we say atomic manager status
...
```

rhatdan: Perhaps we should setup a bluejeans to discuss this.  I really do not understand what Commissaire does, so tough for me to give opinion.

mbarnes: >   Maybe we can flatten out the commands we add so that they are at a higher level like:

> atomic cluster get $NAME
> atomic host delete $NAME
> atomic status # maybe this one is a special case where we say atomic manager status

That's going to trample on the `atomic` namespace too much I think.  `atomic host`, for example, is already a thing (wraps `rpm-ostree`).  I'm on board with your first suggestion though, but picking a good name is hard.

One thing that maybe wasn't stated explicitly enough in https://github.com/projectatomic/atomic/issues/329 was that we already have a standalone command-line tool for this named `commctl` and it shares the same command syntax (and code) with what we're proposing for `atomic cluster`.

IOW, `atomic cluster get cluster` == `commctl get cluster`.

But it's not clear to me (or even @ashcrow I think) whether `commctl` is meant to be kept if/when this stuff lands in `atomic`.

mbarnes:
@rhatdan, @ashcrow: I know we agreed on a syntax when we chatted, but the `atomic host` thing still isn't sitting right with me and I think I know why now.

Currently `atomic host` commands all act on the *local* host through `rpm-ostree`, whereas these new commands all act on *remote* hosts through Commissaire.

I think conflating local vs remote actions into the same `atomic host` prefix will be confusing to users.

@ashcrow and I were tossing around more ideas and we hit upon making it more explicit that a subset of `atomic` commands apply to *remote* hosts.

What do you think about these?
```
atomic remote cluster get|create|delete|list
atomic remote cluster reboot|upgrade start|status
atomic remote host get|create|delete|list
```
(@ashcrow, I took some liberties here vs what's in `commctl`)

cgwalters:
Seems sane to me.  On a semi-related topic, Cockpit ships `remotectl` which I always thought was clever, but AFAIK it was never used for anything more than the cockpit cert.

rhatdan:
Is the cluster remote  Or just the host?
Aren't I just joining the cluster.

Not crazy about 4 commands.  
atomic remote host delete foobar?

mbarnes:
> Is the cluster remote Or just the host?
> Aren't I just joining the cluster.

You *can* add hosts to a cluster manually, but I believe the intent is for a cluster node to add itself on boot via cloud-init (which just uses Python libraries) and reserve the `commctl`/`atomic whatever` command for admins on a machine *outside* the cluster. (@ashcrow, right?)

> Not crazy about 4 commands.

> atomic remote host delete foobar?

Yeah, it's a little wordy.  Would hyphenating help?
```
atomic remote-host delete foobar
```

rhatdan:
How about 
atomic rhost delete foobar
atomic cluster get foobar

ashcrow:
rhost sounds good to me.

mbarnes:
I'm good with that too.  It will require a bit of rework on the `commctl` side, so I'll do that first.

rhatdan:
Any update?

mbarnes:
Not quite yet.  Got side tracked, but I'm back on this now.

mbarnes:
@rhatdan Before I start on docs / bash completion updates here, can you take a look at https://github.com/projectatomic/commctl/pull/12?  That shows the new syntax for `commctl`.

`atomic` would be identical except `rhost` instead of `host`.

rhatdan:
That looks good to me, except for the `commctl passhash`, Are you going to make a atomic equivalent?

mbarnes:
> That looks good to me, except for the commctl passhash, Are you going to make a atomic equivalent?

That one seems like a poor fit for `atomic` to me.  It's just a helper script.  I'm inclined to omit it for now and just leave it in `commctl`.  If `commctl` ends up going away then maybe reevaluate.

Sound alright?

rhatdan:
Fine with me.  Your the boss.

ashcrow:
I'm fine with keeping ```passhash``` in ```commctl``` only.

mbarnes:
:arrow_up: Complete rewrite works with latest `commctl`

 (which the Jenkins environment appears to be missing)

mbarnes:
Rebased and updated bash completion + manpages for a new Commissaire operation coming soon.  (https://github.com/projectatomic/commissaire/pull/135)

I don't know how to correct for the Jenkins failure.  The `commctl` package this feature needs is either outdated or missing in the test environment.

rhatdan:
@jlebon How do we get the latest version of commictl into the test environment?

jlebon:
Shouldn't the import and cli parsers be conditional on commissaire being present? E.g. like we did for the recent system containers feature and ostree.

atomic
jlebon: I don't follow this, why should we require it? Atomic on its own is useful without commissaire (e.g. I like using atomic on my workstation, but I'm not planning to join a cluster anytime soon).

mbarnes:
> Shouldn't the import and cli parsers be conditional on commissaire being present? E.g. like we did for the recent system containers feature and ostree.

I had it conditional at first and some discussion ensued.  See outdated patch review comments above from Apr 23 (can't seem to link to them directly...)

jlebon:
Yup, commented there as well, but you're right, the [link](https://github.com/projectatomic/atomic/pull/354#discussion_r65389095) to it doesn't actually bring you to the discussion. Reposting here to make it easier:

> I don't follow this, why should we require it? Atomic on its own is useful without commissaire (e.g. I like using atomic on my workstation, but I'm not planning to join a cluster anytime soon).

rhatdan:
I am fine with going back to making it optional, just not crazy about man pages being available but functionality not.  But we set the president of ostree, so I guess we can follow it here.

cgwalters:
Talking about this in terms of Jenkins is mostly a red herring - there's two types of dependencies:

 - Build time
 - Runtime

(And each type can be either hard or soft/optional)

The current internal Jenkins PR tester conflates these, but the rpmdistro-gitoverlay builder does not, it uses RPM semantics.  (On the other hand it doesn't yet run per-project `make check`)

One major pain point specific to this project is the way we extract the version number from the python code is that it turns all hard runtime dependencies into build-time too.

Philosophy aside it does seem like having the dep be soft/optional at runtime makes sense.

mbarnes:
:arrow_up:  Jenkins is happy again.

We could revisit the optional aspect once `commctl` is available where we need it.

docs/atomic-cluster.1.md
```diff
@@ -0,0 +1,52 @@
+% ATOMIC(1) Atomic Man Pages
+% Matthew Barnes
+% May 2016
+# NAME
+atomic-cluster - Manage cluster hosts
+
+# SYNOPSIS
+**atomic cluster [OPTIONS] SUBCOMMAND**
+
+**atomic cluster** will issue commands to a Commissaire server over HTTP.
```
jlebon: Maybe a small line here to say that it will only be available if Commissaire is installed?

docs/atomic-rhost.1.md
```diff
@@ -0,0 +1,34 @@
+% ATOMIC(1) Atomic Man Pages
+% Matthew Barnes
+% May 2016
+# NAME
+atomic-rhost - Manage cluster hosts
+
+# SYNOPSIS
+**atomic rhost [OPTIONS] SUBCOMMAND**
+
+**atomic rhost** will issue commands to a Commissaire server over HTTP.
```
jlebon: And here

docs/atomic-cluster.1.md
mbarnes: I'm not sure where manpages get packaged for AH these days but my thought was if we could temporarily delete these two pages in a spec file or whatever, that seems cleaner to me than a disclaimer.  Failing that, yeah I can add a line.

rhatdan: They are packaged up in the RHELtools container

rhatdan:
What is the latest on this pull request?

ashcrow:
I'm :+1: on @jlebon's suggestions. @mbarnes can you rebase?

mbarnes:
:arrow_up: I added the man page disclaimers @jlebon suggested, rebased on current `master` and squashed my previous `fixup!` comments.  Should be good to go.

jlebon:
:+1:
@rh-atomic-bot r+ 2581030

PR #355, opened 2016-04-23
[merged] Change atomic to allow it to be run as non root

rhatdan:


cgwalters:
I would have liked some sort of rationale in the commit message itself, something like "A future addition will be passthrough for some commands like commctl which do not require root privilege".

@cgwalters-bot r+ d545897

rhatdan:
I also want atomic pull to work in certain situations.  Plus if you do setup docker socket to be available to members of the docker group,  atomic command will work without privs.

PR #358, opened 2016-04-25
[merged] Test that the atomic command is handling params properly

rhatdan:
Want to make sure that we do not regress the handling of test params.

### DIFF
--- a/tests/integration/test_params.sh
+++ b/tests/integration/test_params.sh
@@ -1,0 +1,38 @@
+#!/bin/bash -x
+set -euo pipefail
+IFS=$'\n\t'
+
+test() {
+    expected=$(for i in $@; do echo $i; done)
+    output=$(${ATOMIC} install atomic-test-6 $@ | grep -v docker)
+
+if [[ "${expected}" != "${output}" ]]; then
+    # Test failed
+    echo "Test Failed"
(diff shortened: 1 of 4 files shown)


tests/integration/test_params.sh
```diff
@@ -0,0 +1,38 @@
...
+#          SECRET     (a generated sha256 hash inserted into test containers)
+
+# In addition, the test harness creates some images for use in testing.
+#   See tests/test-images/
+
+OUTPUT=$(/bin/true)
+
+rc=0
+# Test different values
+test jezec 'asdf jkl'
```
baude: do you think we should be testing non-ascii characters?

rhatdan: Yes  for internationalization.

adelton: Could you make it into
```
test ježek 'asdf jkl'
```

then?

rhatdan: Yes, but could you tell us the proper way to fix this.

rhatdan:
@baude PTAL

rhatdan:
Since this test is now passing with @baude latest patch I will merge.

rhatdan:
@cgwalters-bot r+

PR #359, opened 2016-04-25
atomic: Handle subcommands that require root auth

baude:
Certain subcommands like diff and scan require root auth
to work correctly.  Added an except clause to catch the
permission denied errors.


rhatdan:
LGTM

PR #360, opened 2016-04-25
Atomic/atomic.py: Fix def update()

baude:
When self.args.container was not defined, the update() def
was failing.  Fixed the conditional to check if self.args.container
exists first.

This failure would make an atomic install <image> fail when the
image was not locally present.


rhatdan:
LGTM

PR #361, opened 2016-04-26
Atomic/scan.py, atomic: Fix scan related errors

baude:
When atomic scan needs to create the output directories, it needs to
do this recursively in case they don't already exist.  Changed from
os.mkdir to os.makedirs to handle this.


rhatdan:
LGTM, although you are sneaking in bugfixes.

PR #362, opened 2016-04-27
atomic: Fix NameError when calling with --help

baude:
When calling atomic or atomic <subcommand> with the
--help switch, we expose a NameError exception because
args is not evidently parsed for displaying of help. I
added a conditional to make sure the args variable exists
before using it.


giuseppe:
LGTM

rhatdan:
LGTM

PR #364, opened 2016-04-27
python3-pylint blows up on this undefined variable

rhatdan:
The check takes care of it, but python3-pylint is not smart enough to
catch this.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -21,7 +21,7 @@ _default_docker=None
 _default_docker_lib=None
 
 if sys.version_info[0] < 3:
-    input = raw_input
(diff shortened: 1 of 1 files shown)


rhatdan:
@lsm5 I don't think we are running the python3-pylint during our build or this would blow it up.

For python3 we should be executing

make PYTHON=python3 PYLINT=python3-pylint install

I think we should make this the default going forward and force RHEL to use 

make PYTHON=python3 PYLINT=pylint install

rhatdan:
@runcom @cgwalters @giuseppe PTAL

runcom:
:+1:

PR #365, opened 2016-04-27
[merged] Added atomic storage to handle resetting of storage

rhatdan:
Also moved atomic migrate to atomic storage.

Currently we support
atomic storage export
atomic storage import
atomic storage reset

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -28,8 +28,6 @@ import requests
 from . import util
 from . import satellite
 from . import pulp
-from .Export import export_docker
-from .Import import import_docker
 import re
 from .client import get_docker_client
 from .util import NoDockerDaemon, DockerObjectNotFound
@@ -204,19 +202,6 @@ class Atomic(object):
             prevstatus = status
         self.write_out("")
 
-    def Export(self):
-        try:
-            export_docker(self.args.graph, self.args.export_location)
-        except requests.exceptions.ConnectionError:
-            raise NoDockerDaemon()
-
-    def Import(self):
-        self.ping()
-        try:
-            import_docker(self.args.graph, self.args.import_location)
-        except requests.exceptions.ConnectionError:
-            raise NoDockerDaemon()
-
     def push(self):
         self.ping()
         prevstatus = ""
@@ -338,11 +323,11 @@ class Atomic(object):
             self._check_system_oci_image(repo, image, upgrade)
 
     def pull_image(self):
-        if self.storage == "ostree":
+        if self.backend == "ostree":
             repo = self._get_ostree_repo()
             self._pull_image_to_ostree(repo, self.args.image, True)
         else:
-            raise ValueError("Destination not known, please choose --storage=ostree")
+            raise ValueError("Destination not known, please choose --backend=ostree")
         return
 
     def set_args(self, args):
@@ -382,12 +367,12 @@ class Atomic(object):
             pass
 
         try:
-            self.storage = args.storage
+            self.backend = args.backend
         except:
-            self.storage = None
+            self.backend = None
 
-        if not self.storage:
-            self.storage = self.get_atomic_config_item(["default_storage"])
+        if not self.backend:
+            self.backend = self.get_atomic_config_item(["default_storage"])
 
         if not self.name and self.image is not None:
             self.name = self.image.split("/")[-1].split(":")[0]
--- a/Atomic/storage.py
+++ b/Atomic/storage.py
@@ -1,0 +1,52 @@
+import errno
+import shutil
+import selinux
+import requests
+import os, sys
+
+from . import util
+from .Export import export_docker
+from .Import import import_docker
+from .util import NoDockerDaemon
+
+try:
+    from subprocess import DEVNULL  # pylint: disable=no-name-in-module
+except ImportError:
+    DEVNULL = open(os.devnull, 'wb')
+
+try:
+    from . import Atomic
+except ImportError:
+    from atomic import Atomic
+
+class Storage(Atomic):
+    def reset(self):
+        root = "/var/lib/docker"
+        try:
+            self.d.info()
+            raise ValueError("Docker daemon must be stop before resetting storage")
+        except requests.exceptions.ConnectionError as e:
+            pass
+
+        util.check_call(["docker-storage-setup", "--reset"], stdout=DEVNULL)
+        util.call(["umount", root + "/devicemapper"], stderr=DEVNULL)
+        util.call(["umount", root + "/overlay"], stderr=DEVNULL)
+        shutil.rmtree(root)
+        os.mkdir(root)
+        try:
+            selinux.restorecon(root.encode("utf-8"))
+        except:
+            selinux.restorecon(root)
+
+    def Export(self):
+        try:
+            export_docker(self.args.graph, self.args.export_location, self.force)
+        except requests.exceptions.ConnectionError:
+            raise NoDockerDaemon()
+
+    def Import(self):
+        self.ping()
+        try:
+            import_docker(self.args.graph, self.args.import_location)
+        except requests.exceptions.ConnectionError:
+            raise NoDockerDaemon()
--- a/bash/atomic
+++ b/bash/atomic
@@ -600,19 +600,20 @@ _atomic_host() {
 	return 0
 }
 
-_atomic_migrate() {
+_atomic_storage() {
 	local commands=(
 		export
 		import
+		reset
 	)
 
-	local completions_func=_atomic_migrate_${prev}
+	local completions_func=_atomic_storage_${prev}
         declare -F $completions_func >/dev/null && $completions_func
 
 	return 0
 }
 
-_atomic_migrate_migrate() {
+_atomic_storage_storage() {
         local boolean_options="
                 --help -h
         "
@@ -633,7 +634,7 @@ _atomic_migrate_migrate() {
 }
 
 
-_atomic_migrate_export() {
+_atomic_storage_export() {
 	case "$cur" in
                 -*)
                         COMPREPLY=( $( compgen -W "--graph --dir -h --help" -- "$cur" ) )
@@ -643,7 +644,7 @@ _atomic_migrate_export() {
         esac
 }
 
-_atomic_migrate_import() {
+_atomic_storage_import() {
 	case "$cur" in
                 -*)
                         COMPREPLY=( $( compgen -W "--graph --dir -h --help" -- "$cur" ) )
@@ -665,7 +666,7 @@ _atomic() {
 		info
 		install
 		images
-		migrate
+		storage
 		mount
 		pull
 		run
--- a/Atomic/Export.py
+++ b/Atomic/Export.py
@@ -9,7 +9,7 @@ from . import util
 
 ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
 
-def export_docker(graph, export_location):
+def export_docker(graph, export_location, force):
     """
     This is a wrapper function for exporting docker images, containers
     and volumes.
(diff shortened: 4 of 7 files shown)


rhatdan:
@shishir-a412ed PTAL

rhatdan:
Anyone have tests machines they want to play with this on.  Really looking for machines without loopback devices.

@rhvgoyal @runcom @baude PTAL

Atomic/atomic.py
```diff
@@ -384,12 +369,12 @@ def set_args(self, args):
...
-            self.storage = args.storage
+            self.backend = args.backend
         except:
-            self.storage = None
+            self.backend = None
 
-        if not self.storage:
-            self.storage = self.get_atomic_config_item(["default_storage"])
+        if not self.backend:
+            self.backend = self.get_atomic_config_item(["default_storageat"])
```
runcom: typo `storageat` ?

bash/atomic
```diff
@@ -662,7 +663,7 @@ _atomic() {
 		info
 		install
 		images
-		migrate
```
runcom: can't we leave `migrate` as a valid command in the code so ppl using `atomic migrate` won't get hurt by its replacement to `atomic storage`? no big deal though

rhvgoyal:
Can't find where is the code to reset the storage.

bash/atomic
rhatdan: I can add an alias.

Atomic/atomic.py
rhatdan: yup, thanks.

rhatdan:
Oops forgot to add storage.py to git, it is there now.  Fixed all other issues, and fixed a bug I created with atomic pull,  by changing the self.storage to self.backend.

rhatdan:
@cevitch we probably need a CI for this to test resetting different types of storage. Pretty difficult to set this up in a unit test.  (Unless people want me blowing away their docker storage...

Atomic/storage.py
```diff
@@ -0,0 +1,59 @@
...
+
+        unit="docker"
+        try:
+            subprocess.check_call(["systemctl", "stop", unit])
+        except:
+            unit = "docker-latest"
+            subprocess.check_call(["systemctl", "stop", unit])
+        
+        if loopfile == "" and loopdev == "":
+            out = subprocess.check_call(["lvmremove", "/dev/mapper/%s" % name.decode(enc)])
```
mvollmer: There is no `lvmremove`... didn't you run this yet?

mvollmer: This probably needs `--force` to convince LVM to really remove the logical volume.

Atomic/storage.py
```diff
@@ -0,0 +1,59 @@
...
+        shutil.rmtree(root)
```
mvollmer: In my prototype I found that I needed to also `umount /var/lib/docker/devicemapper`.

rhvgoyal: That should not be required. Docker daemon does it.

Atomic/storage.py
rhvgoyal: Why --force? I think without --force should be just fine. Otherwise its an anomaly which should be looked into.

rhatdan: Have not tried this with a system other then loopback devices.  Wanted others to try and tell me what is wrong.
@rhvgoyal What command should I use to reomve the device?

Atomic/storage.py
rhatdan: I think that happened to me also.  Perhaps in certain situations docker fails to remove it?

Atomic/storage.py
rhvgoyal: which device? to remove thin pool you just have to use "lvremove <thin-pool-name>"

mvollmer: > Why --force? I think without --force should be just fine. Otherwise its an anomaly which should be looked into.

I always get an extra confirmation prompt about removing an active logical volume.  I would say lvm does that unconditionally, but I might be wrong.  I will investigate this more.

Atomic/storage.py
mvollmer: Interesting, I'll have a closer look.

Atomic/storage.py
rhvgoyal: In that case let us first deactivate the volume using "lvchange" and then call "lvremove". --force can imply lot of things and can be dangerous. Want to avoid using --force until and unless we implement a --force option in atomic reset.

mvollmer: > In that case let us first deactivate the volume using "lvchange" and then call "lvremove". 

Very good.  This works without any prompts.

rhatdan:
Did some more testing with both python2 and python3, on devicemapper on loopback and storage.  Have not tried on physical or on other backends like overlayfs and btrfs.  But it is getting closer.

Atomic/storage.py
```diff
@@ -0,0 +1,89 @@
...
+            
+        shutil.rmtree(root)
+        os.mkdir(root)
+        try:
+            selinux.restorecon(root.encode("utf-8"))
+        except:
+            selinux.restorecon(root)
+                               
+        try:
+            os.remove("/etc/sysconfig/%s-storage-setup" % unit )
```
mvollmer: This should remove `docker-storage`, not `docker-storage-setup`.

mvollmer:
This used to start docker at the end, and thus would implicitly run docker-storage-setup.  Why isn't this done anymore?

Without running docker-storage-setup at the end, the storage pool is in a slightly weird intermediate state.

This state is useful since one can cleanly remove physical volumes at that point.  However, I am thinking that "atomic storage" should also take care of adding/removing pvs, so we might be better off without letting the user see this intermediata state.

mvollmer:
"atomic storage reset" requires a running docker daemon.  This means that one can't use it to repair a broken storage pool that prevents the daemon from starting in the first place.

mvollmer:
> Without running docker-storage-setup at the end, the storage pool is in a slightly weird intermediate state.

One indication that this state is weird is that one can't run "atomic storage reset" twice in a row.  The second run will fail because the daemon isn't running.

mvollmer:
I get a whole lot of output like this:

```
File descriptor 3 (socket:[260846]) leaked on lvchange invocation. Parent PID 30365: /usr/bin/python
File descriptor 4 (/dev/null) leaked on lvchange invocation. Parent PID 30365: /usr/bin/python
File descriptor 5 (/dev/null) leaked on lvchange invocation. Parent PID 30365: /usr/bin/python
File descriptor 6 (/dev/null) leaked on lvchange invocation. Parent PID 30365: /usr/bin/python
File descriptor 7 (socket:[260858]) leaked on lvchange invocation. Parent PID 30365: /usr/bin/python
File descriptor 8 (socket:[260870]) leaked on lvchange invocation. Parent PID 30365: /usr/bin/python
File descriptor 3 (socket:[260846]) leaked on lvremove invocation. Parent PID 30365: /usr/bin/python
File descriptor 4 (/dev/null) leaked on lvremove invocation. Parent PID 30365: /usr/bin/python
File descriptor 5 (/dev/null) leaked on lvremove invocation. Parent PID 30365: /usr/bin/python
File descriptor 6 (/dev/null) leaked on lvremove invocation. Parent PID 30365: /usr/bin/python
File descriptor 7 (socket:[260858]) leaked on lvremove invocation. Parent PID 30365: /usr/bin/python
File descriptor 8 (socket:[260870]) leaked on lvremove invocation. Parent PID 30365: /usr/bin/python
```

Atomic/storage.py
rhatdan: Ok I will fix this.

rhatdan:
@mvollmer @rhvgoyal Wanted the restart removed, so that an admin could go an make changes to the config before restarting the docker.  He sees this as putting you back to the state you were in before the first start of the docker service, not right after the first start of docker.

rhatdan:
Not sure what the best way to handle the non running docker is, unless we hard code to go and look at the storage configuration, which would be difficult to implement.

rhatdan:
@rhvgoyal Could you comment on the weird output of lvchange?

mvollmer:
> @rhvgoyal Could you comment on the weird output of lvchange?

The output is produced by lvm itself, `man lvm`:

```
On invocation, lvm requires that only the standard file descriptors
stdin, stdout and stderr are available.  If others are found, they
get closed and messages are issued warning about the leak.  This
warning can be suppressed by setting the environment variable
LVM_SUPPRESS_FD_WARNINGS.
```

Strange...

cgwalters:
Python is an ancient language that predates modern things like `O_CLOEXEC` and inherits the Unix default of leaking fds to children by default (a huge design mistake).

One should generally use `close_fds=True` to subprocess invocations.

rhvgoyal:
I think we should not start docker automatically after reset. And reason being that what if I want to change storage configuration in /etc/sysconfig/docker-storage before docker is restarted.

Also I agree that requiring docker to be running for docker reset to work seems odd and will not work for the cases where storage is broken.

May be we can parse "/etc/sysconfig/docker-storage" to figure out currently configured storage and reset storage accordingly. I think even better is to implement "docker-storage-setup reset" and let it parse the file and reset storage.

I think workflow should be.
- systemctl stop docker
- atomic storage reset
- Make changes to /etc/sysconfig/docker-storage-setup if need be.
- systemctl start docker

rhatdan:
One problem with this is if docker is not running, we can not tell if the user intends to use docker-latest or docker, so we can't even decide whether to edit docker-latest-storage-setup or docker-storage-setup.

In the future we would probably want atomic storage reset to be able reset other container storage besides docker.  We can add options to do this, but I am not sure how discoverable the environment would be.

rhatdan:
This patch set now requires this to get merged.

mvollmer:
I am fine with `atomic reset` not restarting docker.  Cockpit handles starting of docker already, and it's not difficult in any case.

Cockpit would need to know which unit to start (docker or docker-latest), but it needs to know that anyway.  Should there be a UI somewhere to let the user choose?

Also, when `docker-storage-setup` is run during `systemctl start docker`, any failure is ignored and the user will never know that something is wrong.  I think that can be improved, as can Cockpits reporting of docker startup errors.

rhatdan:
Could we add something to docker-storage-setup to tell it to not ignore errors.  I think if the system has a modified /etc/sysconfig/docker-storage-setup, then a failure should be a true failure and should not fail over to loopback devices.

rhvgoyal:
docker-storage-setup does not mask errors. It is docker which has Wants= dependency and not Requires= dependency on docker-storage-setup. That means even if docker-storage-setup failed, docker will continue to start.

I am not sure how docker-storage-setup will stop docker from running. (Until and unless it decides to drop invalid storage options in /etc/sysconfig/docker-storage) and docker fails to start. But that sounds ugly.

And all these systemd dependencies are pretty static in nature. One can't change it based on whether /etc/sysconfig/docker-storage-setup was modified or not.

Atomic/storage.py
```diff
@@ -0,0 +1,52 @@
...
+class Storage(Atomic):
+    def reset(self):
+        root = "/var/lib/docker"
+        try:
+            self.d.info()
+            raise ValueError("Docker daemon must be stop before resetting storage")
+        except requests.exceptions.ConnectionError as e:
+            pass
+            
+        util.check_call(["docker-storage-setup", "reset"], stdout=DEVNULL)
```
mvollmer: This needs to be `--reset` to match https://github.com/projectatomic/docker-storage-setup/pull/128

rhatdan:
@baude @shishir-a412ed PTAL
This now requires the latest docker-storage-setup from github to work.

https://github.com/projectatomic/docker-storage-setup/commit/194eca25fd0d180b62f3ecf1b7b408992fd6a083

rhatdan:
Be careful testing atomic storage reset, because it will blow away all of your containers and images.

mvollmer:
The interface looks good to me, and the implementation works in my simple tests.

I am manually testing this via https://github.com/cockpit-project/cockpit/pull/4350 with a regular Fedora 24 Server VM.  It works for me in my simple setup, but I did not test whether this can reset all the possible docker storage configurations.

rhatdan:
Ok I am going to merge, and we will fix any issues that arise with atomic storage export/import.

rhatdan:
@cgwalters-bot r+

PR #366, opened 2016-04-28
Label temporary containers with /usr label

rhatdan:


rhatdan:
@baude PTAL, this will label the atomic mount images as usr_t which should be readable inside of atomic scanners.

baude:
@rhatdan I tested scan and diff and a few others; this LGTM.  I'll make a comment where I think the tests are failing.  I also found a bug that I think existed prior to this (maybe?) but I will fix it once this is merged.

tests/integration/test_printenv.sh
```diff
@@ -32,7 +32,7 @@ trap teardown EXIT
 
 OUTPUT=$(/bin/true)
 
-${ATOMIC} run atomic-test-5 | grep -v ^NAME= | grep -v ^IMAGE= | grep -v printenv | grep -v atomic | grep -v coverage | sort > ${TESTDIR}/atomic-test-5.1
+${ATOMIC} run atomic-test-5 | grep -v ^NAME= | grep -v ^IMAGE= | grep -v ^SUDO | grep -v printenv | grep -v atomic | grep -v coverage | sort > ${TESTDIR}/atomic-test-5.1
 
 printenv | grep -v printenv | grep -v atomic | grep -v coverage | sort > ${TESTDIR}/atomic-test-5.2
```
baude: In the previous comparison, you grep "out" SUDO vars.  Should you do the same here as well?

rhatdan: Yes.

rhatdan:
@cgwalters-bot r+ ba2330e

rhatdan:
@cgwalters-bot r+ ba2330e

PR #367, opened 2016-04-29
A couple of testing fixes

rhatdan:
On certain systems loginuid might not be set which could give a value of
4294967295

docker will not allow users > 2147483647, if you loginuid is -1 it ends up
being bigger

SUDO_UID and SUDO_GID, not always defined in environment


rhatdan:
@baude @cgwalters @mrunalp @runcom PTAL

giuseppe:
LGTM

runcom:
LGTM

PR #368, opened 2016-05-02
[merged] Atomic/diff.py: Do not hide SelectionMatchError and clean-up

baude:
Previously, the raise of SelectionMatchError was not getting to
stderr so diff would fail silently.  We now write the error to
stderr and clean up any mounts and dm devices created.

### DIFF
--- a/Atomic/diff.py
+++ b/Atomic/diff.py
@@ -77,7 +77,13 @@ class DiffHelpers(object):
         """
         image_list = []
         for image in images:
-            image_list.append(DiffObj(image))
+            try:
+                image_list.append(DiffObj(image))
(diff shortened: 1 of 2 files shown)


Atomic/diff.py
```diff
@@ -67,8 +67,7 @@ def _cleanup(image_list):
         for image in image_list:
             image._remove()
 
-    @staticmethod
-    def create_image_list(images):
+    def create_image_list(self, images):
```
rhatdan: Why can't this stay a staticmethod and call helpers._cleanup(image_list)?

baude:
fixed and passes tests

rhatdan:
@cgwalters-bot r+ f573361

PR #369, opened 2016-05-03
[merged] Atomic/scan.py: Add custom out for main scan output

baude:
We had the ability to print out custom key/value pairs in the
vulnerability and results sections.  But it is clear folks
who develop scan plugins will want to also output custom
key/value pairs for the scan itself.  For example, BlackDuck
wants to output a URL for their report.


rhatdan:
LGTM, but you need to update your pull request.

baude:
@cgwalters-bot r+ e316faf

PR #370, opened 2016-05-03
[merged] We should close all FD's when executing a subprocess

rhatdan:
This patch changes most calls to subprocess to be through
util package.  And the util subprocess calls all close_fds=True.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -78,16 +78,30 @@ def subp(cmd, cwd=None):
     """
     proc = subprocess.Popen(cmd, cwd=cwd,
                             stdout=subprocess.PIPE,
-                            stderr=subprocess.PIPE)
+                            stderr=subprocess.PIPE, close_fds=True)
     out, err = proc.communicate()
     return ReturnTuple(proc.returncode, stdout=out, stderr=err)
 
-
+# Wrappers for Python's subprocess which override the default for close_fds,
+# since we are a privileged process, and we don't want to leak things like
+# the docker socket into child processes by default
 def check_call(cmd, env=os.environ, stdin=None, stderr=None, stdout=None):
     # Make sure cmd is a list
     if not isinstance(cmd, list):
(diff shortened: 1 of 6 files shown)


rhatdan:
@cgwalters @baude @runcom PTAL

Atomic/util.py
```diff
@@ -87,7 +87,19 @@ def check_call(cmd, env=os.environ, stdin=None, stderr=None, stdout=None):
     # Make sure cmd is a list
     if not isinstance(cmd, list):
         cmd = shlex.split(cmd)
-    return subprocess.check_call(cmd, env=env, stdin=stdin, stderr=stderr, stdout=stdout)
+    return subprocess.check_call(cmd, env=env, stdin=stdin, stderr=stderr, stdout=stdout, close_fds=True)
+
+def check_output(cmd, env=os.environ, stdin=None, stderr=None, stdout=None):
```
cgwalters: I would add a comment here of the form:

```
# Wrappers for Python's subprocess which override the default for close_fds,
# since we are a privileged process, and we don't want to leak things like
# the docker socket into child processes by default
```

cgwalters:
I'd prefer a commit message that matches http://chris.beams.io/posts/git-commit/

Something like:

```Use close_fds=True for all subprocess invocations

The Python default is to keep all FDs open, which is how Unix
traditionally worked.  This default behavior is particularly
problematic for the atomic command, which e.g. may
hold a reference to the Docker socket, and we don't
necessarily want to give that reference to all
child processes.
```

runcom:
LGTM (with Colin's comment)

Atomic/util.py
rhatdan: Added

rhatdan:
Changed commit message to match Colins.

rhatdan:
@cgwalters PTAL

cgwalters:
@cgwalters-bot r+ 754b6dd

PR #371, opened 2016-05-03
[merged] Atomic scan chroot

baude:


baude:
@rhatdan I should probably merge these two commits into a single one?

atomic
```diff
@@ -444,6 +444,7 @@ if __name__ == '__main__':
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scanner"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
     scanp.add_argument("--verbose", action='store_true', default=False, help=_("Show more output from scanning container"))
+    scan_group.add_argument("--rootfs", nargs='*', help=_("One or more paths to scan"))
     scan_group.add_argument("--all", default=False, action='store_true', help=_("scan all images (excluding intermediate layers) and containers"))
```
rhatdan: I think it would be better to call --rootfs multiple times then attempt to deal with multiple paths on the same command.

rhatdan:
Yes merge them together.

baude:
@rhatdan updated per your comments

Atomic/scan.py
```diff
@@ -55,19 +55,36 @@ def get_scan_info(scanner, scan_type):
...
 
+        # Create the input directory
+        if not os.path.exists(self.chroot_dir):
+            os.makedirs(self.chroot_dir)
         if self.debug:
-            util.write_out("Creating the output dir at {}".format(self.results_dir))
+            util.write_out("Created {}".format(self.chroot_dir))
+
+        if self.args.rootfs is None:
+            scan_list = self._get_scan_list()
```
rhatdan: Shouldn't rootfs be a list now?  If len(self.args.rootfs[] == 0:

baude: Nope, if unpopulated, it is None.

[bbaude@localhost atomic (atomic_scan_chroot)]$ sudo ./atomic scan foobar
Namespace(_class=<class 'Atomic.scan.Scan'>, all=False, containers=False, debug=False, func='scan', images=False, list=False, rootfs=None, scan_targets=['foobar'], scan_type=None, scanner='openscap', verbose=False)

rhatdan: Try something like:
parser.add_argument('--foo', nargs='?', default=[], action="append", help='foo help')

baude: Does this mean you want it to behave that way and don't like the None behaviour?

rhatdan: I like to think of rootfs as a list.  I would prefer to always treat it as a list.  The problem with having it be sometimes a none and sometimes a list, you end up with code that looks like

if self.args.rootfs:
      for r in self.args.rootfs:
                print r

It it is always a list I don't need the first if.

baude:
@rhatdan reworked as a list ...

Atomic/scan.py
```diff
@@ -60,19 +60,36 @@ def get_scan_info(scanner, scan_type):
...
 
+        # Create the input directory
+        if not os.path.exists(self.chroot_dir):
+            os.makedirs(self.chroot_dir)
         if self.debug:
-            util.write_out("Creating the output dir at {}".format(self.results_dir))
+            util.write_out("Created {}".format(self.chroot_dir))
+
+        if len(self.args.rootfs) < 1:
+            scan_list = self._get_scan_list()
```
rhatdan: Why not == 0?

Atomic/scan.py
```diff
@@ -60,19 +60,36 @@ def get_scan_info(scanner, scan_type):
...
+        else:
+            # Check to make sure all the chroots provided are legit
+            for _path in self.args.rootfs:
+                if not os.path.exists(_path):
+                    raise ValueError("The path {} does not exist".format(_path))
 
         # Create the output directory
         os.makedirs(self.results_dir)
 
+        if len(self.args.rootfs) > 0:
```
rhatdan: Why not put this above makedirs in the same else: block?

atomic
```diff
@@ -444,6 +444,7 @@ if __name__ == '__main__':
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scanner"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
     scanp.add_argument("--verbose", action='store_true', default=False, help=_("Show more output from scanning container"))
+    scan_group.add_argument("--rootfs", nargs='?', default=[], action='append', help=_("One or more paths to scan"))
     scan_group.add_argument("--all", default=False, action='store_true', help=_("scan all images (excluding intermediate layers) and containers"))
```
rhatdan: scan_group.add_argument("--rootfs", nargs='?', default=[], action='append', help=_("Rootfs path to scan"))

docs/atomic-scan.1.md
```diff
@@ -6,24 +6,33 @@ atomic-scan - Scan for CVEs in a container or image
 # SYNOPSIS
 **atomic scan**
 [**-h**|**--help**]
-[**--fetch_cves=True|False**][**--json** | **--detail**] [**--all** | **--images** | **--containers** |
-IMAGE or CONTAINER name ...]
+[**--fetch_cves=True|False**][**--json** | **--detail**] [**--all** | **--images** | **--containers** | **--rootfs** list of paths to scan |
+IMAGE or CONTAINER names ...]
```
rhatdan: **--rootfs** rootfs path to scan

docs/atomic-scan.1.md
```diff
@@ -34,26 +43,42 @@ IMAGE or CONTAINER name ...]
 **--containers**
   Scan all containers.  Similar to the results of `docker ps -a`
 
+**--rootfs**
+  Scan one or more paths which are presumably chroots.
+
```
rhatdan: Rootfs path to scan

rhatdan:
You have to disable SELinux inside the scanning container for --rootfs to work.

Command completion needs a fix for this also.

baude:
@rhatdan I believe all the comments are now addressed with the exception of the selinux issue.

rhatdan:
LGTM

rhatdan:
@cgwalters-bot r+ 423c226

PR #372, opened 2016-05-05
[merged] SELinux: Don't scan with selinux separation if scan is using rootfs

rhatdan:
We don't have a way to relabel content off of random rootfs so that
a SELinux confined process can read the content.

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -77,13 +77,15 @@ class Scan(Atomic):
 
             if self.debug:
                 util.write_out("Creating the output dir at {}".format(self.results_dir))
-
+            security_args = []
         else:
             # Check to make sure all the chroots provided are legit
             for _path in self.args.rootfs:
                 if not os.path.exists(_path):
                     raise ValueError("The path {} does not exist".format(_path))
             self.setup_rootfs_dirs()
+            # Have to disable SELinux checking of scanning not docker images
+            security_args = [ "--security-opt", "label:disable" ]
 
         # Create the output directory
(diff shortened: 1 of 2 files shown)


rhatdan:
@baude  could you try this patch out, I don't have the environment to test.

Atomic/scan.py
```diff
@@ -77,13 +77,15 @@ def get_scan_info(scanner, scan_type):
...
-
+            security_args = []
         else:
             # Check to make sure all the chroots provided are legit
             for _path in self.args.rootfs:
                 if not os.path.exists(_path):
                     raise ValueError("The path {} does not exist".format(_path))
             self.setup_rootfs_dirs()
+            # Have to disable SELinux checking of scanning not docker images
+            security_args = [ "--security-opt", "label=disable" ]
```
baude: should be label:disable

baude:
With the one correction, it is good.  Feel free to merge

Atomic/scan.py
rhatdan: They are changing the default to label=disable

rhatdan:
@baude did you try it out?

baude:
Yeah and it failed because of the label syntax

rhatdan:
It works with the ":",  I guess this is fixed in newer dockers...

rhatdan:
@baude See if it works with the ":"

baude:
Works perfectly for me

rhatdan:
@cgwalters-bot r+ 7326c2d

PR #374, opened 2016-05-06
[merged] Atomic/atomic.py: Deal with non-ascii input

baude:
https://bugzilla.redhat.com/show_bug.cgi?id=1323165#c2 suggests
that atomic install does not deal with non-ascii input very
well.  I believe this addresses the concerns with the proper
UTF-8 encode and decode operations.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -140,6 +140,7 @@ class Atomic(object):
         self.atomic_config = None
         self.docker_cmd = None
         self.debug = False
+        self.is_python2 = (int(sys.version[0])) < 3
 
     def docker_binary(self):
         if not self.docker_cmd:
@@ -662,7 +663,10 @@ class Atomic(object):
                 args.append(self.name)
                 continue
             args.append(c)
-        return " ".join(args)
+        if self.is_python2:
+            return " ".join([x.decode('utf-8') for x in args])
+        else:
+            return " ".join(args)
 
     def get_fq_name(self, image_info):
         if len(image_info['RepoTags']) > 1:
@@ -1301,7 +1305,7 @@ class Atomic(object):
             self.write_out("%s %s %s" % (layer["Id"], version, layer["Tag"]))
 
     def display(self, cmd):
-        util.write_out(self.sub_env_strings(cmd))
+        util.write_out(cmd)
 
     def sub_env_strings(self, in_string):
         """
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -20,10 +20,12 @@ ATOMIC_CONFD = os.environ.get('ATOMIC_CONFD', '/etc/atomic.d/')
 _default_docker=None
 _default_docker_lib=None
 
-if sys.version_info[0] < 3:
+if int(sys.version_info[0]) < 3:
     input = raw_input # pylint: disable=undefined-variable
+    is_python2 = True
 else:
     input = input
+    is_python2 = False
 
 
 def _decompose(compound_name):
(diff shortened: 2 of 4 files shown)


rhatdan:
@adelton  PTAL

cgwalters:
I'm almost certain this is going to need to handle Python 2 vs Python 3.  Did you test it in both scenarios?

http://lucumr.pocoo.org/2014/5/12/everything-about-unicode/

baude:
@cgwalters Thanks for the reference, Im going to sit elsewhere and read it.  I think I have the conditional handled but maybe the article will inform me differently.

Atomic/util.py
```diff
@@ -119,7 +123,7 @@ def default_ro_container_context():
 
 def write_out(output, lf="\n"):
     sys.stdout.flush()
-    sys.stdout.write(str(output) + lf)
+    sys.stdout.write(output + lf)
```
cgwalters: For example it looks like the `str` there came from 6e1f4dde ...which seems like it was probably copied from somewhere for a reason.  I'm guessing it's something like needing to handle stdout being passed unicode but expecting bytes?

baude:
Passes all tests with python2 and python3 including your proposed new test -> https://github.com/projectatomic/atomic/pull/358

Atomic/util.py
```diff
@@ -119,7 +123,7 @@ def default_ro_container_context():
 
 def write_out(output, lf="\n"):
     sys.stdout.flush()
-    sys.stdout.write(str(output) + lf)
+    sys.stdout.write(output.encode('utf-8') + lf)
```
cgwalters: Are you really sure this works on Python 3?

```
[walters@9845f984f09d ostree]$ python3
Python 3.4.3 (default, Mar 31 2016, 20:42:37) 
[GCC 5.3.1 20151207 (Red Hat 5.3.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "foo".encode('utf-8') + "\n"
  File "<stdin>", line 1, in <module>
TypeError: can't concat bytes to str
>>> 
```

i.e. don't we have to do `lf.encode('utf-8')` too?

Atomic/atomic.py
```diff
@@ -142,6 +142,9 @@ def __init__(self):
         self.atomic_config = None
         self.docker_cmd = None
         self.debug = False
+        self.is_python2 = True
+        if int(sys.version[0]) > 2:
+            self.is_python2 = False
```
rhatdan: self.is_python2 = (int(sys.version[0]) <= 2)

Atomic/atomic.py
```diff
@@ -677,7 +680,10 @@ def gen_cmd(self, cargs):
                 args.append(self.name)
                 continue
             args.append(c)
-        return " ".join(args)
+        if self.is_python2:
+            return " ".join([x.decode('utf-8') for x in args])
+        else:
+            return " ".join(x for x in args)
```
rhatdan: return " ".join(args)

Atomic/util.py
```diff
@@ -119,7 +125,10 @@ def default_ro_container_context():
 
 def write_out(output, lf="\n"):
     sys.stdout.flush()
-    sys.stdout.write(str(output) + lf)
+    if is_python2:
+        sys.stdout.write(output.encode('utf-8') + lf)
+    else:
+        sys.stdout.write(output + str(lf))
```
rhatdan: Why is the str(lf) necessary?

Atomic/util.py
```diff
@@ -86,9 +88,13 @@ def subp(cmd, cwd=None):
 # since we are a privileged process, and we don't want to leak things like
 # the docker socket into child processes by default
 def check_call(cmd, env=os.environ, stdin=None, stderr=None, stdout=None):
-    # Make sure cmd is a list
+    # Make sure cmd is a list; break if needed
     if not isinstance(cmd, list):
-        cmd = shlex.split(cmd)
+        try:
```
rhatdan: Should we just do the is_python2 check here?

Atomic/util.py
baude: BC in python 3, output is a str and lf is a byte and then cannot be concatenated.

Atomic/util.py
baude: Yeah, seems prudent.

Atomic/util.py
rhatdan: ok.

Atomic/util.py
```diff
@@ -86,9 +88,13 @@ def subp(cmd, cwd=None):
 # since we are a privileged process, and we don't want to leak things like
 # the docker socket into child processes by default
 def check_call(cmd, env=os.environ, stdin=None, stderr=None, stdout=None):
-    # Make sure cmd is a list
+    # Make sure cmd is a list; break if needed
     if not isinstance(cmd, list):
-        cmd = shlex.split(cmd)
+        if not is_python2:
```
rhatdan: Can you make this positive like you did below.

rhatdan:
```
if is_python2:
            # The command contains a non-ascii character
            cmd = shlex.split(" ".join([x.encode('utf-8') for x in cmd.split()]))
else:
...
```

rhatdan:
@cgwalters-bot r+

PR #376, opened 2016-05-10
[merged] Extend Try block by adding create_parser() function

rhatdan:
Currently we do not catch all exceptions, since we setup the Atomic
class before the try block.  This patch moves all of the parser
creation into its own function.  And wraps all calls in a try block.

### DIFF
--- a/atomic
+++ b/atomic
@@ -115,9 +115,7 @@ def need_root():
     exit("Some operations for '%s' require root access." % sub_function)
     sys.exit(1)
 
-if __name__ == '__main__':
-    atomic_config = get_atomic_config()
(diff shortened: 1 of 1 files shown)


rhatdan:
@cgwalters @baude @mrunalp @runcom PTAL

baude:
out of curiosity, what was the problem you were hitting prior to the change?

rhatdan:
@baude docker was hanging while trying to connect to the docker service.  When I hit ctrl-c, I saw a Keyboard Exception.

baude:
LGTM

rhatdan:
@cgwalters-bot r+

PR #380, opened 2016-05-11
[merged] Improve errors msgs for no configured scanners

baude:
Eventually we will not ship any default scanner configurations.  Previous
code relied on there always being one (default) defined so improved
error messages and a conditional check was in order.

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -344,6 +344,9 @@ class Scan(Atomic):
                              format(self.args.scan_type, self.args.scanner))
 
     def print_scan_list(self):
+        if len(self.scanners) == 0:
+            util.write_out("There are no scanners configured for this system.")
+            sys.exit(0)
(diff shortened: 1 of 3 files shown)


baude:
@rhatdan How should we handle to configuration file that is still in the repo and Makefile for install?

rhatdan:
Make it optional.  Basically we want to move this to atomic-openscap.rpm subpackage and then have it only installed on RHEL systems.

baude:
@rhatdan do you want me to change that in this PR or do we handle it in the spec files?

rhatdan:
I think you should move the installing of rhe openscap config into a separate make install-scap step.

Then on RHEL we can call it and put this content into a separate subpackage.

rhatdan:
Please update pull request.

rhatdan:
@cgwalters-bot r+

Issue #378, opened 2016-05-10
Using --names-only as option to 'atomic diff' does not produce any RPM differences


miabbott:
Using atomic 1.9 on RHELAH 7.2.4:
```
# atomic --version
1.9
# rpm -q atomic
atomic-1.9-4.gitff44c6a.el7.x86_64
```
When running `atomic diff -r <image1> <image2>`, I get a bunch of file differences and then a table with RPM differences:

```
# atomic diff -r registry.access.redhat.com/rhel7.2:7.2-46 registry.access.redhat.com/rhel7.2:7.2-56

Files only in registry.access.redhat.com/rhel7.2:7.2-46:
     /var/log/anaconda/ks-script-qAOJrD.log
     /var/log/anaconda/ks-script-k8yjSj.log
     /var/log/anaconda/ks-script-7LFMQs.log
     /var/lib/yum/history/history-2016-03-01.sqlite
     /var/lib/yum/history/history-2016-03-01.sqlite-journal
     /var/lib/yum/history/2016-03-01
...
registry.access.redhat.com/rhel7.2:7.2-46 | registry.access.redhat.com/rhel7.2:7.2-56
--------------------------------------- | ---------------------------------------
Red Hat Enterprise Linux Server relea   | Red Hat Enterprise Linux Server relea  
se 7.2 (Maipo)                          | se 7.2 (Maipo)                         
--------------------------------------- | ---------------------------------------
kpartx-0-0.4.9-85.el7                   |                                        
                                        | kpartx-0-0.4.9-85.el7_2.1              
krb5-libs-0-1.13.2-10.el7               |                                        
                                        | krb5-libs-0-1.13.2-12.el7_2            
libblkid-0-2.23.2-26.el7                |                                        
                                        | libblkid-0-2.23.2-26.el7_2.2           
libmount-0-2.23.2-26.el7                |                                        
                                        | libmount-0-2.23.2-26.el7_2.2           
libssh2-0-1.4.3-10.el7                  |                                        
                                        | libssh2-0-1.4.3-10.el7_2.1             
libuuid-0-2.23.2-26.el7                 |                                        
                                        | libuuid-0-2.23.2-26.el7_2.2            
nss-util-0-3.19.1-4.el7_1               |                                        
                                        | nss-util-0-3.19.1-9.el7_2              
openldap-0-2.4.40-8.el7                 |                                        
                                        | openldap-0-2.4.40-9.el7_2              
[...]

PR #381, opened 2016-05-11
[merged] Atomic/diff: Improve docs and output messages for diff

baude:
https://github.com/projectatomic/atomic/issues/378 points out that
when an rpm diff is done with --names-only, the output messaging was
not clear enough.  Reworked the man page, --help, and output message
to clarify that when --names-only is used, it only compares RPMs
based on names and NOT versions.

Nice find by Micah.

### DIFF
--- a/docs/atomic-diff.1.md
+++ b/docs/atomic-diff.1.md
@@ -14,8 +14,9 @@ atomic-diff - show the differences between two images|containers RPMs
 image|container image|container ...]
 
 # DESCRIPTION
-**atomic diff** will compare the RPMs found in two different images or containers and output to stdout or as JSON.
-By default, the comparison is done by name and version of the RPMs.
+**atomic diff** will compare the RPMs found in two different images or containers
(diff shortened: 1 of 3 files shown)


docs/atomic-diff.1.md
```diff
@@ -57,4 +58,5 @@ Compare the files and RPMs (without versions) in images 'foo1' and 'foo2' and ou
     atomic diff -r --json foo1 foo2
 
 # HISTORY
+Updated by Brent Baude (bbaude at redhat dot com) May 2061
```
miabbott: Do you have a time machine? :smile:

miabbott:
Small nit about the date, otherwise LGTM

baude:
Fixed date

rhatdan:
@cgwalters-bot r+

rhatdan:
@jlebon @cgwalters Jenkins seems to be stuck?

jlebon:
@rhatdan Just unstuck it, there was a slight blip earlier. Should report in a few.

PR #383, opened 2016-05-12
[merged] Change rules around declaring a default scanner.

baude:
Previously, with the understanding we would be shipping a single
generic scanner configuration file, would 1) install the config
file and we had preset a default scanner in atomic.conf.

With the realization that the scanner could be variable based
on the distribution, the generic file is not longer installed
by default. Therefore, we should probably not set a default.

The prior code failed if a default was not defined in
/etc/atomic.conf.  That code check has been altered and we now
adhere to the following rules:

* If no default is defined AND there is only one configured
 scanner, we assume the single scanner IS the default.

* If no default is defined AND there are multiple scanners, we
 error out and ask the user to define one.

* If there is a default defined, good to go!

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -26,6 +26,7 @@ class Scan(Atomic):
         self.scanners = util.get_scanners()
         self.debug = False
         self.rootfs_mappings = {}
+        self.scanner = None
 
     def scan(self):
         def get_scan_info(scanner, scan_type):
@@ -34,11 +35,33 @@ class Scan(Atomic):
                     for x in i['scans']:
                         if x['name'] == scan_type:
                             return i['image_name'], x['args'], i.get('custom_args')
+
+        def set_scanner():
+            # If no scanners are defined
+            if len(self.scanners) == 0:
+                raise ValueError("No scanners are configured for your system.")
+
+            if len(self.scanners) == 1:
+                self.scanner = self.scanners[0]['scanner_name']
+
+            # If a scanner is defined in parseargs
+            if self.args.scanner:
+                    self.scanner = self.args.scanner
+
+            if self.scanner is None:
+                raise ValueError("You must specify a scanner")
+
+            # Check if the scanner name is valid
+            if self.scanner not in [x['scanner_name'] for x in self.scanners]:
+                raise ValueError("Unknown scanner '{}' defined in {}".format(self.scanner, util.ATOMIC_CONF))
+
         if self.args.debug:
             self.debug = True
 
         # Set debug bool
         self.set_debug()
+
+        set_scanner()
 
         if self.args.list:
             self.print_scan_list()
@@ -46,20 +69,18 @@ class Scan(Atomic):
         scan_type = self.get_scan_type()
         # Load the atomic config file and check scanner settings
         yaml_error = "The image name or scanner arguments for '{}' is not " \
-                     "defined in /etc/atomic.conf".format(self.args.scanner)
+                     "defined in /etc/atomic.conf".format(self.scanner)
 
-        if self.args.scanner not in [x['scanner_name'] for x in self.scanners]:
-            raise ValueError("Unknown scanner '{}' defined in {}".format(self.args.scanner, util.ATOMIC_CONF))
-        scanner_image_name, scanner_args, custom_args = get_scan_info(self.args.scanner, scan_type)
+        scanner_image_name, scanner_args, custom_args = get_scan_info(self.scanner, scan_type)
 
         if not isinstance(scanner_args, list):
             raise ValueError("The scanner arguments for {} must be in list"
-                             " ([]) form.".format(self.args.scanner))
+                             " ([]) form.".format(self.scanner))
 
         if None in [scanner_image_name, scanner_args]:
             raise ValueError(yaml_error)
 
-        self.results_dir = os.path.join(self.results, self.args.scanner, self.cur_time)
+        self.results_dir = os.path.join(self.results, self.scanner, self.cur_time)
 
         # Create the input directory
         if not os.path.exists(self.chroot_dir):
@@ -330,7 +351,7 @@ class Scan(Atomic):
         default_scan_type = None
         scan_types = []
         for i in self.scanners:
-            if i['scanner_name'] == self.args.scanner:
+            if i['scanner_name'] == self.scanner:
                 default_scan_type = i.get('default_scan')
                 if self.args.scan_type is None and default_scan_type is None:
                     raise ValueError("No scan type was given and there is no "
@@ -343,13 +364,13 @@ class Scan(Atomic):
             return self.args.scan_type
         else:
             raise ValueError("Unable to find the scan type '{}' for '{}'.".
-                             format(self.args.scan_type, self.args.scanner))
+                             format(self.args.scan_type, self.scanner))
 
     def print_scan_list(self):
         if len(self.scanners) == 0:
             util.write_out("There are no scanners configured for this system.")
             sys.exit(0)
-        default_scanner = self.atomic_config.get('default_scanner')
+        default_scanner = self.atomic_config['default_scanner']
         if default_scanner is None:
             default_scanner = ''
         for scanner in self.scanners:
--- a/atomic
+++ b/atomic
@@ -394,20 +394,13 @@ def create_parser(atomic):
         return True if myarg.lower() in \
             ('yes', 'true', 't', 'y', '1') else False
     scanners = get_scanners()
-    atomic_config = get_atomic_config()
-    if atomic_config is None or atomic_config.get('default_scanner') is None:
-        sys.stderr.write("\nYou must have a default scanner defined in /etc/atomic.conf. Use --list to see "
-                         "what scanners are available for your system.\n")
-        sys.exit(1)
-
-    default_scanner = atomic_config.get('default_scanner')
     scanp = subparser.add_parser(
         "scan", help=_("scan an image or container for CVEs"),
         epilog="atomic scan <input> scans a container or image for CVEs")
     scanp.set_defaults(_class=Scan, func='scan')
     scan_group = scanp.add_mutually_exclusive_group()
     scanp.add_argument("scan_targets", nargs='*', help=_("container image"))
-    scanp.add_argument("--scanner", choices=[x['scanner_name'] for x in scanners], default=default_scanner, help=_("define the intended scanner"))
+    scanp.add_argument("--scanner", choices=[x['scanner_name'] for x in scanners], default=None, help=_("define the intended scanner"))
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scanner"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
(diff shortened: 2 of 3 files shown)


atomic
```diff
@@ -395,12 +395,16 @@ def create_parser(atomic):
             ('yes', 'true', 't', 'y', '1') else False
     scanners = get_scanners()
     atomic_config = get_atomic_config()
-    if atomic_config is None or atomic_config.get('default_scanner') is None:
-        sys.stderr.write("\nYou must have a default scanner defined in /etc/atomic.conf. Use --list to see "
-                         "what scanners are available for your system.\n")
+    if atomic_config is None or (atomic_config.get('default_scanner') is None and len(scanners) > 1):
+        sys.stderr.write("\nBecause you have multiple scanners configured for your system, you must have a default "
+                         "scanner defined in /etc/atomic.conf. Use --list to see "
```
rhatdan: You have multiple scanners configured on your system, but you did not define a default in /etc/atomic.conf.  You need to specify the scanner you wish to execute.  Use --list to show all installed scanners.

atomic
```diff
@@ -411,6 +415,7 @@ def create_parser(atomic):
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scanner"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
     scanp.add_argument("--verbose", action='store_true', default=False, help=_("Show more output from scanning container"))
+    scanp.add_argument("--default_scanner", action='store', default=default_scanner, help=argparse.SUPPRESS)
     scan_group.add_argument("--rootfs", nargs='?', default=[], action='append', help=_("Rootfs path to scan"))
```
rhatdan: I don't like this, either the specify in the config the default scanner, or they should always specify the scanner they want to run.

baude: Do you not like the concept of the hidden parseargs or do you not like that if there is only one, i temporarily set it to the default?

rhatdan: I am not sure what you are trying to do with this.  I would specify the default_scanner in the config file or force the user to select the scanner they want to use.  I don't see what an option on the command line for default-scanner would do.

baude: Further explanation ...  

1.  As of today, if there is no default scanner declared, any subcommand of atomic including scan will fail with an error message pointing that out.  Obviously that can be fixed.
2. With the plan to have the scanner images themselves deliver the configuration file to /etc/atomic.d/ through the the install command, there will be no automatic editing of /etc/atomic.conf to set a default scanner.
3. Therefore, my feeling was that if there was only *one* scanner installed, it could become the "default" even though it was not declared so in /etc/atomic.conf.
4. The argparse option for default_scanner is hidden and never exposed to the users.  It is populated with a default scanner name based on the following:
  * if a default scanner is declared in /etc/atomic.conf, it is the default  
  * if only one scanner is installed, it becomes the default.
  * if more than one scanner is installed but not default defined in /etc/atomic.conf, it errors out with a meaningful message to set the default.

rhatdan: That makes sense, but why not just define an default_scanner in Scanner() and not define something in the parser.

baude: I certainly can.  I didnt in this version because I was altering as little code as possible.  I will rework outside the context of parseargs....

baude:
@rhatdan is this more what you had in mind?

Atomic/scan.py
```diff
@@ -34,32 +36,62 @@ def get_scan_info(scanner, scan_type):
...
         # Set debug bool
         self.set_debug()
 
+        set_default_scanner()
+
+        # If no scanners are defined
+        if len(self.scanners) < 1:
+            count = 1
+            raise ValueError("No scanners are configured for your system.")
+        # If a scanner is defined in parseargs
```
rhatdan: Should do this in set_default_scanner.

Atomic/scan.py
```diff
@@ -34,32 +36,62 @@ def get_scan_info(scanner, scan_type):
...
+        elif self.args.scanner is not None:
+            count = 3
+            self.scanner = self.args.scanner
```
rhatdan: What is with the counts?  Is this for debugging?

Atomic/scan.py
```diff
@@ -34,32 +36,62 @@ def get_scan_info(scanner, scan_type):
...
+        # If only one scanner is defined and no default is defined
+        elif len(self.scanners) == 1 and self.default_scanner is None and self.args.scanner is None:
+            count = 4
+            self.scanner = self.scanners[0]['scanner_name']
+        # If more than one scanner is defined but no default
```
rhatdan: In setup_default_scanner, you should just set this

if len(self.scanners) == 1:
     self.default_scanner = THISSCANNER>

Atomic/scan.py
```diff
@@ -34,32 +36,62 @@ def get_scan_info(scanner, scan_type):
...
+        elif len(self.scanners) > 1 and self.default_scanner is None:
+            count = 5
+            raise ValueError("Because you have multiple scanners configured for your system, you must have a default "
+                             "scanner defined in /etc/atomic.conf. Use --list to see what scanners are "
+                             "available for your system.")
+        # If no scanner is defined by parseargs, but there is a default scanner
+        elif self.args.scanner is None and self.default_scanner is not None:
+            count = 6
+            self.scanner = self.default_scanner
+
```
rhatdan: No need for this check.  I would set self.scanner to self.default_scanner, if self.args.scanner is not set.

Atomic/scan.py
baude: yeah, my fail

Atomic/scan.py
baude: Well it is possible that if not default is defined, self.default_scanner can == None.

rhatdan: Right and if user does not define scanner, then you fail.

baude:
Included comments

baude:
@rhatdan good 'nuff?

Atomic/scan.py
```diff
@@ -34,32 +36,61 @@ def get_scan_info(scanner, scan_type):
...
+        def set_default_scanner():
+            # If no scanners are defined
+            if len(self.scanners) < 1:
+                count = 1
+                raise ValueError("No scanners are configured for your system.")
+            elif len(self.scanners) == 1:
+                self.default_scanner = self.scanners[0]['scanner_name']
+            else:
+                self.default_scanner = self.atomic_config['default_scanner']
+
```
rhatdan: Do we need an error check to make sure default_scanner is in the self.scanners here?

rhatdan: Should also set self.scanner = self.default_scanner here.

Atomic/scan.py
```diff
@@ -34,32 +36,61 @@ def get_scan_info(scanner, scan_type):
...
 
+        set_default_scanner()
+
+        # If a scanner is defined in parseargs
+        if self.args.scanner is not None:
+            self.scanner = self.args.scanner
+        # If only one scanner is defined and no default is defined
+        elif len(self.scanners) == 1 and self.default_scanner is None and self.args.scanner is None:
+            self.scanner = self.scanners[0]['scanner_name']
+        # If more than one scanner is defined but no default
```
rhatdan: Should set this default in set_default_scanner, then override it above, eliminating this line altogether.

Atomic/scan.py
```diff
@@ -34,32 +36,61 @@ def get_scan_info(scanner, scan_type):
...
+        elif len(self.scanners) > 1 and self.default_scanner is None:
+            raise ValueError("Because you have multiple scanners configured for your system, you must have a default "
+                             "scanner defined in /etc/atomic.conf. Use --list to see what scanners are "
+                             "available for your system.")
+        # If no scanner is defined by parseargs, but there is a default scanner
+        else:
+            self.scanner = self.default_scanner
+
```
rhatdan: This should become 
if self.scanner is None:
         raise ValueError("You must specify a scanner to run or set a default scanner in ...)

rhatdan:
Lets get this merged on Monday so we can finally cut atomic-1.10.

Atomic/scan.py
```diff
@@ -34,32 +36,60 @@ def get_scan_info(scanner, scan_type):
...
+            # If more than one scanner is defined but no default
+            if len(self.scanners) > 1 and self.default_scanner is None and self.args.scanner is None:
+                raise ValueError("Because you have multiple scanners configured for your system, you must have a default "
+                                 "scanner defined in /etc/atomic.conf. Use --list to see what scanners are "
+                                 "available for your system.")
+            # Check if the scanner name is valid
+            if self.scanner not in [x['scanner_name'] for x in self.scanners]:
+                raise ValueError("Unknown scanner '{}' defined in {}".format(self.scanner, util.ATOMIC_CONF))
+
         if self.args.debug:
```
rhatdan: How about
```
        if self.default_scanner and self.default_scanner not in scanner_names:
	    raise ValueError("The default scanner '{}' defined in /etc/atomic.conf does not have a configuration "
                             "file in /etc/atomic.d.".format(self.default_scanner))
```

Atomic/scan.py
```diff
@@ -34,32 +36,60 @@ def get_scan_info(scanner, scan_type):
...
             self.debug = True
 
         # Set debug bool
         self.set_debug()
 
+        set_default_scanner()
+
```
rhatdan: set_scanner()

Atomic/scan.py
```diff
@@ -342,13 +372,13 @@ def get_scan_type(self):
...
-                             format(self.args.scan_type, self.args.scanner))
+                             format(self.args.scan_type, self.scanner))
 
     def print_scan_list(self):
         if len(self.scanners) == 0:
             util.write_out("There are no scanners configured for this system.")
             sys.exit(0)
-        default_scanner = self.atomic_config.get('default_scanner')
+        default_scanner = self.default_scanner
         if default_scanner is None:
```
rhatdan: Go back to original.  No reason to have self.default_scanner

Atomic/scan.py
```diff
@@ -26,6 +26,8 @@ def __init__(self):
         self.scanners = util.get_scanners()
         self.debug = False
         self.rootfs_mappings = {}
+        self.default_scanner = None
+        self.scanner = None
```
rhatdan: Remove self.default_scanner.

baude:
updated ....

Atomic/scan.py
```diff
@@ -34,32 +35,78 @@ def get_scan_info(scanner, scan_type):
...
+
+            ## If more than one scanner is defined but no default
+            #if len(self.scanners) > 1 and self.default_scanner is None and self.args.scanner is None:
+            #    raise ValueError("Because you have multiple scanners configured for your system, you must have a default "
+            #                     "scanner defined in /etc/atomic.conf. Use --list to see what scanners are "
+            #                     "available for your system.")
+            ## Check if the scanner name is valid
+            #if self.scanner not in [x['scanner_name'] for x in self.scanners]:
+            #    raise ValueError("Unknown scanner '{}' defined in {}".format(self.scanner, util.ATOMIC_CONF))
+
```
rhatdan: Can you remove all of these comments?

baude:
@rhatdan sorry about that, removed commented code now.

rhatdan:
@rh-atomic-bot r+

PR #385, opened 2016-05-16
[merged] Modify

rhatdan:


rhatdan:
I would like this pull request to replace https://github.com/projectatomic/atomic/pull/379

rhatdan:
@mvollmer @rhvgoyal PTAL

rhatdan:
@rhvgoyal Can docker-storage-setup work with btrfs or aufs?

tests/unit/test_util_sh.py
```diff
@@ -0,0 +1,74 @@
...
+                         'VAR="VAL1"\nVAR="VAL2"\n')
+
+        # Setting on partial line is recognized
+        self.assertEqual(util.sh_modify_var_in_text('VAR="val"', "VAR", uppercasify),
+                         'VAR="VAL"')
+
+        # Setting with extra whitespace is recognized
+        self.assertEqual(util.sh_modify_var_in_text('   VAR  =  "val"  \n', "VAR", uppercasify),
+                         'VAR="VAL"\n')
+
```
mvollmer: This is actually not valis bourne shell syntax.  Should we still recognize it?

Atomic/storage.py
```diff
@@ -20,6 +20,9 @@
     from atomic import Atomic
 
 class Storage(Atomic):
+    dss_conf = "/etc/sysconfig/docker-storage-setup"
+    dss_conf_bak = "/etc/sysconfig/docker-storage-setup" + ".bkp"
```
mvollmer: This is better written as `dss_conf_bak =  dss_conf + ".bkp"`.  (This was in my original code, sorry.)

Atomic/storage.py
```diff
@@ -38,6 +41,29 @@ def reset(self):
...
+            if os.path.exists(self.dss_conf_bak):
+                os.remove(self.dss_conf_bak)
+
+    def _add_device(self, devices):
+        util.sh_modify_var_in_file(self.dss_conf, "DEVS",
+                                   lambda old: util.sh_set_add(old, devices))
+
+    def _driver(self, driver):
+        util.sh_modify_var_in_file(self.dss_conf, "STORAGE_DRIVER",
+                                   lambda old: util.sh_set_add(old, [driver]))
```
mvollmer: `STORAGE_DRIVER` is not a set, so the modification function should be simply `lambda old: driver`.

mvollmer:
Tested manually with Cockpit, works.

mvollmer:
> Tested manually with Cockpit, works.

I have only tested `--add-device`.

rhvgoyal:
@rhatdan no, docker-storage-setup does not support btrfs or aufs graph drivers. Only devicemapper and overlay are supported.

mvollmer:
> I would like this pull request to replace #379

Ok.  The "add" integration test from #379 might still be interesting.

Atomic/storage.py
rhatdan: Fixed.

rhatdan:
@mvollmer @rhvgoyal PTAL

mvollmer:
> @mvollmer @rhvgoyal PTAL

This comment still applies, I think: 
> STORAGE_DRIVER is not a set, so the modification function should be simply `lambda old: driver`.

rhatdan:
@mvollmer missed that one.  Check it out now.

Atomic/storage.py
```diff
@@ -38,6 +41,29 @@ def reset(self):
...
+        try:
+            shutil.copyfile(self.dss_conf, self.dss_conf_bak)
+            if len(self.args.devices) > 0:
+                self._add_device(self.args.devices)
+            if self.args.driver:
+                self._driver(self.args.driver)
+            if util.call(["docker-storage-setup"]) != 0:
+                os.rename(self.dss_conf_bak, self.dss_conf)
+                util.call(["docker-storage-setup"])
+                raise ValueError("Not all devices could be added")
```
mvollmer: The error message only makes sense for `--add-device`.  We might not need a message at all since dpcker-storage-setup should have explained itself already.

mvollmer:
Tested --driver manually as well.  Works as expected, except for the error message.

rhatdan:
Changed the error message to just show that docker-storage-setup failed, also added an exception block to bring back /etc/sysconfig/docker-storage-setup.

mvollmer:
LGTM, thanks!

rhatdan:
@rh-atomic-bot r+

Issue #386, opened 2016-05-17
Wrong usage exits with code zero


mvollmer:
```
# atomic foo bar
/usr/bin/atomic: invalid choice: [...]
# echo $?
0
```

rhatdan:
https://github.com/projectatomic/atomic/pull/387 
 Should fix this issue.

PR #387, opened 2016-05-17
[merged] Exit with proper exit code

rhatdan:


rhatdan:
Fixes https://github.com/projectatomic/atomic/issues/386

rhatdan:
@baude PTAL

baude:
You will likely want something like this to protect from tuple errors when the exit is legit.  This is why the tests are failing.

+        rc = 0 if len(e.args) < 1 else e.args[0]
+        sys.exit(rc)

atomic
```diff
@@ -613,10 +613,11 @@ if __name__ == '__main__':
             need_root()
     except Exception as e:
         sys.stderr.write("%sn" % str(e))
-    except SystemExit:
+    except SystemExit as e:
         # Overriding debug args to avoid a traceback from sys.exit()
         if 'args' in locals():
             args.debug = False
+        sys.exit(e.args[0])
```
cgwalters: Confused by `e.args[0]` - isn't that `e.code`?

rhatdan:
@baude I went with @cgwalters fix, does this fix the problem you were seeing?

baude:
yes'sir.

rhatdan:
@rh-atomic-bot r+

PR #389, opened 2016-05-23
[merged] atomic: read default values from manifest file

giuseppe:
Introduce a new "manifest.json" file where it is possible to store
more information related to the system container.

A first need of this file is to store default values when
--set=NAME=VALUE is not provided.

The Hello World example was modified to take advantage of this new feature:

https://github.com/giuseppe/atomic-oci-containers/blob/master/hello-world/manifest.json

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -1142,8 +1142,18 @@ class Atomic(object):
         if extract_only:
             return True
 
-        values["DESTDIR"] = destination
-        values["NAME"] = name
+        # When installing a new system container, set values in this order:
+        #
+        # 1) What comes from manifest.json, if present, as default value.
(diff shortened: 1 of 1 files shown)


rhatdan:
Where is this file stored? Are you shipping it inside of the container bundle?

giuseppe:
yes, the file is stored under `/exports` as all other files

rhatdan:
Ok I have not read this code, sot this is just atomic command reading into the container image for files under /exports.

giuseppe:
yes, correct, all the logic to handle files under `/exports` is in the atomic command.  Both runc and systemd have no direct access to them

rhatdan:
@rh-atomic-bot r+

PR #393, opened 2016-05-27
[merged] Fix live mount, pass over ostree mount errors

willmtemple:
`atomic mount` attempted to use an OSTree system before the older DockerMount, but an exception encountered in OSTree() would prevent the attempt, resulting in a runtime error. This patch corrects that behavior.

Additionally, DockerMount.mount() was called with the wrong parameters, causing any `--live` mount to produce an error.

All current tests related to atomic mount now pass.

*However*, I don't like this solution. The nested try/except clauses are ugly, and I think that we could explore a design for the mount system that will reduce the complexity of this code and make the decision tree it follows easier to understand.

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -87,13 +87,16 @@ class Mount(Atomic):
 
     def mount(self):
         try:
-            d = OSTreeMount(self.mountpoint, live=self.live, shared=self.shared)
-            if d.mount(self.image, self.live):
-                return
+            try:
+                d = OSTreeMount(self.mountpoint, live=self.live, shared=self.shared)
+                if d.mount(self.image, self.options):
+                    return
+            except (MountError,NoDockerDaemon):
+                pass
 
             d = DockerMount(self.mountpoint, self.live)
             d.shared = self.shared
-            d.mount(self.image, self.live)
+            d.mount(self.image, self.options)
 
             # only need to bind-mount on the devicemapper driver
             if self.d.info()['Driver'] == 'devicemapper':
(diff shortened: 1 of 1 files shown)


giuseppe:
LGTM

Atomic/mount.py
```diff
@@ -87,13 +87,16 @@ def set_args(self, args):
 
     def mount(self):
         try:
-            d = OSTreeMount(self.mountpoint, live=self.live, shared=self.shared)
-            if d.mount(self.image, self.live):
-                return
+            try:
+                d = OSTreeMount(self.mountpoint, live=self.live, shared=self.shared)
+                if d.mount(self.image, self.live):
+                    return
```
rhatdan: Should this be self.options?

cgwalters:
@willmtemple You should use `git commit --fixup` for addressing review comments on repos (like this one) which use Homu.  That way they get cleanly squashed when merging, but are visible in the PR history.

You might find this useful:
https://github.com/cgwalters/homegit/blob/master/bin/git-addfixup

cgwalters:
You wrote a good commit message except it's missing a bit of information that seems rather pertinent: What's "an exception encountered in OSTree()" ?  What was the actual error?

willmtemple:
@cgwalters The error is that OSTree mounts take priority over regular docker mounts and the OSTreeMount class raises MountError. The handler for MountError is considered fatal rather than attempting a DockerMount.

OSTreeMount was throwing two different errors: "Containers and images managed through OSTree do not support --live." and "xattr nrequired to mount OSTree images," but I'm not familiar with the OSTree `atomic mount` backend as it's undocumented.

rhatdan:
@jlebon Tests are not completing.

jlebon:
bot, retest this please

rhatdan:
@rh-atomic-bot r+

PR #395, opened 2016-05-29
Add atomic images --noheading

rhatdan:
In certain cases like piping to grep, you do not want to include headings


rhatdan:
@baude PTAL

baude:
LGTM

baude:
@rh-atomic-bot r+

PR #400, opened 2016-05-31
[merged] atomic: More completely handle failure to import OSTree

cgwalters:
GI raises `ValueError` apparently, not `ImportError`.  So rework
things to be nested and handle the cases of:

 - No GI
 - GI but no OSTree
 - OSTree

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -15,9 +15,12 @@ from string import Template
 
 try:
     import gi
-    gi.require_version('OSTree', '1.0')
-    from gi.repository import Gio, GLib, OSTree
-    OSTREE_PRESENT = True
+    try:
+        gi.require_version('OSTree', '1.0')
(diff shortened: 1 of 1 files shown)


lsm5:
seems to get rid of the `ValueError: OSTree` not found error for me

giuseppe:
thanks for looking into this and fix my old version.  It looks fine and I've done some extra testing.

@rh-atomic-bot r+ fdbe345

rhatdan:
Is this something we need to get into 7.2.5?

giuseppe:
this doesn't affect Atomic Host

rhatdan:
Right but is it something we see blowing up in RHEL when users run atomic CLI?

giuseppe:
correct, this will cause an error as soon as we import Atomic.atomic

rhatdan:
Ok I will push v1.10.3

PR #401, opened 2016-06-01
[merged] More system containers stuff

giuseppe:
Add some tests for the system containers + some fixes

The only new feature is that the OSTree repo is automatically created now if it is not found


rhatdan:
@rh-atomic-bot  r+

PR #402, opened 2016-06-01
[merged] Error on V1 Registries

baude:
Skopeo only works with V2 registries.  So any time we use
skopeo to perform a remote probe of an image, we should error
if the remote registry is not V2.  Added a def to check
the registry looking at the v1/_ping.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -11,6 +11,11 @@ from yaml import load as yaml_load
 import tempfile
 import shutil
 import re
+import requests
+try:
+    from urlparse import urlparse
+except:
+    from urllib.parse import urlparse #pylint: disable=no-name-in-module,import-error
 
 """Atomic Utility Module"""
 
@@ -209,7 +214,9 @@ def skopeo_inspect(image, args=[]):
     except OSError:
         raise ValueError("skopeo must be installed to perform remote inspections")
     if results.return_code is not 0:
-        raise ValueError(results.stderr)
+        # Need to check if we are dealing with a v1 registry
+        check_v1_registry(image)
+        raise ValueError("Unable to interact with this registry: {}".format(results.stderr))
     else:
         return json.loads(results.stdout.decode('utf-8'))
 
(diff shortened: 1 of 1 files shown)


Atomic/util.py
```diff
@@ -209,7 +214,9 @@ def skopeo_inspect(image, args=[]):
     except OSError:
         raise ValueError("skopeo must be installed to perform remote inspections")
     if results.return_code is not 0:
-        raise ValueError(results.stderr)
+        # Need to check if we are dealing with a v1 registry
+        is_v1_registry(image)
+        raise ValueError("Unable to interact with this registry: {}".format(results.stderr))
     else:
```
rhatdan: Why do you handle this error differently?  
Could  you change function call to check_v1_registry, or try_v1_registry.  
is_ sounds like a boolean check.

Atomic/util.py
```diff
@@ -227,6 +234,7 @@ def skopeo_layers(image, args=[], layers=[]):
         args = ['skopeo', 'layers'] + args + [image] + layers
         r = subp(args, cwd=temp_dir)
         if r.return_code != 0:
+            is_v1_registry(image)
             raise ValueError(r.stderr.decode(sys.getdefaultencoding()))
     except OSError:
```
rhatdan: Above 
```
 raise ValueError("Unable to interact with this registry: {}".format(results.stderr))
```
Below

```
raise ValueError(r.stderr.decode(sys.getdefaultencoding()))
```
Why not the same?

baude: Just an oversight as the second def was last touched by someone else.  Made the same now.

Atomic/util.py
baude: roger that

Atomic/util.py
```diff
@@ -11,6 +11,11 @@
 import tempfile
 import shutil
 import re
+import requests
+try:
+    from urlparse import urlparse
+except:
+    from urllib.parse import urlparse #pylint: disable=no-name-in-module
```
jlebon: Try `#pylint: disable=no-name-in-module,import-error`.

runcom:
sad LGTM :(

rhatdan:
@rh-atomic-bot r+

PR #403, opened 2016-06-02
Fixes

runcom:
This PR does two things:

- add generated `dockertar-sha256-helper` to `.gitignore`
- exit with `1` when `docker-storage-setup --reset` doesn't work, this can happen if the version if too old and there's no `--reset` flag. Old versions of docker-storage-setup exited by printing the help message + exit 0. Catch the help message and exit with 1 if it's the case - bug report here https://bodhi.fedoraproject.org/updates/atomic-1.10-2.git1d6aecf.fc24#comment-440048


runcom:
@rhatdan @baude PTAL

rhatdan:
Why not just require a version of docker-storage-setup that has the --reset flag?

runcom:
docker-storage-setup is shipped along with docker and it's not a package on its own - the docker-storage-setup needed for atomic-1.10.3 would be in docker-1.10.3 - is it fine to add a Require in atomic.spec for Fedora for docker >= 1.10.3? I did this not to break older dockers

rhatdan:
I would say yes, but OpenShift may still force us to drop back to docker-1.9.  Other stuff atomic storage export/import will also break with docker < 1.10.3

runcom:
let's hold this up for a while - I've still not built atomic-1.10.3 for Fedora (just waiting on this)

PR #405, opened 2016-06-03
[merged] System containers more tests

giuseppe:
a few more tests

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -32,6 +32,7 @@ from . import util
 import requests
 from .util import NoDockerDaemon
 import shutil
+from .atomic import OSTREE_PRESENT as OSTREE_PRESENT
 
 """ Module for mounting and unmounting containerized applications. """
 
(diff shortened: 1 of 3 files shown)


Atomic/mount.py
```diff
@@ -87,12 +88,9 @@ def set_args(self, args):
...
         try:
-            try:
-                d = OSTreeMount(self.mountpoint, live=self.live, shared=self.shared)
-                if d.mount(self.image, self.options):
-                    return
-            except (MountError,NoDockerDaemon):
-                pass
+            d = OSTreeMount(self.mountpoint, live=self.live, shared=self.shared)
+            if d.mount(self.image, self.options):
+                return
```
rhatdan: Are you reverting a previous patch?

giuseppe: yes, I moved the failing exception after we check that there is a system container with that name.  It is needed so I can properly show an error message when users specify --live or --shared with a system container.

To avoid reintroducing the issue, I have added some more tests for --live and --shared when used on a regular  Docker container.

rhatdan:
Ok now I see.  Had to look at original source to see what you were doing,  github format made this confusing.

LGTM

Please update the patch

giuseppe:
updated here :arrow_up:

rhatdan:
@rh-atomic-bot r+

PR #407, opened 2016-06-03
[merged] Define openscap image

baude:
The fully qualified name of the openscap image is:
  registry.access.redhat.com/rhel7/openscap

### DIFF
--- a/atomic.d/openscap
+++ b/atomic.d/openscap
@@ -1,6 +1,6 @@
 type: scanner
 scanner_name: openscap
-image_name: openscap
+image_name: rhel7/openscap
 default_scan: cve
 scans: [ 
       { name: cve,


atomic.d/openscap
```diff
@@ -1,6 +1,6 @@
 type: scanner
 scanner_name: openscap
-image_name: openscap
+image_name: registry.access.redhat.com/rhel7/openscap
 default_scan: cve
```
rhatdan: I don't think you should have the registry name in here.  Just the rhel7/openscap.  That way people could use it with internal registries.

baude: On the other hand, not doing so could allow someone to spoof and image in there.  You decide, I'll change it.

rhatdan: Well we would rely on signing for that,  I would say remove the registry.

baude:
Removed registry

baude:
r+

baude:
@rh-atomic-bot r+

PR #408, opened 2016-06-04
[merged] Atomic/scan.py: Set default scanner

baude:
After reworking the logic in set_scanner(), we missed the
condition that if a default scanner is defined in atomic.conf,
it should be set as self.scanner unless otherwise overriden.

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -43,13 +43,14 @@ class Scan(Atomic):
 
             if len(self.scanners) == 1:
                 self.scanner = self.scanners[0]['scanner_name']
-
             # If a scanner is defined in parseargs
-            if self.args.scanner:
-                    self.scanner = self.args.scanner
-
-            if self.scanner is None:
-                raise ValueError("You must specify a scanner")
+            elif self.args.scanner:
+                self.scanner = self.args.scanner
+            # Use default
+            elif default_scanner is not None:
+                self.scanner = default_scanner
+            elif self.scanner is None:
+                raise ValueError("You must specify a scanner (--scanner) or set a default in /etc/atomic.conf")
 
             # Check if the scanner name is valid
             if self.scanner not in [x['scanner_name'] for x in self.scanners]:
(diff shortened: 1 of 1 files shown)


Atomic/scan.py
```diff
@@ -48,7 +49,10 @@ def set_scanner():
             if self.args.scanner:
                     self.scanner = self.args.scanner
 
-            if self.scanner is None:
+            # Use default
+            if self.default_scanner is not None:
+                self.scanner = self.default_scanner
+            else:
                 raise ValueError("You must specify a scanner")
```
rhatdan: This is wrong, the user specified a scanner and you just over wrote it.

Atomic/scan.py
```diff
@@ -370,12 +375,11 @@ def print_scan_list(self):
         if len(self.scanners) == 0:
             util.write_out("There are no scanners configured for this system.")
             sys.exit(0)
-        default_scanner = self.atomic_config['default_scanner']
-        if default_scanner is None:
+        if self.default_scanner is None:
             default_scanner = ''
```
rhatdan: I thought we were staying away from a self.default_scanner

bkoz:
atomic scan --list now works as expected when > 1 scanner is configured. The default scanner is also picked up.

Atomic/scan.py
```diff
@@ -43,13 +43,14 @@ def set_scanner():
...
-
             # If a scanner is defined in parseargs
-            if self.args.scanner:
-                    self.scanner = self.args.scanner
-
-            if self.scanner is None:
-                raise ValueError("You must specify a scanner")
+            elif self.args.scanner:
+                self.scanner = self.args.scanner
+            # Use default
```
rhatdan: This is still wrong.

It should be 
```

if self.args.scanner:
     self.scanner = self.args.scanner
elif len(self.scanners) == 1:
                 self.scanner = self.scanners[0]['scanner_name']
elif self.default_scanner:
                self.scanner = self.default_scanner
else:
                raise ValueError("You must specify a scanner") 
```

baude:
@rh-atomic-bot r+

PR #410, opened 2016-06-06
[merged] Used system images

giuseppe:
Add an indicator for used system images as well + some other small fixes


rhatdan:
@rh-atomic-bot r+

PR #415, opened 2016-06-08
[merged] Atomic/scan: fix image naming error

chuanchang:
The image naming looks like 'sha256:XXXXX' in atomic, but set_scanner() in Atomic/scan.py replaces '**:**' with '**-**', the result is atomic scan can't match any image.

```shell
# atomic images|grep rhel7
  registry.access.redhat.com/rhel7/rhel-tools           latest   sha256:00211   2016-05-06 17:49   1.27 GB       
  registry.access.redhat.com/rhel7                      latest   sha256:bf203   2016-05-05 12:43   203.43 MB     

# atomic scan --verbose sha256:bf203
sha256-bf203442783741aad6d82b528bcfecd45f40e63c83d981eb5e644a2fa6356e60 did not match any image or container.
```

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -91,7 +91,6 @@ class Scan(Atomic):
         if len(self.args.rootfs) == 0:
             scan_list = self._get_scan_list()
             for i in scan_list:
-                i['Id'] = i['Id'].replace(":", "-")
                 self.scan_content[i['Id']] = i.get('input')
 
             # mount all the rootfs


giuseppe:
bot, test pull request

Atomic/scan.py
```diff
@@ -91,7 +91,6 @@ def set_scanner():
         if len(self.args.rootfs) == 0:
             scan_list = self._get_scan_list()
             for i in scan_list:
-                i['Id'] = i['Id'].replace(":", "-")
```
giuseppe: shouldn't this be left to handle the previous naming?  I think the condition must be changed to replace : only when it is not part of "sha256:"

baude:
This is a systemic issue and not just in scan.  To fix this properly, we need more than just this particular fix.

rhatdan:
bot, add author to whitelist

rhatdan:
@baude I agree.  Lets get this through out the atomic code.

lsm5:
@rhatdan has this been included throughout atomic yet? It's blocking rhbz#1343939, I filed rhbz#1346942 for ostree dep.

Atomic/scan.py
mpreisler: that is my impression as well.

rhatdan:
OK I will merge this and we can fix the rest later.

@rh-atomic-bot r+

lsm5:
thanks

PR #416, opened 2016-06-08
[merged] Fix scanner selection.

rhatdan:
Currently if the user selects a scanner that does not exist, but only
has one scanner, it will run a scanner different then what the user
specified,  This patch will cause the scan to fail.

Also eliminated the need for the default_scanner variable.


rhatdan:
@baude PTAL

baude:
Dan, Ok this looks good.

baude:
@rh-atomic-bot r+

PR #422, opened 2016-06-14
[merged] add "atomic ps" to show containers

giuseppe:
I'll probably need something like this for https://github.com/openshift/openshift-ansible/ to be able to query what containers are running.  In any case, this is generally a good thing to have.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -516,6 +516,8 @@ class Atomic(object):
         command += self.image
         return command
 
+    #def ps -> Atomic/ps.py
+
     #def run -> Atomic/run.py
 
     def stop(self):
--- a/atomic
+++ b/atomic
@@ -36,6 +36,7 @@ from Atomic.mount import Mount
 from Atomic.verify import Verify
 from Atomic.help import AtomicHelp
 from Atomic.run import Run
+from Atomic.ps import Ps
 from Atomic.storage import Storage
 from Atomic.atomic import NoDockerDaemon, OSTREE_PRESENT
 from Atomic.util import get_atomic_config, get_scanners, default_docker_lib
@@ -430,6 +431,20 @@ def create_parser(atomic):
     scan_group.add_argument("--all", default=False, action='store_true', help=_("scan all images (excluding intermediate layers) and containers"))
     scan_group.add_argument("--images", default=False, action='store_true', help=_("scan all images (excluding intermediate layers)"))
     scan_group.add_argument("--containers", default=False, action='store_true', help=_("scan all containers"))
+
+    # atomic ps
+    pss = subparser.add_parser(
+        "ps", help=_("list the containers"),
+        epilog="By default this shows only the running containers.")
+    pss.set_defaults(_class=Ps, func='ps')
+    pss.add_argument("-n", "--noheading", dest="heading", default=True,
+                     action="store_false",
+                     help=_("do not print heading when listing the containers"))
+    pss.add_argument("-a", "--all", action='store_true',dest="all", default=False,
+                     help=_("show all containers"))
+    pss.add_argument("--json", action='store_true',dest="json", default=False,
+                     help=_("print in a machine parseable form"))
+    add_opt(pss)
 
     # atomic stop
     stopp = subparser.add_parser(
--- a/tests/integration/test_system_containers.sh
+++ b/tests/integration/test_system_containers.sh
@@ -47,6 +47,10 @@ trap teardown EXIT
 
 ${ATOMIC} --debug install --name=${NAME} --set=RECEIVER=${SECRET} --system oci:atomic-test-system
 
+${ATOMIC} ps | grep -q "test-system"
+${ATOMIC} ps --json | grep -q "test-system"
+${ATOMIC} ps --all | grep -q "test-system"
+
 test -e /etc/systemd/system/${NAME}.service
 
 # The value we set is exported into the config file
--- a/Atomic/ps.py
+++ b/Atomic/ps.py
@@ -1,0 +1,60 @@
+import os
+import sys
+import rpm
+from filecmp import dircmp
+import json
+
+from . import util
+from . import mount
+from . import Atomic
+from dateutil.parser import parse as dateparse
+
+class Ps(Atomic):
+    def ps(self):
+        all_containers = []
+
+        # Collect the system containers
+        for i in self.get_system_containers():
+            container = i["Id"]
+            inspect_stdout = util.check_output(["runc", "state", container])
+            ret = json.loads(inspect_stdout)
+            status = ret["status"]
+            if not self.args.all and status != "running":
+                continue
+
+            image = i['Image']
+            command = ""
+            created = dateparse(ret['created']).strftime("%F %H:%M") # pylint: disable=no-member
+            all_containers.append({"type" : "systemcontainer", "container" : container,
+                                   "image" : image, "command" : command, "created" : created,
+                                   "status" : status})
+
+        # Collect the docker containers
+        for container in [x["Id"] for x in self.d.containers(all=self.args.all)]:
+            ret = self._inspect_container(name=container)
+            status = ret["State"]["Status"]
+            image = ret['Config']['Image']
+            command = u' '.join(ret['Config']['Cmd']) if ret['Config']['Cmd'] else ""
+            created = dateparse(ret['Created']).strftime("%F %H:%M") # pylint: disable=no-member
+            all_containers.append({"type" : "docker", "container" : container,
+                                   "image" : image, "command" : command,
+                                   "created" : created, "status" : status})
+
+        if self.args.json:
+            self.write_out(json.dumps(all_containers))
+            return
+
+        col_out = "{0:12} {1:20} {2:20} {3:15} {4:9}"
+        if self.args.heading:
+            self.write_out(col_out.format("CONTAINER ID",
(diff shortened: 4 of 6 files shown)


cgwalters:
Ones that are running should showup in `systemctl` (which has a DBus API and one can extract bits from the command line too), right?

giuseppe:
yes, running system containers showup in systemctl as well, but I plan to add more information to the atomic ps output, at least when --json is used, which cannot be easily retrieved from systemctl.  

For example have all the information required for having an ansible task that "will recreate the container if a different configuration is used", so that means at least exporting all the set variables for the system container.

Another difference is that "atomic ps" unifies both Docker containers and system containers.

rhatdan:
Should we list all container on the system?  IE other containers started with runc ?  runc list?
What about containers started with systemd-nspawn?  Is this just listing running containers or all containers?

rhatdan:
rkt containers? 

machinectl list

giuseppe:
what I would like is to have a way to show system containers and query their status at a higher level of what "runc list" does, as runc does not know how a system container was configured or what version of the image is used.  I thought of Ansible as a consumer of this information (with `--json` output).
By default "atomic ps" lists only running containers, you can list all containers with `--all`.

I added the support for docker containers as well so that it works like "atomic images" where only system containers and docker images are known.

Should it be an internal API only?

rhatdan:
No I like the idea, but I still think we should provide less info about other containers like runc, rkt, systemd-nspawn ...
ps --all would only work with the containers we know about I guess.

Atomic/atomic.py
```diff
@@ -516,6 +518,55 @@ def container_run_command(self):
         command += self.image
         return command
 
+    def ps(self):
```
rhatdan: Can you split this out into a separate file ps.py, rather then continuously growing atomic.py.  I eventually would like to break each command into its own py file.

atomic
```diff
@@ -431,6 +431,20 @@ def create_parser(atomic):
     scan_group.add_argument("--images", default=False, action='store_true', help=_("scan all images (excluding intermediate layers)"))
     scan_group.add_argument("--containers", default=False, action='store_true', help=_("scan all containers"))
 
+    # atomic ps
+    pss = subparser.add_parser(
+        "ps", help=_("List containers"),
+        epilog="show the installed containers")
+    pss.set_defaults(func='ps')
```
rhatdan: man docker shows
       List the containers in the local repository. By default  this  shows  only
       the running containers.

Should we add to the epilog something about showing only running containers.

atomic
```diff
@@ -431,6 +431,20 @@ def create_parser(atomic):
...
+    pss.add_argument("-n", "--noheading", dest="heading", default=True,
+                     action="store_false",
+                     help=_("do not print heading when listing the images"))
+    pss.add_argument("-a", "--all", action='store_true',dest="all", default=False,
```
rhatdan: listing the containers not images

rhatdan:
@baude PTAL

cgwalters:
I agree that systemctl not supporting JSON is an annoying gap, but OTOH this already exists: https://docs.ansible.com/ansible/systemd_module.html

Though probably even with that we do need the union of information with the container image data etc...so that argues for either having this or a dedicated ansible module for syscontainers.

So to be clear I'm not objecting to this, but I also see it as a key advantage of syscontainers that we're making use of systemd.

Atomic/atomic.py
```diff
@@ -516,6 +518,55 @@ def container_run_command(self):
...
+
+            image = i['Image']
+            command = ""
+            created = dateparse(ret['created']).strftime("%F %H:%M") # pylint: disable=no-member
+            all_containers.append({"type" : "systemcontainer", "container" : container,
+                                   "image" : image, "command" : command, "created" : created,
+                                   "status" : status})
+
+        # Collect the docker containers
+        for container in [x["Id"] for x in self.d.containers(all=self.args.all)]:
```
baude: Can you use get_containers OR get_active_containers here conditionally?  Something like?

func = self.containers if self.args.all else self.get_active_containers

The do your list comprehension using in func()]:

Atomic/atomic.py
```diff
@@ -789,7 +840,6 @@ def get_col_lengths(_images):
             _max_repo, _max_tag = get_col_lengths(_images)
             col_out = "{0:1} {1:" + str(_max_repo) + "} {2:" + str(_max_tag) + \
                       "} {3:14} {4:18} {5:14}"
-            self.write_out("")
```
baude: @rhatdan There is a util.write out and a atomic.write out.  They differ.  Should they both exist?

Atomic/atomic.py
```diff
@@ -516,6 +518,55 @@ def container_run_command(self):
...
+            image = i['Image']
+            command = ""
+            created = dateparse(ret['created']).strftime("%F %H:%M") # pylint: disable=no-member
+            all_containers.append({"type" : "systemcontainer", "container" : container,
+                                   "image" : image, "command" : command, "created" : created,
+                                   "status" : status})
+
+        # Collect the docker containers
+        for container in [x["Id"] for x in self.d.containers(all=self.args.all)]:
+            ret = self._inspect_container(name=container)
```
baude: Have you given any thought to using the .top() api from docker-py to get this sort of information?  One good thing about it is you can add ps arguments to the query which could make your ps function more powerful.  There is an example in atomic Top -> https://github.com/projectatomic/atomic/blob/master/Atomic/top.py#L171

Atomic/atomic.py
rhatdan: I don't think so.

Atomic/atomic.py
rhatdan: Seems like a good idea.

Atomic/atomic.py
giuseppe: I have moved the new functionality to a separate ps.py file

Atomic/atomic.py
giuseppe: I cannot use get_containers or get_active_containers as they already include the system containers while I want only the docker containers.  get_(active_)containers must probably be refactored to use the ps command.

tests/integration/test_system_containers.sh
```diff
@@ -31,16 +31,22 @@ ${ATOMIC} pull dockertar:/${WORK_DIR}/atomic-test-system.tar
...
+    runc delete $NAME &> /dev/null  || true
+
     # Ensure there is no systemd service left running
-    if test -e /etc/systemd/system/test-system-container.service; then
-        (systemctl stop $NAME) &> /dev/null
-        rm -rf /etc/systemd/system/test-system-container.service
-    fi
+    systemctl stop $NAME &> /dev/null || true
+    systemctl disable $NAME &> /dev/null || true
+    rm -rf /etc/systemd/system/${NAME}.service || true
```
rhatdan: I did not know about this syntax &> /dev/null.  Cool

rhatdan: rm -rf /etc/systemd/system/${NAME}.service || true

No need for || true.  -f will not error out if the file does not exist.  If you got a permission error, you would probably want to see that.

rhatdan:
Can we split these patch sets in two, one for images and fixes for testing, the other for atomic ps.

Atomic/atomic.py
rhatdan: Probably should unify on atomic.write. But lets do that as a separate patch.

giuseppe:
I moved the other patches that are not part of "atomic ps" here: https://github.com/projectatomic/atomic/pull/439

rhatdan:
rebase please.

yuqi-zhang:
So I was trying this out on a Centos-CI vagrant vm, and found that "atomic ps" needs "runc state mycontainer" from runc 1.0 release. But when I built runc from upstream, trying to install giuseppe's etcd system container fails ("atomic install --system --name=etcd gscrivano/spc-etcd").

To be more exact, the command itself succeeds, but etcd fails to start. Looking into journalctl, it complains that "open /run/runc/etcd/state.json: no such file or directory" (obviously was never created). If I had to guess, this is due to runc 1.0 splitting the create/start process, and thus just running "atomic install --system" is no longer enough to set up the system container? (In previous versions of runc, this whole process worked fine without having to directly interact with runc)

I could be misunderstanding some of this. Any thoughts are appreciated.

rhatdan:
@yuqi-zhang Please open a separate issue on the atomic install --system issue or better yet a bugzilla.

rhatdan:
We need to fix the tests.

```
************* Module Atomic.ps
E: 10, 0: Unable to import 'dateutil.parser' (import-error)
Makefile:28: recipe for target 'pylint-check' failed
make: *** [pylint-check] Error 2
```

I would like to see us add some flexibility like @baude added for atomic top to show different fields from the ps command.  But lets get the test fixed and we can merge.

yuqi-zhang:
The install seems to be an issue with how images were set up, will look into. Since giuseppe isn't here, I'll also take a look at tests.

jlebon:
bot, retest this please

jlebon:
I installed `python-dateutil` and `python3-dateutil` on the tester to give the PR a try, but I think they should probably be added to requirements.txt instead.

rhatdan:
Ok I am going to merge, but I would like to see this extended to allow users to specify which fields they would like to see in the ps output.

rhatdan:
@rh-atomic-bot r+

PR #425, opened 2016-06-16
[merged] Card container_security_114 - Add delete image support

jwhonce:
* Add support for calling skopeo to delete v2 images from v2 registry
* Update documentation and bash tab expansion
* Cleanup code and comments
* Update integration tests

This is a breaking change to the atomic images command CLI.

Signed-off-by: Jhon Honce <jhonce@redhat.com>


atomic
```diff
@@ -302,18 +303,35 @@ def create_parser(atomic):
...
+    list_parser.add_argument("-n", "--noheading", dest="heading", default=True,
+                     action="store_false",
+                     help=_("do not print heading when listing the images"))
+
+    # atomic images delete
+    delete_parser = images_subparser.add_parser("delete",
+                                                help=_("mark image for deletion"),
+                                                epilog="Marks image. registry garbage-collection "
+                                                "when invoked will recover used disk space")
+    delete_parser.set_defaults(_class=Delete, func='delete_image')
```
rhatdan: I would think this would delete the local image not the image at the registry?  
What do you think of ?
```
atomic image delete IMAGE

atomic image delete --remote IMAGE
``

**Comment by rhatdan** [comment]
I prefer the breaking change.

You need to update the man pages and the bash completions.

@baude @cgwalters @runcom What do you think?

File: Atomic/util.py [hunk 2]
```diff
@@ -245,11 +264,23 @@ def skopeo_layers(image, args=[], layers=[]):
...
+    :return: Raise exception unless we can confirm v2 registry access
+    """
+    registry = list(urlparse(image))
+    registry[0] = "https"
+    registry[2] = "/v2/"
+    try:
+        if requests.get(urlunparse(registry)).status_code == 200:
+            pass
+    except requests.exceptions.SSLError:
+        registry[0] = "http"
```
cgwalters: I'm not sure about automatically falling back to `http`.  Shouldn't we have some sort of per-registry flag like docker has?  I personally like the "remote" model like git and ostree have...might be interesting to try doing that here for docker images.

runcom: +1 (sadly docker has never been keen on adding _remotes_ via `docker  remote`)

rhatdan:
Jon we need a rebase.  Also I think we need to have the ability to delete image locally and remote.

jwhonce:
@rhatdan +1 on the rebase, I'm researching ideas for the --remote delete now following @baude  's work.

rhatdan:
@jwhonce Needs rebase, and then lets work to get this merged.

giuseppe:
we are also considering how to delete system containers images instead of deleting directly the OSTree branch.  Having something like `atomic images --delete IMAGE` for local images and `atomic images --delete --remote IMAGE` for remote ones would probably help us.  What do you think?

@yuqi-zhang

atomic
```diff
@@ -306,18 +307,39 @@ def create_parser(help_text):
...
+    list_parser.set_defaults(func='images')
+    list_parser.add_argument("--prune", action="store_true",
                          help=_("prune dangling images"))
-    imagesp.add_argument("-n", "--noheading", dest="heading", default=True,
-                         action="store_false",
-                         help=_("do not print heading when listing the images"))
+
+    list_parser.add_argument("-n", "--noheading", dest="heading", default=True,
+                     action="store_false",
+                     help=_("do not print heading when listing the images"))
```
rhatdan: Fix the indentation.

rhatdan:
Yes I would prefer default to local images and then use --remote to handle this patch.

yuqi-zhang:
+1 for the local/remote split.

Locally, an "ostree refs --delete" for system container images and a "docker rmi" for docker images will probably do the trick. Although ostree refs --delete only deletes one tag, so atomic images --prune would likely have to happen for a complete cleanup (perhaps not necessary as part of the delete command?)

Skopeo can handle remote deletions, and I noted that after skopeo marks an image for deletion, the user must also run docker registry garbage collection to reclaim the disk space. Not sure if that's relevant for this delete command.

I was also thinking of making an "atomic rmi" to handle image deletion locally before I saw this PR, but delete should probably handle both cases.

jwhonce:
@rhatdan please review and/or merge.  Thanks.

Atomic/delete.py
```diff
@@ -0,0 +1,61 @@
...
+                sys.stderr.write("Failed to mark Image {} for deletion: {}\n".format(img, e))
+                results = 2
+        return results
+
+    def _delete_local(self, targets):
+        results = True
+        for target in targets:
+            try:
+                util.check_call([self.docker_binary(), "rmi", target])
+            except subprocess.CalledProcessError:
```
rhatdan: Can you use the docker python bindings.

self.d.remove_image(target)

Also I think we might want to add support for SystemContainers, although I am not sure how you remove these images.

@giuseppe PTAL

giuseppe:
I have some doubts about the breaking cli change.  I think `atomic images` should still work as now without requiring the additional `list` command.  We already have an additional command `--prune` so I think it will be fine to have `atomic images --delete` to handle delete.  Having `--delete` instead of splitting images into the new commands `list` and `delete` will: 1) require less typing, 2) won't break the existing interface, 3) `atomic list --prune` won't look so strange (unless it is moved under delete).

Deleting a system container is just a matter of deleting the OSTree branch, I can do that after this patch is merged, but if you prefer to include the change here then it will be something like (untested):

```
def rmi(self, img)
    if not repo:
        return
     ref = OSTree.parse_refspec(imagebranch)
     repo.set_ref_immediate(ref[1], ref[2], None)
```

jwhonce:
bot, please retest

jwhonce:
bot, retest this please

jwhonce:
retest this please

jwhonce:
@rhatdan How about we merge this work and I do a follow on PR for the OSTree work?  The rmi is now being done via API vs. docker call.

giuseppe:
if we keep the breaking change, could `--prune` be promoted later to be a command on its own?  Or maybe it be can be moved under `delete`?

jwhonce:
@giuseppe I would be happy to move --prune under delete in a follow-on PR.  I would like to get this work committed and not have to keep rebasing. :+1:

giuseppe:
@jwhonce sure, it is fine for me. @rhatdan what do you think?

rhatdan:
@rh-atomic-bot r+

PR #426, opened 2016-06-20
[merged] document ways of connecting to docker engine

### LINKED ISSUE #41: atomic run doesn't work with Docker TCP connection
Hi, atomic run ... doesn't work if the docker daemon was started with TCP connection

more /etc/sysconfig/docker
....
OPTIONS='--insecure-registry dockerregistry.jaja.corp:5000 --selinux-enabled -H [IP]:5320'
.....

-bash-4.2# atomic run registry.access.redhat.com/rhel7/rhel-tools
Unable to communicate with docker daemon: [Errno 2] No such file or directory

If I change to unix socket it works OK, 

Bye.

TomasTomecek:
Resolves #41 

Not sure if I put the text to correct place/manpage. Feel free to suggest a better place.

~~Unfortunately I wasn't able to find any docs related to this, I was getting the info from code of docker-py.~~ The environment variables are same as those for docker machine: https://docs.docker.com/machine/reference/env/


jlebon:
This is probably a better link to the upstream docs: https://docs.docker.com/engine/security/https/
Would be nice to have it in the man page as well.

TomasTomecek:
good point; updated

rhatdan:
@rh-atomic-bot r+

PR #427, opened 2016-06-20
[merged] Atomic/mount.py: Break from loop when info found

baude:
In the def _get_cid_from_mountpoint, once the cid is found,
we should break from the for loop.  This makes the function
faster.  And also, for an unknown reason (to me), this ensures
that temporary dirs in /run/atomic/<ts> are cleaned up.


rhatdan:
@rh-atomic-bot r+

PR #430, opened 2016-06-21
[merged] Fix atomic scan NoneType error on bad name

ddarrah:
./atomic: fix typo in Exception handling at end
Atomic/scan.py: fixed code path in _get_scan_list() that resulted in NoneType error if name was not found in either images or containers.


rhatdan:
bot, add author to whitelist

rhatdan:
@baude PTAL

Atomic/scan.py
```diff
@@ -176,8 +176,9 @@ def gen_containers():
                                        if item['Id'] == self.get_input_id(scan_input)), None))
                 docker_object = docker_object if docker_object is not None \
                     else (next((item for item in images if item['Id'] == self.get_input_id(scan_input)), None))
-                docker_object['input'] = scan_input
-                scan_list.append(docker_object)
+                if docker_object is not None:
```
baude: I dont follow the logic here...what happens if docker_object IS none, should we throw an exception?

ddarrah: With the fix it gives an error that the name was not found:

[root@rhel-base atomic]# atomic scan asldkfjasdlfj
Unable to associate 'asldkfjasdlfj' with an image or container
[root@rhel-base atomic]# 

instead of the previous behavior:

[root@rhel-base docker]# atomic scan asldkfjasdlfj
    'NoneType' object does not support item assignmentn[root@rhel-base docker]#

If you follow the code, it basically says:
docker_obj = (search the image list, default None)
docker_obj = (search the container list, default None) if docker_object is not None
there was no clause where it handled the condition where both of those statements resulted in None.

ddarrah: Plus the fix of '\n' for 'n' in the error message.

baude:
@ddarrah Please rebase on master and I will merge it.  Thanks much!

ddarrah:
@baude Thanks. I merged w/ master and the tests passed - do I have to do anything else?

rhatdan:
@rh-atomic-bot r+

PR #435, opened 2016-06-22
Changes made to diff and atomic_dbus.

AmartC:


Atomic/atomic.py
```diff
@@ -134,7 +134,7 @@ class Atomic(object):
                 "--name", "${NAME}",
                 "${IMAGE}"]
 
-    def __init__(self):
+    def __init__(self, is_dbus = False):
         self.d = get_docker_client()
```
rhatdan: No reason to pass this in, flag should be something along the lines of 
self.args.tty=False

Atomic/diff.py
```diff
@@ -43,8 +43,12 @@ def diff(self):
...
             helpers._cleanup(image_list)
 
-            if self.args.json:
+            if self.args.json and not self.is_dbus:
                 util.output_json(helpers.json_out)
+
+            elif self.is_dbus:
+                return str(helpers.json_out)
+                
         except KeyboardInterrupt:
```
rhatdan: Isn't this going to spew lots of output to the screen?

rhatdan:
Also fix the description of the pull request.

Atomic/diff.py
baude: This looks correct to me.  Basically he appears to be saying, if you want JSON and we arent using dbus, dump to stdout.  Thats what it originally did.

Atomic/atomic.py
baude: Not sure I follow you here @rhatdan .  it will be nice to know if things are coming in on dbus in the code.  He could also do:

```
if self.is_dbus:
     self.is_tty = True
```

Atomic/diff.py
```diff
@@ -43,8 +43,12 @@ def diff(self):
             # Clean up
```
baude: Somewhere in diff.py, you will need to make an association between self.is_dbus and setting self.args.json (or make a new var).  In other words, you always want to output JSON if is_dbus.

rhatdan:
You don't need to close the pull requests, you can just fix up your git and push it again with a --force.

rhatdan:
@baude I want diff to always return the json.

If user does not specify json, then he will get nothing. 
If he does specify json, then he can output the return if he so chooses.

I prefer to have 

diff_tty():
      util.output(self.Diff())

Then we can only change atomic command to call diff_tty() rather then diff,

PR #436, opened 2016-06-22
[merged] Atomic/scan.py - Better cleanup with KeyboardInterupt

baude:
When the user issues a control-C to interupt the scan
process, we now handle cleaning up the mount points,
directories, and dm devices better.  There used to be
a small window where control-C would leave behind
mounts and often created Dead containers.


rhatdan:
@rh-atomic-bot r+

PR #437, opened 2016-06-22
[merged] Changes made to Diff.

AmartC:


rhatdan:
LGTM  @baude PTAL

Atomic/diff.py
```diff
@@ -8,6 +8,12 @@
 from . import Atomic
 
 class Diff(Atomic):
+
+    def diff_tty(self):
```
baude: I'm not in love with how this is formulated.  The main use case for diff is to use it from the CLI; which you therefore expect something on stdout.  But in the code, we can diff_tty and then diff.  Can we reverse the logic there and call diff_dbus (instead of tty) and go from there?

AmartC: So would you then recommend that I have a member variable for Atomic called is_dbus which is default set to False?

baude: I guess @rhatdan  has strong feelings it should be this way.  I'll defer to him.

rhatdan:
@rh-atomic-bot r+

PR #438, opened 2016-06-22
Added proper documentation/errors and changed atomic pull to pull from docker by default. 

mtohmaz:
List of changes:
*Updated man pages to list proper registries to pull from, such as docker, or ostree.
*Updated atomic pull --help to list proper registries
*Changed atomic pull to pull from docker by default. 
*Returns an error stating proper format for atomic pull and proper registries when user inputs improper registry for atomic pull


rhatdan:
bot, add author to whitelist

rhatdan:
@mtohmaz Please rebase to latest from upstream.  Please merge your pull requests

Atomic/atomic.py
```diff
@@ -326,12 +326,19 @@ def _pull_dockertar_layers(self, repo, imagebranch, temp_dir, input_layers):
...
-        else: # Assume "oci:"
-            self._check_system_oci_image(repo, image, upgrade)
+        else: # Assume "docker:"
+            image = image.replace("docker:", "")
+            try:
+                if ":" in image:
+                    raise Exception, image
+            except:
+                print "Error: you must specify a proper registry to pull from. Proper registries are docker, dockertar, oci, and ostree."
+                sys.exit(1)
```
rhatdan: If  you raise a ValueError, the atomic CLI will handle this. Never call sys.exit() from a library.

Atomic/util.py
```diff
@@ -43,6 +43,23 @@ def _decompose(compound_name):
         repo, tag = repo.rsplit(':', 1)
     return reg, repo, tag
 
+def atomic_pull_image_by_name(img_name):
+    """
+    Pulls img-name image with atomic. Returns 1 if successful, 0 otherwise.
+    """
+
+    print "Pulling image %s" % img_name
+    try:
```
rhatdan: No debug output please.  Library should not output.

Atomic/util.py
```diff
@@ -43,6 +43,23 @@ def _decompose(compound_name):
...
+        output = subprocess.check_output(["atomic", "pull", img_name])
+    except subprocess.CalledProcessError as grepexc:
```
rhatdan: Why not call the pull command directly versus executing a subprocess?

rhatdan:
You have a lot of files that should not be in your git pull.

rhatdan:
In your git commit line you need a white space between the title and the comments

```
Atomic gives proper errors when specifying non-existant registry.

Also defaults to docker registry.

Atomic gives proper errors when specifying non-existant registry. Also defaults to docker registry.

Committing changes which add error checking for atomic pull and also docker is default.
```

rhatdan:
If you are still here, please stop by my office.

mtohmaz:
Wfh unfortunately, I'm on IRC (mtohmaz)

rhatdan:
Lets work on this tomorrow.

Atomic/util.py
baude: If you really feel the debug output would be beneficial to others, put a if self.debug in front of it.  Either way, then use util.write_out to spill it to stdout.

Atomic/atomic.py
giuseppe: I think this change should not be included.  When "oci:" is specified the image is pulled from a Docker V2 registry through Skopeo. 

"docker:" instead pulls the image from the local Docker (i.e. docker save the image + import the image into OSTree).  This one doesn't seem like the correct default.

What we would like to have though, is the possibility to pull to Docker instead of OSTree.

docs/atomic-pull.1.md
```diff
@@ -11,7 +11,7 @@ atomic-pull - fetch an image locally
 IMAGE
 
 # DESCRIPTION
-**atomic pull**, will fetch a remote image and store it locally.
+**atomic pull**, will fetch a remote image and store it locally. Can pull images from ostree, oci, docker, and dockertar. Atomic pulls from docker by default.
```
giuseppe: This should be changed accordingly to say that by default it pulls from a Docker registry

rhatdan:
@rh-atomic-bot r+

giuseppe:
@rhatdan I had some comments in the code, I think one change must be reverted.  I think by default we want to pull from the registry, not the local Docker instance

giuseppe:
"oci:" means pull from the Docker registry V2 via Skopeo, while "docker:" is the local Docker instance.

The change in the code must be reverted, since assuming by default "oci:" is already what the code is doing

cgwalters:
Offhand my guess here is the PR tester has an old ostree since it's Fedora 23.  I just submitted https://bodhi.fedoraproject.org/updates/ostree-2016.6-1.fc23

Atomic/atomic.py
```diff
@@ -326,12 +326,15 @@ def _pull_dockertar_layers(self, repo, imagebranch, temp_dir, input_layers):
...
+            self._check_system_oci_image(repo, image, upgrade)
         elif self.args.image.startswith("dockertar:"):
             self._pull_docker_tar(repo, image.replace("dockertar:", ""))
-        else: # Assume "oci:"
-            self._check_system_oci_image(repo, image, upgrade)
+        else: # Assume "docker:"
+            image = image.replace("docker:", "")
+            if ":" in image:
+                raise ValueError("Error: improper registry %s. Proper registries are docker, dockertar, oci, and ostree." % image)
+            self._pull_docker_image(repo, image)
```
giuseppe: This chunk must be removed, the code was already doing the pull from the registry by default, this is the behaviour with "oci:".

"docker:" means to do the pull from the local Docker instance, that internally is docker save + import into OSTree.

giuseppe:
the patchset should be rebased on top of origin/master with "git pull --rebase origin master", instead of merging the master branch

jlebon:
> Offhand my guess here is the PR tester has an old ostree since it's Fedora 23. I just submitted https://bodhi.fedoraproject.org/updates/ostree-2016.6-1.fc23

Awesome. Migrating the testers to F24 is on the TODO. But this helps in the short term.

Atomic/atomic.py
rhatdan: We were seeing issues here when trying to pull fedora, it was trying to pull it via ostree, I thought?

@mtohmaz PTAL

rhatdan:
@mtohmaz Could you update this please.

mtohmaz:
Updated the branch

rhatdan:
@giuseppe Are you ok with this patch set now?

rhatdan:
@mtohmaz Needs an update Sorry.

giuseppe:
@rhatdan @mtohmaz I still see the same code in the last version.  The code must assume "oci:" (which is pull from a Docker registry through Skopeo) when nothing is specified.  The last patch changed the code to assume instead "docker:" (which is pull from the local Docker).  I also don't understand the new error condition.  I suggest:

- Revert the changes to `atomic.py`.
- In `docs/atomic-pull.1.md` says that the default is to pull an OCI image from a Docker registry through Skopeo.
- The change in `atomic` must be fixed likewise.
- Adjust `tests/unit/test_pull.py` to follow the existing logic (*oci:* is assumed)

PR #443, opened 2016-06-24
Introduce ATOMIC_TRACEBACK

cgwalters:
I'm not sure if there's a standard for this in the Python community,
but basically if I want to debug something I'd like to be able to
easily get a traceback.


baude:
I currently get tracebacks by using the --debug option to atomic itself.

cgwalters:
That works for me, thanks!  :white_check_mark:

PR #444, opened 2016-06-24
[merged] Atomic/dockerimageid.py: Base structure

baude:
Atomic-1.10+ now prepends its sha algorithm to any Image ID
in returns.  For example, 'sha256:1df32...'.  This can make
it difficult for code that does things like str comparison
between image IDs.  For example, suppose you want to perform
something like if <ID> in [list_of_ids].  You might struggle
to make sure you knew if the ID and list of ids has the
prepending sha algorithm.

This base structure helps handling these cases in two ways:

1. We implement a custom class class DockerImageID that has
   dunder implementations for comparison.  It also has a
   dunder for conversion to a string.  All of this is
   expandable.

2. We propose a class called AtomicDocker intercepts the
   returns from docker.Client.  It runs a iterative
   definition that finds any value that begins with the
   sha algorithm.  It replaces that value with a DockerImageID
   object.

Once merged, we can begin to ensure everything is well behaved
and remove instances of s/sha256://.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -13,6 +13,8 @@ import tarfile
 import stat
 from string import Template
 import calendar
+from Atomic.dockerimageid import DockerImageID
+from Atomic.dockerimageid import iter_subs
 
 try:
     import gi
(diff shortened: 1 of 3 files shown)


runcom:
Make sense, also algorithm isn't changing anytime soon in docker afaict 

LGTM

cgwalters:
It seems like this patch doesn't actually use `AtomicDocker`?  The implementation here is kind of scary in how we're modifying *everything* that starts with `sha256:` - what if that happens to be in a `Description` field or something?

Did you investigate having a simple "canonicalize_image_id" method and using it at call sites?

(BTW, if newer docker is using `sha256:` and older isn't, I'd probably lean towards the canonical form being with it prepended, since it's now canonical.  At least until next week when Docker does something new! :calendar: :gift:  )

runcom:
The pretended algorithm is only used in docker internals and isn't actually used by users (code in docker is adapted to show algorithm-less ID in almost every case)

baude:
@cgwalters you bring up a point i was initially concerned about.  I *think* this should only be seen in ID and ParentID fields.  I could easily refine it to only that.  Would that be preferable to you?

The reason I went away from a simple method is that it would need to be spread all over the code and enforced on new code.  This seemed most straight-forward FWIW.

cgwalters:
I don't have a strong sense of how many call sites that'd be - but it seems like a better plan.  It's hard to predict all the ramifications of a potential change to every string we get from docker.

baude:
@cgwalters I have added some conditionals in now as discussed.  For example,

https://github.com/projectatomic/atomic/pull/444/files#diff-79d970d79c7e0d53bcf146fce1bedaeeR81

Mind peeking?

baude:
@rh-atomic-bot r+

PR #445, opened 2016-06-25
[merged] Work correctly with fully qualified image names

giuseppe:
Atomic.find_remote_image can return None if the image name provided by
the user has not the same name as the one found.  In that case, use the
name provided by the user.

Problem introduced by d29c801ee78597b0ef3f9773467882da8ac9ec71


cgwalters:
This fixes the issue I was seeing, and code looks right.  Thanks!

@rh-atomic-bot r+ 966db2f

PR #446, opened 2016-06-27
[merged] Atomic/scan.py: Fail on Dead containers

baude:
When someone trys to scan a dead container (typically
with --containers or --all), we usually observe DM or
mount issues.  In turn, this usually causes Atomic to
traceback and fail.  Adding in a quick check for those
cases where we nicely fail and cite the Dead container
as a reason.


cgwalters:
Seems reasonable to me.

@rh-atomic-bot r+ 9c6e4fe

PR #452, opened 2016-07-07
Atomic/scan.py: Don't keep open STDIN for scan

ybznek:
`atomic scan fedora:24` runs `docker -it ...` and it needs to have interactive terminal.

Consequences: `atomic scan fedora:24` is not usable in cases like
- `nohup atomic scan fedora:24`
- `ssh root@localhost atomic scan fedora:24`
- ...

It only prints `cannot enable tty mode on non tty input`.

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -113,7 +113,7 @@ class Scan(Atomic):
         # Create the output directory
         os.makedirs(self.results_dir)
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@baude What do you think?  This looks good to me.

baude:
LGTM, I didnt test but assume contributor did.

rhatdan:
@rh-atomic-bot r+

PR #455, opened 2016-07-07
[merged] Run SPC containers with /sys/fs/selinux mounted as read/only

rhatdan:
This tells libselinux inside the container that SELinux is disabled,
even though we are in a SPC container, we don't want the container attempting
to do labeling and especially do not want it loading policy.  If I accidently
load a RHEL6 policy on a RHEL7 machine, bad stuff will happen.


cgwalters:
I think you meant `/sys/fs/selinux` in the commit message.

rhatdan:
Yup, brain and fingers not communicating. Should be fixed now.

cgwalters:
@rh-atomic-bot r+ ddcd55d

PR #457, opened 2016-07-08
[merged] atomic: add dateutil to requirements

yuqi-zhang:
Add python-dateutil, which is used by "atomic ps".


rhatdan:
@rh-atomic-bot r+

PR #458, opened 2016-07-08
[merged] Complete handling of docker-1.10 Images, IDs changes

baude:
In docker-1.10, they began prepending Images IDs with the
checksum algorithm.  For example, sha256:<id>.  This was
problematic in the atomic CLI because we used those IDs
for equality checks (as an example) in the our code. A
previous PR provided a "wrapper" class for the docker-py
calls in which iterate through the return values from
docker-py calls.  We now perform the actual removal of
the ALGO: bits from the relevant strings.  All returns from
docker-py calls no longer contain the prepended text.


rhatdan:
@baude your build is failing python3-pylint.

baude:
@rhatdan ok, did the pylint disable workaround as we needed to check for a unicode type in that code.

rhatdan:
Nice cleanup, although I think we want to eventually move AtomicDocker into docker.py or something like that rather then having it in atomic.py.

rhatdan:
@rh-atomic-bot r+

Issue #450, opened 2016-07-05
Change polkit policy to `<allow_any>auth_admin</>`


cgwalters:
Right now we authorize things like `StorageReset` under `org.atomic.read` which is wrong.

Second, for some reason at least with the Cockpit session we had to change `<allow_any>yes</>` when (as far as I understand) the Cockpit session should be under `<allow_active>`).

mvollmer:
As far as I understand, `allow_active` and `allow_inactive` only apply to local sessions, like the console and gdm, but not remote sessions like ssh and Cockpit.  So the problem is that the Cockpit sessions are remote, not that they are inactive.

`<allow_any>auth_admin</allow_any>` might be what you want.

cgwalters:
Has Cockpit been changing other services similarly?  Looks like it's likely, `org.freedesktop.NetworkManager.network-control` has `<allow_any>auth_admin</allow_any>` too.

rhatdan:
I agree these settings are bad.  Since the primary use case for these bindings is intended to be cockpit, we need to satisfy its use case.

rhatdan:
FIxed in master

PR #459, opened 2016-07-08
[merged] Change atomic dbus permissions

rhatdan:
When actually modifying the system we need to require org.atomic.readwrite


rhatdan:
This should fix https://github.com/projectatomic/atomic/issues/450

rhatdan:
@mvollmer @cgwalters PTAL

cgwalters:
@rh-atomic-bot r+ 662077d

PR #460, opened 2016-07-08
Cockpit client which works for version, diff and scan_list

AmartC:
Updates to DBus API


Atomic/atomic.py
```diff
@@ -149,6 +149,7 @@ def __init__(self):
         self.docker_cmd = None
         self.debug = False
         self.is_python2 = (int(sys.version[0])) < 3
+        self.notTty = False
```
rhatdan: Change to self.TTY = True and then remove double negatives in code.

Atomic/scan.py
```diff
@@ -125,7 +124,8 @@ def set_scanner():
         scan_cmd = self.sub_env_strings(" ".join(scan_cmd))
 
         # Show the command being run
-        util.write_out(scan_cmd)
+        if not self.notTty:
+            util.write_out(scan_cmd)
```
rhatdan: if self.useTTY:

Atomic/scan.py
```diff
@@ -142,13 +142,16 @@ def set_scanner():
             util.check_call(scan_cmd, stdout=stdout, env=self.cmd_env())
 
             # output results
-            self.output_results()
+            if not self.notTty:
+                self.output_results()
```
rhatdan: If self.useTTY:

Atomic/scan.py
```diff
@@ -142,13 +142,16 @@ def set_scanner():
...
 
             # record environment
             self.record_environment()
         finally:
             # unmount all the rootfs
             self._unmount_rootfs_in_dir()
+            if self.notTty:
+                return (json.dumps(self.get_scan_data()))
```
rhatdan: if not self.useTTY:

atomicClient.html
```diff
@@ -0,0 +1,249 @@
+<head>
```
rhatdan: This code should be put into tests directories, since we are not intending on shipping this.

atomic_client.py
```diff
@@ -14,38 +15,54 @@ def __init__(self):
...
 
     @polkit.enable_proxy
     def scan_list(self):
-        ret = self.dbus_object.scan_list(dbus_interface="org.atomic")
+        ret = self.dbus_object.ScanList(dbus_interface="org.atomic")
+        return ret
+
+    @polkit.enable_proxy
+    def asyncscan(self, scan_targets, scanner, scan_type, rootfs, _all, images, containers):
+        token = self.dbus_object.ScheduleScan(scan_targets, scanner, scan_type, rootfs, _all, images, containers, dbus_interface="org.atomic", timeout = 2147400)
```
rhatdan: Shouldn't this be async_scan

atomic_client.py
```diff
@@ -14,38 +15,54 @@ def __init__(self):
...
+            ret = self.dbus_object.GetScanResults(token, dbus_interface="org.atomic")
+            if ret:
+                break
+            time.sleep(1)
+
+        return ret
+
+    @polkit.enable_proxy
+    def syncscan(self, scan_targets, scanner, scan_type, rootfs, _all, images, containers):
+        ret = self.dbus_object.SyncScan(scan_targets, scanner, scan_type, rootfs, _all, images, containers, dbus_interface="org.atomic", timeout = 2147400)
```
rhatdan: This should be just scan, not sysncscan, it is implied.

atomic_client.py
```diff
@@ -109,8 +126,20 @@ def print_scan_list(all_scanners):
             print(resp)
 
         elif(sys.argv[1] == "scan"):
-            if(sys.argv[2] == "list"):
+            if("--list" in sys.argv):
                 all_scanners = json.loads(dbus_proxy.scan_list())
                 print_scan_list(all_scanners)
+
```
rhatdan: This should be just a switch statement.  We don't want to allow users to specify multiple commands in sys.argv.

PR #461, opened 2016-07-11
[merged] Remove unused and bogus ;

rhatdan:
Beginning to cleanup code to remove warnings from pylint

Testing with

	$(PYTHON3_PYLINT) --disable=all --enable=E --enable=W --additional-builtins=_ *.py atomic Atomic tests/unit/*.py

Reveals these unused and bogus ;.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -195,7 +195,7 @@ def urllib3_disable_warnings():
             if not 'urllib3' in sys.modules:
                 import urllib3
                 have_urllib3 = True
-        except ImportError as e:
+        except ImportError:
             pass
         if have_urllib3:
             # Except only call disable-warnings if it exists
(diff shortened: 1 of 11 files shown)


cgwalters:
There's a bit more here than removal of `;`, but sure, looks sane to me.

@rh-atomic-bot r+ 048988b

rhatdan:
@cgwalters I will have some more coming.

Atomic/util.py
```diff
@@ -241,9 +241,9 @@ def skopeo_layers(image, args=[], layers=[]):
             raise ValueError("Unable to interact with this registry: {}".format(r.stderr))
     except OSError:
         raise ValueError("skopeo must be installed to perform remote inspections")
-    except Exception as e:
+    finally:
```
giuseppe: this changes the logic of the function and rm the directory even on success.  We should rm the temporary directory only on an exception.  I'll post another patch to address it.

giuseppe: PR here: https://github.com/projectatomic/atomic/pull/464

PR #464, opened 2016-07-12
[merged] util: skopeo_layers doesn't rm the layers directory on success

giuseppe:
partly revert a change in f2974f5bff5b3ea58edfb0e85fa0b4f03e35fb61

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -232,6 +232,7 @@ def skopeo_layers(image, args=[], layers=[]):
     :param layers: if set, specify what layers must be downloaded
     :return: Returns the temporary directory with the layers
     """
+    success = False
     temp_dir = tempfile.mkdtemp()
     try:
         args = ['skopeo', 'layers'] + args + [image] + layers
@@ -239,10 +240,12 @@ def skopeo_layers(image, args=[], layers=[]):
         if r.return_code != 0:
             check_v1_registry(image)
             raise ValueError("Unable to interact with this registry: {}".format(r.stderr))
+        success = True
     except OSError:
         raise ValueError("skopeo must be installed to perform remote inspections")
     finally:
-        shutil.rmtree(temp_dir)
+        if not success:
+            shutil.rmtree(temp_dir)
 
     return temp_dir
 


Atomic/util.py
```diff
@@ -235,14 +235,16 @@ def skopeo_layers(image, args=[], layers=[]):
...
-    finally:
+        try:
+            r = subp(args, cwd=temp_dir)
+            if r.return_code != 0:
+                check_v1_registry(image)
+                raise ValueError("Unable to interact with this registry: {}".format(r.stderr))
+        except OSError:
+            raise ValueError("skopeo must be installed to perform remote inspections")
+    except Exception as e:
         shutil.rmtree(temp_dir)
```
rhatdan: Can  you be more specific on the Exceptions (ValureError)?  pylint does not like this.    Do you want to remove the layers on success?  This looks like it was before, is that the intended change?

giuseppe:
I want to keep the layers on success, so that the caller of the function can access them.  The "except Exception" will catch all the exceptions but won't catch the successful case, while "finally" would do that as well. 

The difference now compared to f2974f5~1 is that the temporary directory will be deleted also on OSError (previously it wasn't and it was a bug).

I rewrote it :arrow_up: to not use the "except Exception" part.

Atomic/util.py
```diff
@@ -232,19 +232,22 @@ def skopeo_layers(image, args=[], layers=[]):
     :param layers: if set, specify what layers must be downloaded
     :return: Returns the temporary directory with the layers
     """
+    ret = None
```
rhatdan: Call this return_dir

Atomic/util.py
```diff
@@ -232,19 +232,22 @@ def skopeo_layers(image, args=[], layers=[]):
...
     temp_dir = tempfile.mkdtemp()
     try:
         args = ['skopeo', 'layers'] + args + [image] + layers
         r = subp(args, cwd=temp_dir)
         if r.return_code != 0:
             check_v1_registry(image)
             raise ValueError("Unable to interact with this registry: {}".format(r.stderr))
+        ret, temp_dir = temp_dir, None
     except OSError:
```
rhatdan: Split this into two lines and add a comment.
# Successfully created return_dir, clear temp_dir field so it will not get removed.

Atomic/util.py
rhatdan: Change this to failed=False

Atomic/util.py
```diff
@@ -232,19 +232,22 @@ def skopeo_layers(image, args=[], layers=[]):
...
     """
+    ret = None
     temp_dir = tempfile.mkdtemp()
     try:
         args = ['skopeo', 'layers'] + args + [image] + layers
         r = subp(args, cwd=temp_dir)
         if r.return_code != 0:
             check_v1_registry(image)
             raise ValueError("Unable to interact with this registry: {}".format(r.stderr))
+        ret, temp_dir = temp_dir, None
```
rhatdan: success=True

Atomic/util.py
```diff
@@ -232,19 +232,22 @@ def skopeo_layers(image, args=[], layers=[]):
...
     except OSError:
         raise ValueError("skopeo must be installed to perform remote inspections")
     finally:
-        shutil.rmtree(temp_dir)
+        if temp_dir:
```
rhatdan: if not success:

Atomic/util.py
```diff
@@ -232,19 +232,22 @@ def skopeo_layers(image, args=[], layers=[]):
...
+            shutil.rmtree(temp_dir)
 
-    return temp_dir
+    return ret
```
rhatdan: return temp_dir

giuseppe:
fixed here :arrow_up:

rhatdan:
@rh-atomic-bot r+

PR #473, opened 2016-07-13
[merged] status: support JSON output

### LINKED ISSUE #469: support '--json' option for 'atomic host status'
The latest version of `rpm-ostree status` now supports JSON output, as referenced in https://github.com/projectatomic/rpm-ostree/pull/315.

The `atomic host status` command should also support this option.

miabbott:
Add support for `rpm-ostree status --json`

Fixes #469

Signed-off-by: Micah Abbott <miabbott@redhat.com>

### DIFF
--- a/atomic
+++ b/atomic
@@ -181,6 +181,9 @@ def create_parser(atomic):
         # atomic host status
         statusp = host_subparser.add_parser(
             "status", help=_("list information about all deployments"))
+        statusp.add_argument("-j", "--json", dest="json",
+                             action="store_true",
+                             help=_("Display status in JSON format"))
         statusp.add_argument("-p", "--pretty", dest="pretty",
                               action="store_true",
                               help=_("Display status in formatted rows"))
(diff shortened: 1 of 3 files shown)


atomic
```diff
@@ -181,6 +181,9 @@ def create_parser(atomic):
         # atomic host status
         statusp = host_subparser.add_parser(
             "status", help=_("list information about all deployments"))
+        statusp.add_argument("-j", "--json", dest="json",
+                             action="store_true",
+                             help=_("Display status in JSON format"))
         statusp.add_argument("-p", "--pretty", dest="pretty",
                               action="store_true",
                               help=_("Display status in formatted rows"))
```
jlebon: Actually, this option is deprecated in `rpm-ostree` now. We should probably deprecate it here as well. Maybe change the help to the same one as upstream -- "This option is deprecated and no longer has any effect".

miabbott:
@jlebon I was going to let that be, but pushed a fixup for you.  :arrow_up:

jlebon:
Heh, thanks.
@rh-atomic-bot r+ e9f1727

PR #474, opened 2016-07-13
Making scan work for Dbus API

AmartC:


tests/atomicClient.html
```diff
@@ -187,6 +187,10 @@
                 call = proxy.Scan(scan_targets, scanner, scan_type, rootfs, false, false, true);
               }
 
+              else {
+                call = proxy.Scan([command[1]], scanner, scan_type, rootfs, false, false, false);
+              }
+
```
rhatdan: Is this code correct?

Atomic/scan.py
```diff
@@ -414,15 +414,15 @@ def print_scan_list(self):
...
         m.mountpoint = mountpoint
         m.image = image
         m.shared = True
         m.mount()
 
     def unmount(self, mountpoint):
         m = mount.Mount()
-        m.set_args(self.args)
+        m.args = self.args
         m.mountpoint = mountpoint
```
rhatdan: Please explain these?

AmartC: Instead of set_args which was causing problems based on type difference within mount's function since that checks for certain terms within the arguments, in the case of scan, the arguments simply only need to be initialized but they are not used.

tests/atomicClient.html
AmartC: Yes, I tested it with Cockpit.

rhatdan:
@rh-atomic-bot r+

rhatdan:
@jlebon we need python3-pylint also.

PR #475, opened 2016-07-14
Migrate script should now work since we are after docker-1.10

rhatdan:


rhatdan:
@shishir-a412ed PTAL

rhatdan:
Looks like the migrate script needs some work.

shishir-a412ed:
@rhatdan Yes this should work now. Let me test it with docker 1.10 and confirm.

Shishir

rhatdan:
I want it tested with docker-1.12 also.  The test is blowing up now.  See Details

jlebon:
The tester was still at 1.9, just updated it to 1.10.
Long-term I'd like to try migrating the testers to AH and let them auto-update.

jlebon:
bot, retest this please

jlebon:
The error is now:

```
Error response from daemon: empty export - not implemented
```

shishir-a412ed:
@rhatdan I have opened a follow up PR #498  for this. Please check that.
We can close this one, and merge that one instead.

PR #476, opened 2016-07-14
[merged] Retain persistent data from vulnerability scans

baude:
We now want to retain a summary of each results when
performing a vulnerability scan.  This persistent data
will be used for marking images or containers that are
vulnerable in both atomic and cockpit.  This marking of
the image has been implemented in atomic images.

There is a core function get_vulnerable_ids() which
can returns a list of uuids that are vulnerable.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -99,6 +99,9 @@ class Atomic(object):
                 "--name", "${NAME}",
                 "${IMAGE}"]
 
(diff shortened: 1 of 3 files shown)


rhatdan:
Will this list grow forever?  Should we purge IDS from the list? Not sure we can.

rhatdan:
@rh-atomic-bot r+
Merging for now, and we can make changes later.

PR #478, opened 2016-07-15
[merged] Show containers type

giuseppe:
add information about the type of the containers and the images in the "atomic ps" and "atomic images" output

### DIFF
--- a/Atomic/ps.py
+++ b/Atomic/ps.py
@@ -22,7 +22,7 @@ class Ps(Atomic):
             created = dateparse(ret['created']).strftime("%F %H:%M") # pylint: disable=no-member
             all_containers.append({"type" : "systemcontainer", "container" : container,
                                    "image" : image, "command" : command, "created" : created,
-                                   "status" : status})
+                                   "status" : status, "runtime" : "runc"})
 
         # Collect the docker containers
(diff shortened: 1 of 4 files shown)


Atomic/ps.py
```diff
@@ -22,7 +22,7 @@ def ps(self):
             created = dateparse(ret['created']).strftime("%F %H:%M") # pylint: disable=no-member
             all_containers.append({"type" : "systemcontainer", "container" : container,
                                    "image" : image, "command" : command, "created" : created,
-                                   "status" : status})
+                                   "status" : status, "kind" : "runc"})
```
rhatdan: I am not crazy about kind.

machinectl calls this SERVICE

```
MACHINE                          CLASS     SERVICE
ec7dacdbe7859e15c3e4fb2d73226f15 container docker 
```

rhatdan: Perhaps "RUNTIME"?

rhatdan:
I like this patch, but we have to come to some consistency on the naming.  kind, imageKind, type ...

giuseppe:
@rhatdan thanks for the review.  I've addressed the comments in the version here :arrow_up:

rhatdan:
@rh-atomic-bot r+

PR #483, opened 2016-07-15
Atomic now has --user flag.

mtohmaz:
Atomic now has --user flag which creates a BARE_USER repo locally to allow non-privileged user to pull images.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -110,6 +110,7 @@ class Atomic(object):
         self.setvalues = None
         self.inspect = None
         self.backend = None
+        self.user = None
         self.force = False
         self._images = []
         self.containers = False
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -51,6 +51,7 @@ class SystemContainers(object):
     def __init__(self):
         self.atomic_config = util.get_atomic_config()
         self.backend = None
+        self.user = None
         self.args = None
         self.setvalues = None
 
@@ -60,6 +61,10 @@ class SystemContainers(object):
     def set_args(self, args):
         self.args = args
 
+        try:
+            self.user = args.user
+        except (NameError, AttributeError):
+            self.user = None
         try:
             self.backend = args.backend
         except (NameError, AttributeError):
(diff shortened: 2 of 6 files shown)


rhatdan:
I need man page changes, test suite, bash completions...

giuseppe:
I'd move the logic of the new function `_get_local_ostree_repo` into the existing `_get_ostree_repo` as the two differences (path and use `OSTree.RepoMode.BARE_USER`) can be easily managed there.  `_get_ostree_repo` is already used in other places, so I think it is better to check if it is running as user only in one place (`_get_ostree_repo`).

More important: how is an user image going to look like?  The current system images use runc and systemd, while, I assume, we will use bubblewrap for user images.  We will need to store some extra information on how the user image is installed and executed.  Where this information will be?  How installing and executing an user image will look like?

rhatdan:
Yes that is what we need to begin designing.  We need some kind of local configuration to define how to run a bubblewrap container, a possible solution would be to drop a OCI specification and then have a tool to translate OCI Specs to bubblewrap CLI.

giuseppe:
would it reasonable to add OCI specs support to bubblewrap so that the same configuration (and to some extend the same process to install an image) can be used?

rhatdan:
There is real reluctance to add any kind of object processing to a SETUID app.  So it would be preferred that this happens in non priv mode before executing the SETUID app.

rhatdan:
@mtohmaz Man pages? Command Completion?  Tests?

@cgwalters PTAL

mtohmaz:
I'm still working on the tests, but I'll push everything once I finish them this morning.

giuseppe:
I've spent some time during the weekend to write a wrapper around bubblewrap that reads the configuration from an OCI conf file and generates the command line for bubblewrap:

https://github.com/giuseppe/bubblewrap/tree/bwrap-oci

Still it is far from being complete, it lacks many options and it is quite hacky/buggy, but it can be used as a starting point:

```
$ sudo dnf install--releasever 24 --installroot=rootfs emacs
$ ocitools generate --tty --args /usr/bin/emacs > config.json
$ bwrap-oci
# ...emacs running..
```

jlebon:
bot, retest this please

rhatdan:
@mtohmaz Needs a rebase.

rhatdan:
@mtohmaz PTAL

jlebon:
bot, retest this please

PR #490, opened 2016-07-19
[merged] Makefile: make pylint executables more generic

jlebon:
Use "$PYTHON -m pylint" rather than harcoding a pylint path, which is
more generic and will work regardless of whether pylint was installed by
yum/dnf or pip.

Otherwise you can get nasty issues like pip and RPM stepping on each
other's toes trying to own /usr/bin/pylint. Also,
/usr/bin/python3-pylint is only installed by the RPM package, not by the
pip package.


cgwalters:
@rh-atomic-bot r+ 09bb56c

Issue #470, opened 2016-07-13
Add more functionality to "atomic ps"


yuqi-zhang:
Currently the "atomic ps" command only accepts 3 extra arguments (not counting help). Some potentially useful sub-commands to enhance functionality could be:

  --quiet - display only the IDs
  --no-trunc - don't truncate output
  --sys - only display system containers, or better,
  --filter - filter content
  or some variation of --format="format-string"

Originally, this was intended to help fix bash auto-complete, which did not display atomic containers and images. But a workaround fix has been provided here: https://github.com/projectatomic/atomic/pull/467 . Regardless, extra options to increase/manipulate information would be helpful.

If any of these seem useful, or should be named differently, please point it out.

rhatdan:
These sound good to me.

giuseppe:
I like the idea of `--filter="all|system|docker"` instead of `--sys`

runcom:
+1 to a filter flag instead of so many other flags (ala docker)

yuqi-zhang:
Ok, thanks for the comments!

rhatdan:
If we add these filters flags, please do it for other atomic calls, so we are consistent.

atomic images --filter ...

yuqi-zhang:
images currently only has --no-heading and --prune at the moment, should probably be expanded as well. For example, docker images supports: all, digests, filter, format, no-trunc, quiet

rhatdan:
Right we need them to match.

yuqi-zhang:
Working on adding similar functionality to images as well, will close issue once that is done.

PR #493, opened 2016-07-20
[merged] atomic: add sub-commands to atomic ps

yuqi-zhang:
Add the following sub-commands to ps:

  --filter: filter output based on given VAR=VALUE

  --no-trunc: do not truncate output

  --quiet: only display container IDs

and corresponding bash auto-complete, tests, and documentation.


yuqi-zhang:
In reference to https://github.com/projectatomic/atomic/issues/470

Atomic/ps.py
```diff
@@ -31,15 +36,39 @@ def ps(self):
...
         if self.args.json:
             util.write_out(json.dumps(all_containers))
             return
 
-        col_out = u"{0:12} {1:20} {2:20} {3:16} {4:9} {5:10}"
+        if self.args.truncate:
+            max_len_container = 12
+            max_len_image = 20
+            max_len_command = 20
+
```
giuseppe: please drop this empty line before the `else:`

tests/integration/test_system_containers.sh
```diff
@@ -59,6 +59,10 @@ ${ATOMIC} ps | grep -q "test-system"
 ${ATOMIC} ps --json | grep -q "test-system"
 ${ATOMIC} ps --all | grep -q "test-system"
 ${ATOMIC} ps --json --all | grep -q "test-system"
+${ATOMIC} ps --filter id=test-system | grep -q "test-system"
+${ATOMIC} ps --no-trunc | grep -q "test-system"
+${ATOMIC} ps --quiet | grep -q "test-system"
+${ATOMIC} ps -aq --no-trunc --filter=id=test-system | grep -q "test-system"
```
giuseppe: can we add a test for no matches?

`+${ATOMIC} ps -aq --no-trunc --filter=id=no-test-system | grep -v -q "test-system"`

giuseppe:
Great work!  Just some minor things from my side

rhatdan:
Not sure of --filter?  Why not use grep for this?  Do we have similar options for atomic top?

@baude PTAL

yuqi-zhang:
A use case where this is better than grep would be say, if you had many IDs with "abc" and images with "abc", but you just wanted to view the containers with IDs including "abc", grep would not be able to differentiate.

Currently we do not support filter in any of our commands, but I will probably port it over to "atomic images" as well if you're ok with it.

docs/atomic-ps.1.md
```diff
@@ -24,11 +27,21 @@ Using --all will list all the installed containers.
 **-a** **--all**
   Print all the installed containers
 
-**-n** **--noheading**
-  Do not print heading when listing the containers
+**-f** **--filter**
+  Filter output based on given filters in the format of VARIABLE=VALUE
```
rhatdan: Give me more data here on what is filterable.

+        filterables = ["container", "image", "command", "created", "status", "runtime"]

bash/atomic
```diff
@@ -311,9 +311,19 @@ _atomic_pull() {
 }
 
 _atomic_ps() {
+	local all_options="
+		--all -a
+		--filter -f
+		--help
```
rhatdan: Not sure how easy this would be to do, but have command completion on filterables.
+        filterables = ["container", "image", "command", "created", "status", "runtime"]

tests/integration/test_system_containers.sh
```diff
@@ -63,6 +63,9 @@ ${ATOMIC} ps --filter id=test-system | grep -q "test-system"
 ${ATOMIC} ps --no-trunc | grep -q "test-system"
 ${ATOMIC} ps --quiet | grep -q "test-system"
 ${ATOMIC} ps -aq --no-trunc --filter=id=test-system | grep -q "test-system"
+if ${ATOMIC} ps -aq --no-trunc --filter=id=non-existing-system | grep -q "test-system"; then
+    exit 1
```
rhatdan:
```
--filter=id=non-existing-system
```
Ugly
```
--filter id=non-existing-system
```
Should work.

docs/atomic-ps.1.md
rhatdan: Give me an example in the man page.

yuqi-zhang:
Fixed up above issues with documentation, and added auto-complete for filterable options

rhatdan:
@rh-atomic-bot r+

PR #497, opened 2016-07-22
[merged] Various syscontainers fixes

giuseppe:


Atomic/syscontainers.py
```diff
@@ -57,6 +58,33 @@ def __init__(self):
     def get_atomic_config_item(self, config_item):
         return util.get_atomic_config_item(config_item, atomic_config=self.atomic_config)
 
+    # Try different strategies to do a syncfs:
+    # 1) Load it from libc.
+    # 2) Run "sync --file-system", supported by coreutils sync.
+    # 3) Use os.sync, available from Python 3.3.
+    # 4) Run "sync".
```
cgwalters: Holy sync batman!  Do we really need any strategy other than the libc one?  Why wouldn't ctypes work everywhere?

BTW I think we really want `syncfs` and not `sync` because the latter doesn't actually report errors.

giuseppe: I am using a fallback to `sync`, since `sync --file-system FILE` is a GNU extension that probably is not present on other implementations.

cgwalters: Hm...ostree is pretty aggressive about using linux extensions although I could imagine some BSD port for the core filesystem bits.  But we hard require linux for runc/bubblewrap, so I don't see an issue with hard requiring `syncfs` to be available in libc either, right?

cgwalters:
syncfs aside I scanned this and the high level looks sane to me.

giuseppe:
I have pushed a new version where I use only `syncfs`.

I have not used a fixup patch as I wanted to try out signed git commits and that wouldn't work with it

cgwalters:
Unfortunately, because Homu rewrites/rebases it conflicts with signed commits.

giuseppe:
@rhatdan are you fine with these changes?  They fix a few outstanding issues

Atomic/syscontainers.py
```diff
@@ -306,6 +346,19 @@ def get_system_containers(self):
             ret.append(container)
         return ret
 
+    def delete_image(self, image):
+        repo = self._get_ostree_repo()
```
rhatdan: I would prefer that this not return true/false, since it should really throw an exception on failure.  And caller should not call if it is not an ostree image.

giuseppe: are you fine if I add an `allow_no_image` parameter?  I would like that the `delete_image` does not need a time of check to time of use race condition, that can happen if another process deletes the same image between the time the caller check its existence and deletes it.

rhatdan: I am fine with catching the IMAGE DOES NOT EXIST and returning success.  If the intent is to remove an object and the object does not exist, I consider that a success.

giuseppe: ok thanks, I have pushed a new version which implements this (and removed the `allow_no_image` parameter).  We don't report the IMAGE DOES NOT EXIST anymore and attempt to delete the image from OSTree after checking there is one.

giuseppe:
rebased :arrow_up:

giuseppe:
and rebased once again :arrow_up:  :)

@rhatdan is there anything more that should be changed?

rhatdan:
@rh-atomic-bot r+

jlebon:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rhatdan Homu isn't really prepared to mix manual and its merges...can we try to collaboratively maintain the test suite and rather than doing "force merges" instead e.g. tweak the tester to avoid issues temporarily?

cgwalters:
@giuseppe I think you might be able to work around this with rebasing again

jlebon:
@giuseppe Could you rebase and disable the `test_migrate.sh` test for now. I also made sure `runc` was installed (apologies, thought I had done that already).

yuqi-zhang:
To summarize, the original PR here: https://github.com/projectatomic/atomic/pull/498 broke test_migrate locally (reproducible) but passed the upstream testers originally. For this PR, I can reproduce the same errors from upstream tester locally in test_migrate, which look something like this:

```
+ /usr/bin/coverage run --source=./Atomic/ --branch --append atomic mount dc0270793cbce953962e518cd673022d84f2f89ea608886d8435a67e1f46b963 --live /home/jenkins/workspace/atomic-upstream-test/atomic/.tmp.k2F34XDW9P/mnt_work/container
Cannot live mount non-running container.
+ /usr/bin/coverage run --source=./Atomic/ --branch --append atomic unmount /home/jenkins/workspace/atomic-upstream-test/atomic/.tmp.k2F34XDW9P/mnt_work/container
No device mounted at /home/jenkins/workspace/atomic-upstream-test/atomic/.tmp.k2F34XDW9P/mnt_work/container
+ /usr/bin/coverage run --source=./Atomic/ --branch --append atomic mount dc0270793cbce953962e518cd673022d84f2f89ea608886d8435a67e1f46b963 --shared /home/jenkins/workspace/atomic-upstream-test/atomic/.tmp.k2F34XDW9P/mnt_work/container
+ set -e
+ trap cleanup_mount EXIT
```

giuseppe:
@jlebon @cgwalters  rebased again :arrow_up:

jlebon:
@rh-atomic-bot r+ 0165fa5

PR #502, opened 2016-07-28
[merged] images: add more sub-commands to images

yuqi-zhang:
Add the following sub-commands to "atomic images list"

  --all: show all images, including intermediate images

  --filter: filter output based on given filters

  --quiet: only display image IDs

and corresponding bash auto-complete, tests, and documentation.

This is to match similar commands in "atomic ps", as highlighted here: https://github.com/projectatomic/atomic/issues/470


giuseppe:
the code looks good to me

rhatdan:
@rh-atomic-bot r+

PR #503, opened 2016-07-29
[merged] Fix python3 stuff

giuseppe:
fix a bunch of portability issues to Python3.  Now the tests suite is successful with Python3 as well.

Also fix leaked fd's.


cgwalters:
In some cases this looks like it's undoing unicode bits...but since the test suite passes hopefully they weren't necessary!

@rh-atomic-bot r+ c39dcc7

PR #505, opened 2016-07-29
[merged] Run atomic without the docker daemon running

rhatdan:


rhatdan:
@giuseppe @cgwalters PTAL now with @property

Atomic/atomic.py
```diff
@@ -810,7 +810,10 @@ def install(self):
 
     def _container_exists(self, name):
         try:
-            return self.syscontainers.get_system_container_checkout(name) or self._inspect_container(name)
+            if self.system:
+                return self.syscontainers.get_system_container_checkout(name)
+            else:
+                self._inspect_container(name)
```
giuseppe: it should be "return self._inspect_container(name)"

giuseppe:
beside the missing return statement, it looks fine for me

rhatdan:
@giuseppe What is the equivalent of instpect_image and inspect_container for system containers?

rhatdan:
@giuseppe Trying to get atomic verify to work on a system container.  Do we store the image json data anywhere?

rhatdan:
These are the ones that work with docker daemon stopped on both python2 and python3 with a fedora system image.

. atomic ps
. atomic mount fedora /mnt
. atomic umount /mnt
. atomic images
. atomic host

Broken, needs systemcontainer fixes to enable, Biggest thing is getting inspect_image and inspect_container to work.

atomic diff fedora
atomic help fedora
atomic info fedora
atomic install fedora
atomic mount fedora
atomic push fedora
atomic scan fedora
atomic stop fedora
atomic run fedora
atomic storage fedora
atomic top fedora
atomic uninstall fedora
atomic unmount fedora
atomic update fedora
atomic version fedora
atomic verify fedora

cgwalters:
Hum...can't we just do a first pass of this which skips it for `atomic host`?

rhatdan:
I actually have no problem merging what I have at this point.  And we can continue to fix other problems when docker daemon is not started.

Atomic/client.py
```diff
@@ -41,20 +48,26 @@ def __getattr__(self, name):
...
         obj = self._dockerclient
         attr = docker.AutoVersionClient.__getattribute__(obj, name)
         if hasattr(attr, '__call__'):
             def newfunc(*args, **kwargs):
-                result = attr(*args, **kwargs)
-                return iter_subs(result)
+                try:
+                    result = attr(*args, **kwargs)
+                    return iter_subs(result)
+                except requests.exceptions.ConnectionError as e:
```
cgwalters: I'm confused why this is necessary if we have the `@property` changes.  Shouldn't we not do the connection until `.dockerclient` is referenced, only reference it in paths that need it?

rhatdan: If the docker daemon is down, then when the containers or image initially tries to connect to the service, it will throw an exception requests.exceptions.ConnectionError, we want to return and empty list if the user was looking for containers or images.  This way atomic images list can work with system containers and not just throw an exception saying docker is not running.

rhatdan: The goal of these changes is to do 

atomic images list
atomic ps --all

Both of these look for docker images as system containers, and potentially others in the future.  We don't want to catch all calls to self.dockerclient.images() and self.dockerclient.containers(), this little change allows us to not messy up the code when the docker service is not running.

cgwalters:
OK.  I somehow feel this could be cleaner but we can revisit it later if need be.

@rh-atomic-bot r+  	2766a68

PR #506, opened 2016-08-01
[merged] System containers: basic support for verify

giuseppe:


Atomic/syscontainers.py
```diff
@@ -369,6 +368,23 @@ def delete_image(self, image):
...
+        return self._inspect_system_branch(repo, imagebranch)
+
+    def _inspect_system_branch(self, repo, imagebranch):
+        commit_rev = repo.resolve_rev(imagebranch, False)[1]
+        commit = repo.load_commit(commit_rev)[1]
+
+        tag = ":".join(imagebranch.replace("ociimage/", "").rsplit('-', 1))
+        timestamp = OSTree.commit_get_timestamp(commit)
+
+        return {'Id' : commit_rev, 'RepoTags' : [tag], 'Names' : [], 'Created': timestamp,
```
cgwalters: Shouldn't this ID actually be the original docker image digest?  And we insert a new `OstreeChecksum` field or something?

cgwalters:
No tests had to change for this?  Is there no coverage?

Atomic/syscontainers.py
giuseppe: the problem is that the commit having the manifest metadata cannot be mapped to a Docker layer, so I preferred to use the OSTree commit for both layers and the ID, so for now I am using the same ID used for `images list`.

With @yuqi-zhang we are currently discussing about the ID to use in the images list output as well (which currently is the `commit_rev` too), as we are thinking about changing there as well, so that it uses the Docker layer ID, but we didn't yet figure out what ID should be used for the metadata commit, which gives the name to the image.

giuseppe:
there is only a basic  test for `verify` which is not really exaustive.  More tests must be added for the system containers part as well.

giuseppe:
I added a fixup patch that adds a test for `verify` system images.

I made the change for the image ID we use in a separate patch, so now we use:

1) The Digest from the manifest if the commit has the Docker manifest in the metadata
2) The $SHA part from ociimage/$SHA if it is a checksum
3) The OSTree commit checksum

This should enable us now to fully support `images --all` for system containers @yuqi-zhang

rhatdan:
LGTM

rhatdan:
@cgwalters PTAL

giuseppe:
rebased :arrow_up:

rhatdan:
@rh-atomic-bot r+

PR #509, opened 2016-08-02
[merged] Atomic mount will mount\unmount a user image.

mtohmaz:


baude:
Can you please write a proper commit message to provide context on this PR?  Usually this means the first line is a very short description of your PR (what it does) and then you provide a lengthier, more description summary thereafter.

Atomic/atomic.py
```diff
@@ -296,6 +296,11 @@ def set_args(self, args):
         except (NameError, AttributeError):
             pass
 
+        try:
+            self.user = args.user
```
baude: Why not use the hasattr approach like you did in mount.py?

mtohmaz:
Thanks, I've updated the commit message to reflect the changes.

rhatdan:
@mtohmaz Use hassattr like @baude suggests, and your tests are failing.

Atomic/atomic.py
```diff
@@ -296,6 +296,11 @@ def set_args(self, args):
         except (NameError, AttributeError):
             pass
 
+        if hasattr(args, "user"):
+            self.user = args.user
+        else:
+            self.user = False
+
```
rhatdan: Should just default to self.user = None  Not False.  It is not a boolean.

atomic
```diff
@@ -687,6 +685,8 @@ if __name__ == '__main__':
     try:
         with Atomic.Atomic() as atomic:
             aparser = create_parser(atomic.help())
+            if os.geteuid() != 0:
+                aparser.add_argument(dest="user", action="store_true", default=True)
```
cgwalters: It seems a bit odd to create an argument we don't expect people to set - I'd set `self.user` inside `__init__` or so.

rhatdan:
@mtohmaz Please rebase this pull request.

mtohmaz:
Just rebased, added changes to allow for atomic unmount as non-root.

rhatdan:
@mtohmaz Please update the man page to explain what is going on.

mtohmaz:
Updated the man page for both mount and unmount.

Atomic/syscontainers.py
```diff
@@ -106,7 +106,12 @@ def _checkout_layer(self, repo, rootfs_fd, rootfs, rev):
                             stdin=DEVNULL,
                             stdout=DEVNULL,
                             stderr=DEVNULL)
-
+            if self.user:
+                for root,dirs,_ in os.walk(rootfs):
+                    for d in dirs:
+                        dirname = os.path.join(root,d)
+                        os.chmod(dirname,os.stat(dirname).st_mode | stat.S_IWUSR)
+
```
giuseppe: this should not be necessary now, as it is already done on importing the layers.  Could you verify that it works without this code?

giuseppe:
it doesn't work with containers though.  Should we give a better error message when an user tries `atomic mount $CONTAINER_NAME`?  We could probably hard link files from the containers rootfs (and fallback to copy if it fails) to the mount destination.  We cannot use the image in this case as the container could be based on an older version of the image.

rhatdan:
What does that mean, Do we support read/write container images on ostree?  Not sure there is a difference between a container and an image if you are running with R/O Ostree.

rhatdan:
Well in the case of containers on the host, we can commit the image of the container, and then mount that separately.

giuseppe:
the difference could be that the container could be based on an older version of the image.  If you pull a new version for the image, the container will still be using the older version, while the image has a new version.

We have to be careful that we mount/copy exactly the same files as seen by the container, as scan will be using them.

rhatdan:
@giuseppe what should we do with this?  Should we merge and then fix syscontainers?

giuseppe:
ok let's merge it, I'll have a look at the issue I reported later.

rhatdan:
Waiting for https://github.com/projectatomic/atomic/pull/570 to finish to see if that screws up this pull request and then will merge.

rhatdan:
@rh-atomic-bot r+

PR #510, opened 2016-08-02
Implement ps functionality for DBus API

AmartC:
There is one thing I am awaiting feedback on and afterwards I will submit a new PR. In this case, is a separate tty function needed? The useTTY flag I feel would come in handy here however, I can change it to use a tty function if necessary.

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -26,6 +26,7 @@ class Scan(Atomic):
         self.debug = False
         self.rootfs_mappings = {}
         self.scanner = None
+        self.mount_paths = {}
 
     def get_scanners_list(self):
         return json.dumps(self.scanners)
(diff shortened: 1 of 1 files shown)


baude:
please use a commit message with a title (like you do) and then expand on what the PR does subsequently.

atomic_client.py
```diff
@@ -77,6 +77,10 @@ def images(self):
     def vulnerable(self):
         return self.dbus_object.VulnerableInfo(dbus_interface="org.atomic", timeout = 2147400)
 
+    @polkit.enable_proxy
+    def ps(self):
+        return self.dbus_object.Ps(dbus_interface="org.atomic", timeout = 2147400)
```
baude: this may have been discussed before, but why are we setting a timeout limit?

atomic_dbus.py
```diff
@@ -264,6 +266,17 @@ def VulnerableInfo(self):
         self.atomic.set_args(args)
         return self.atomic.get_all_vulnerable_info()
 
+    # The Ps method will send back information that says
+    # whether or not an installed container image is vulnerable
```
baude: i thought atomic ps == docker ps ?

atomic_client.py
AmartC: If the timeout limit is not set, then there is the possibility that there can be a timeout before the data is sent back and output.

baude: Yes of course, i understand that ... but that timeout seems arbitrary and therefore could also be problematic.  would martin's approach here make sense?

AmartC: possibly this is the timeout though set for the test python client however. in cockpit, the client doesn't timeout. stefw had recommended this timeout length for me.

PR #512, opened 2016-08-03
[merged] Handle cases where container does not have a Status key

AmartC:
Have iteration through image start at index 0 instead of index 2. This fixes up an earlier patch.

self.args is an instance at times and cannot be iterated through then. So check way to check if display is present in self.args using hasattr().


Atomic/scan.py
```diff
@@ -170,11 +170,12 @@ def gen_images():
         def gen_containers():
             slist = []
             for con in self.get_containers():
-                if con['Status'].upper() != 'DEAD':
-                    con['input'] = con['Id']
-                else:
-                    raise ValueError("The container ID {} is Dead and therefor cannot be scanned.".format(con['Id'][:12]))
-                slist.append(con)
+                if 'Status' in con:
```
baude: I've grown fond of using .get for these cases.  What do you think of something like:

if con.get('Status').upper() != ...

Atomic/scan.py
```diff
@@ -170,11 +170,12 @@ def gen_images():
...
+                    if con['Status'].upper() != 'DEAD':
+                        con['input'] = con['Id']
+                    else:
+                        raise ValueError("The container ID {} is Dead and therefor cannot be scanned.".format(con['Id'][:12]))
```
baude: Type, therefor should be therefore

rhatdan: Just end at "Container ID FOOBAR is Dead"

Atomic/scan.py
rhatdan: Why do you like that better then con['Status'].upper()

baude: Rather than having the IF conditional to see if Status exists in con and then getting the value, .get combines both.  If Status does not exist, then the value is None.

AmartC: Thing is though when I try using the .get when there is a none type it says none type has no attribute called upper.

rhatdan:
Right I see an exception being thrown.

rhatdan:
@rh-atomic-bot r+

PR #515, opened 2016-08-04
[merged] Show all system containers in ps and the running command

giuseppe:


tests/integration/test_system_containers.sh
```diff
@@ -152,21 +152,32 @@ image_digest=$(ostree --repo=${ATOMIC_OSTREE_REPO} show --print-metadata-key=doc
...
 
-${ATOMIC} images list > images.out
-${ATOMIC} images list --all > images.all.out
+${ATOMIC} images list -f type=system > images.out
+${ATOMIC} images list -f type=system --all > images.all.out
 test $(wc -l < images.out) -lt $(wc -l < images.all.out)
 grep -q '<none>' images.all.out
-OUTPUT=$(! grep -q '<none>' images.out)
+OUTPUT=$(! grep -c '<none>' images.out)
 if test $OUTPUT \!= 0; then
```
cgwalters: This seems odd now...are we just asserting that `images.out` is empty then?

Atomic/syscontainers.py
```diff
@@ -394,7 +394,12 @@ def get_system_containers(self):
                 created = info["created"] if "created" in info else ""
                 image = info["image"] if "image" in info else ""
 
-            container = {'Image' : image, 'ImageID' : revision, 'Id' : x, 'Created' : created, 'Names' : [x]}
+            with open(os.path.join(fullpath, "config.json"), "r") as config_file:
+                config = json.loads(config_file.read())
```
cgwalters: A bit more efficient :racehorse:  to do `json.load(config_file)`.

tests/integration/test_system_containers.sh
giuseppe: I'll revert it, probably the `if test` check later is not needed, as grep would return 0 in case of matches

giuseppe:
I added some fixup patches to address the comments

cgwalters:
@rh-atomic-bot r+ 83e1d5b

PR #520, opened 2016-08-05
[merged] syscontainer: check if system image by ID

stephen679:
`has_system_container_image()` only returns True if user specifies the actual name of a system image. It should also return True if user specifies the same image but with the ID instead.

Signed-off-by: Stephen Chung <schung@redhat.com>


giuseppe:
bot, test pull request

Atomic/syscontainers.py
```diff
@@ -743,6 +743,9 @@ def has_system_container_image(self, img):
         if not repo:
             return False
         try:
+            for i in self.get_system_images(get_all=True):
+                if i['Id'].startswith(img):
+                    return True
```
giuseppe: instead of retrieving all the images here, we should check if the ostree branch exists.

If `repo.resolve_rev(imagebranch, False)[0]` fails then you can try to resolve `OSTREE_OCIIMAGE_PREFIX/$img` and see if it exists

rhatdan:
bot, add author to whitelist

Atomic/syscontainers.py
stephen679: since the patch for `verify` now stores the images in `OSTREE_OCIIMAGE_PREFIX` with the "oci-sha256", is there a clean way to convert the image id to the oci sha256?

Otherwise, would:
```
        try:
            return repo.resolve_rev(imagebranch, False)[0]
        except: #pylint: disable=bare-except
            for i in self.get_system_images(get_all=True):
                if i['Id'].startswith(img):
                    return True
            return False
```
be an acceptable solution?

rhatdan:
@rh-atomic-bot r+

PR #521, opened 2016-08-07
[merged] Cleanup atomic command minor fixes

rhatdan:
1 Remove debugging output when I mount an ostree image.
2 Cleanup cmd_env to not modify current processes os.environ

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -304,20 +304,21 @@ class Atomic(object):
         return list(map(pipes.quote, args))
 
     def cmd_env(self):
-        os.environ['NAME'] = self.name or ""
-        os.environ['IMAGE'] = self.image or ""
+        newenv = dict(os.environ)
+        newenv['NAME'] = self.name or ""
+        newenv['IMAGE'] = self.image or ""
 
         if hasattr(self.args, 'opt1') and self.args.opt1:
-            os.environ['OPT1'] = self.args.opt1
+            newenv['OPT1'] = self.args.opt1
 
         if hasattr(self.args, 'opt2') and self.args.opt2:
-            os.environ['OPT2'] = self.args.opt2
+            newenv['OPT2'] = self.args.opt2
 
         if hasattr(self.args, 'opt3') and self.args.opt3:
-            os.environ['OPT3'] = self.args.opt3
+            newenv['OPT3'] = self.args.opt3
 
         if not hasattr(self.args, 'PWD'):
-            os.environ['PWD'] = os.getcwd()
+            newenv['PWD'] = os.getcwd()
 
         default_uid = "0"
         with open("/proc/self/loginuid") as f:
@@ -325,15 +326,15 @@ class Atomic(object):
             if int(val) <= 2147483647:
                 default_uid = val
 
-        if "SUDO_UID" not in os.environ:
-            os.environ["SUDO_UID"] = default_uid
+        if "SUDO_UID" not in newenv:
+            newenv["SUDO_UID"] = default_uid
 
-        if 'SUDO_GID' not in os.environ:
-            os.environ["SUDO_GID"] = default_uid
+        if 'SUDO_GID' not in newenv:
+            newenv["SUDO_GID"] = default_uid
 
         if self.run_opts is not None:
-            os.environ["RUN_OPTS"] = self.run_opts
-        return os.environ
+            newenv["RUN_OPTS"] = self.run_opts
+        return newenv
 
     def gen_cmd(self, cargs):
         args = []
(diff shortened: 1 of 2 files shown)


Atomic/atomic.py
```diff
@@ -553,17 +553,17 @@ def cmd_env(self):
         os.environ['NAME'] = self.name or ""
         os.environ['IMAGE'] = self.image or ""
 
-        if hasattr(self.args, 'opt1') and self.args.opt1:
+        if self.args.get('opt1', False):
             os.environ['OPT1'] = self.args.opt1
```
cgwalters: Hmm...not related to this change but this looks a bit dangerous - we're mutating the current process' environment, which means that conceptually we can only do one action at a time.  If we ever supported running multiple containers in one go, this would have to change.

It's normally better regardless to prepare the environment for the child process rather than mutating self's.  Anyways, not a blocker for this patch, just noting.

cgwalters:
Looks like the first commit was written by `root@`?

Atomic/atomic.py
rhatdan: Yes good point, I would love to drop most of the opt handling, since we are now passing the environment from the atomic command into the containers runtime.

rhatdan:
@cgwalters Fixed root->dwalsh
Fixed cmd_env to use a copy of os.environ()
Now container three patches.

Atomic/atomic.py
```diff
@@ -550,33 +550,34 @@ def uninstall(self):
             util.check_call([self.docker_binary(), "rmi", self.image])
 
     def cmd_env(self):
-        os.environ['NAME'] = self.name or ""
-        os.environ['IMAGE'] = self.image or ""
+        newenv = os.environ
```
cgwalters: I think you need `newenv = dict(os.environ)` or so.

Atomic/atomic.py
```diff
@@ -554,17 +554,17 @@ def cmd_env(self):
...
+        if self.args.get('opt2', False):
             newenv['OPT2'] = self.args.opt2
 
-        if hasattr(self.args, 'opt3') and self.args.opt3:
+        if self.args.get('opt3', False):
             newenv['OPT3'] = self.args.opt3
 
-        if not hasattr(self.args, 'PWD'):
-            newenv['PWD'] = os.getcwd()
+        if not hasattr(newenv, 'PWD'):
```
cgwalters: I think this is more commonly spelled `if 'PWD' not in newenv:`

adelton:
The reason for tests failing is the

```
    def sub_env_strings(self, in_string):
        """
        Performs substitutions on an input string based on defined
        environment variables.
        :param in_string: string to perform the subs on
        :return: string
        """
        # Set environment variables
        self.cmd_env()

        # Perform variable subs
        in_string = os.path.expandvars(in_string)

        # Replace undefined variables with blank
        in_string = re.sub(r'\$\{?\S*\}?', '', in_string)

        # Solve whitespacing
        in_string = " ".join(in_string.split())

        return in_string
```

That `os.path.expandvars` uses `os.environ` and cannot be passed different `env`.

rhatdan:
@cgwalters What do you think?  We could change the environment expand and then change it back?

adelton:
> @cgwalters What do you think? We could change the environment expand and then change it back?

Or reimplement `os.path.expandvars` to use/accept `self.cmd_env()`.

By the way, why do you want to get rid of the `os.environ` modifications? The changes will go away when the process exits ...

rhatdan:
How about this far simpler fix, Also @adelton does this fix your other problem where we were substituting twice.

cgwalters:
@adelton Regarding mutating `os.environ` - it's not threadsafe.  https://sourceware.org/bugzilla/show_bug.cgi?id=15607

adelton:
How many threads does a typical `atomic` command run?

rhatdan:
1

adelton:
I don't have enough Python foo to say if it sould be `dict(os.environ)` or `os.environ.copy()` and whether that `os.environ=oldenv` does the right thing.

In any case, this patch now does not get rid of the modification of `os.environ` in `cmd_env()` which I thought was your goal and which the commit message still describes.

rhatdan:
I think we can modify it, as long as we are about to exec.  Modifying it permanently or running it multiple times can cause problems.

cgwalters:
Maybe 1 thread right now, but as soon as e.g. we port to `GDBus` it'll run a thread internally.  Using GLib at all really may make your process multithreaded (there's also a worker thread).

rhatdan:
I guess it would be best if somone took os.path.expandvars and made it take an environment param

rhatdan:
No longer needed.

rhatdan:
Added utils.expandvars.  But I am not sure if this will work properly on python2 and python3

rhatdan:
@adelton PTAL

rhatdan:
My testing looks good for python3 and python2 and the test suite passed.

adelton:
LGTM.

rhatdan:
@cgwalters PTAL

cgwalters:
@rh-atomic-bot r+ 9d5aa41

PR #523, opened 2016-08-08
Add functionality for DBus API to be able to know when last scan happened

AmartC:


atomic_dbus.py
```diff
@@ -270,12 +279,12 @@ def VulnerableInfo(self):
     @dbus.service.method("org.atomic", in_signature='', out_signature='s')
     def Ps(self):
         ps = Ps()
-        ps.useTTY = False
```
rhatdan: Why?

atomic_dbus.py
```diff
@@ -270,12 +279,12 @@ def VulnerableInfo(self):
...
         args = self.Args()
         ps.set_args(args)
         return json.dumps(ps.ps())
 
 
+
 if __name__ == "__main__":
```
rhatdan: Cleanup white space.

Atomic/scan.py
```diff
@@ -87,6 +87,10 @@ def set_scanner():
             raise ValueError(yaml_error)
 
         self.results_dir = os.path.join(self.results, self.scanner, self.cur_time)
+        last_scanned_file = os.path.join(self.results, "last_scanned.json")
+        with open(last_scanned_file, 'w') as f:
+            json.dump(self.cur_time, f, indent=4)
+
```
rhatdan: @baude could you look at this?  Seems like there should be a much easier way to figure out when the last scan happened.  Also seems like this check should be per container image, scanner, scanner type.  Just because something was scanned does not mean the image a user is looking at was scanned.

atomic_dbus.py
AmartC: I added a ps_tty function so the useTTY flag is no longer needed.

Atomic/scan.py
AmartC: This was one approach I thought of. Another involves comparing different timedate objects which I was looking into. In the case of Andreas' design, it has a box which lists how many containers and images are vulnerable and then lists the time of the last scan.

baude: The persistent data file has the scan time in it and is json.  This seams like the most straight forward approach.

Atomic/atomic.py
```diff
@@ -1109,6 +1110,27 @@ def set_debug(self):
         if self.args.debug:
             self.debug = True
 
+    def get_last_time_scanned(self):
+        """
+        Will read through the scan_summary.json file and get the most recent scan's time.
+        """
+        try:
+            summary_results = json.loads(open(os.path.join(self.results, "scan_summary.json"), "r").read())
+            mostRecentTime = ""
```
rhatdan: Lets use the with syntax.

```
with open(os.path.join(self.results, "scan_summary.json"), "r") as scan_summary:
       summary_results = json.load(scan_summary.read())
```

Atomic/atomic.py
```diff
@@ -1109,6 +1110,27 @@ def set_debug(self):
...
+            currentTime = ""
+            for uuid in summary_results.keys():
+                if "Time" in summary_results[uuid]:
+                    if not mostRecentTime:
+                        mostRecentTime = datetime.strptime(summary_results[uuid]["Time"].replace("T", " "), "%Y-%m-%d %H:%M:%S")
+                    else:
+                        currentTime = datetime.strptime(summary_results[uuid]["Time"].replace("T", " "), "%Y-%m-%d %H:%M:%S")
+                        if currentTime > mostRecentTime:
```
rhatdan: Can't I compare these times without doing the replace?

AmartC:
bot retest please

AmartC:
bot retest this please

AmartC:
bot, retest this please

rhatdan:
@baude PTAL

rhatdan:
@rh-atomic-bot retest

rhatdan:
bot, retest this please

Atomic/atomic.py
```diff
@@ -1113,6 +1114,29 @@ def set_debug(self):
         if self.args.debug:
             self.debug = True
 
+    def get_last_time_all_scanned(self):
+        """
+        Will read through the scan_summary.json file and get the most recent scan's time.
+        """
+        try:
```
baude: How about you make a generic function that reads in the scan_summary file?

Atomic/atomic.py
```diff
@@ -1113,6 +1114,29 @@ def set_debug(self):
...
+            with open(os.path.join(self.results, "scan_summary.json"), "r") as scan_summary:
+                summary_results = json.loads(scan_summary.read())
+            mostRecentTime = ""
```
baude: Just a nit, but we havent been using camel case but rather underscores

Atomic/scan.py
```diff
@@ -188,6 +188,17 @@ def gen_containers():
             scan_list = []
             images = self.get_images()
             containers = self.get_containers()
+            all_names = []
```
baude: What is this code block for?

Atomic/atomic.py
```diff
@@ -1113,6 +1114,29 @@ def set_debug(self):
...
+            currentTime = ""
+            for uuid in summary_results.keys():
+                if len(summary_results[uuid].keys()) > 0:
+                    if summary_results[uuid]["All"]:
+                        if not mostRecentTime:
```
baude: can you explain the logic behind this conditional?

AmartC: So upon speaking with Andreas from the Cockpit team, he says that there should be a box displayed on the UI that lists information from when all containers and images were scanned and then display the results from that. Thus, I will be going through the scan_summary.json file and specifically looking for the cases when everything was scanned and take information about the time of the last scan.

Atomic/scan.py
AmartC: This is designed for the Cockpit UI if the user manually selects all images and containers, then I can check to see if their scan_targets matches all the images and containers listed in their system.

Atomic/atomic.py
AmartC: That can be done but one of them is to get the vulnerable ids and the idea of this function is to specifically for the most recent scan time. I can look into modifying it if necessary.

baude: Looking closer, there is already a generic method.  It is get_all_vulnerable_info in atomic.py and you should use it.

rhatdan:
@baude could you take this pull request over since @AmartC is going back to school

baude:
@rhatdan I sent an email to Andreas last week with some questions about this PR. I'm awaiting his response.

PR #524, opened 2016-08-09
[merged] [RFC] User containers

giuseppe:
user containers use bwrap-oci (https://github.com/giuseppe/bubblewrap/tree/bwrap-oci) to convert from the OCI configuration, which is already used for runc, to the bubblewrap command line.

I dropped `--user` from pull, as it should happen by default when the user is not root.

I added `--user` to install, so that it is clear that the user wants to install an user container, and also leave space in future to add new types of containers.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -296,6 +296,11 @@ class Atomic(object):
         except (NameError, AttributeError):
             pass
 
+        try:
+            self.user = args.user
+        except (NameError, AttributeError):
+            pass
+
         if not self.name and self.image is not None:
             self.name = self.image.split("/")[-1].split(":")[0]
             if self.spc:
@@ -806,10 +811,14 @@ class Atomic(object):
         if self._container_exists(self.name):
             raise ValueError("A container '%s' is already present" % self.name)
 
-        if self.system:
+        if self.user:
+            if not util.is_user_mode():
+                raise ValueError("--user does not work for privileged user")
+            return self.syscontainers.install_user_container(self.image, self.name)
+        elif self.system:
             return self.syscontainers.install_system_container(self.image, self.name)
         elif self.args.setvalues:
-            raise ValueError("--set is valid only when used with --system")
+            raise ValueError("--set is valid only when used with --system or --user")
 
         self._check_if_image_present()
         args = self._get_args("INSTALL")
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -387,11 +387,8 @@ class Atomic(object):
             if self.syscontainers.has_system_container_image(image):
                 return self.syscontainers.inspect_system_image(image)
             return self.d.inspect_image(image)
-        except NotFound:
+        except (NotFound, requests.exceptions.ConnectionError):
             pass
-        except requests.exceptions.ConnectionError:
-            raise NoDockerDaemon()
-
         return None
 
     def _inspect_container(self, name=None):
@@ -399,10 +396,8 @@ class Atomic(object):
             name = self.name
         try:
             return self.d.inspect_container(name)
-        except NotFound:
+        except (NotFound, requests.exceptions.ConnectionError):
             pass
-        except requests.exceptions.ConnectionError:
-            raise NoDockerDaemon()
         return None
 
     def _get_args(self, label):
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -328,7 +328,7 @@ class SystemContainers(object):
     def _get_ostree_repo_location(self):
         if self.user:
             home_dir = os.getenv("HOME")
-            return os.path.expanduser("%s/ostree/repo" % home_dir)
+            return os.path.expanduser("%s/.containers/repo" % home_dir)
         else:
             return os.environ.get("ATOMIC_OSTREE_REPO") or \
                 self.get_atomic_config_item(["ostree_repository"]) or \
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -12,6 +12,7 @@ import subprocess
 import time
 from .client import AtomicDocker
 from ctypes import cdll, CDLL
+from dateutil.parser import parse as dateparse
 
 try:
     import gi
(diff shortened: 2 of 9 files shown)


Atomic/syscontainers.py
```diff
@@ -171,6 +180,9 @@ def _generate_default_oci_configuration(self, destination):
             conf.write(json.dumps(configuration, indent=4))
 
     def _generate_systemd_startstop_directives(self, name):
+        if self.user:
+            return ["/usr/bin/bwrap-oci '%s'" % (name), ""]
```
rhatdan: Should probably be a constant?

Atomic/syscontainers.py
```diff
@@ -186,7 +198,11 @@ def _checkout_system_container(self, repo, name, img, deployment, upgrade, value
         destination = destination or "%s/%s.%d" % (self._get_system_checkout_path(), name, deployment)
         exports = os.path.join(destination, "rootfs/exports")
         unitfile = os.path.join(exports, "service.template")
-        unitfileout = os.path.join(SYSTEMD_UNIT_FILES_DEST, "%s.service" % name)
+        if self.user:
+            home_dir = os.getenv("HOME")
+            unitfileout = os.path.join(os.path.expanduser(SYSTEMD_UNIT_FILES_DEST_USER % home_dir), "%s.service" % name)
```
rhatdan: Theoretically this could fail.
```
from os.path import expanduser
home = expanduser("~")
```

Might be better.

Atomic/syscontainers.py
```diff
@@ -321,14 +337,20 @@ def _write_template(inputfilename, data, values, outfile):
...
 
     def _get_system_checkout_path(self):
-        return os.environ.get("ATOMIC_OSTREE_CHECKOUT_PATH") or \
-            self.get_atomic_config_item(["checkout_path"]) or \
-            "/var/lib/containers/atomic"
+        if os.environ.get("ATOMIC_OSTREE_CHECKOUT_PATH"):
+            return os.environ.get("ATOMIC_OSTREE_CHECKOUT_PATH")
+        if self.get_atomic_config_item(["checkout_path"]):
+            self.get_atomic_config_item(["checkout_path"])
+        if self.user:
```
rhatdan: Should this be a return?

Atomic/syscontainers.py
```diff
@@ -321,14 +337,20 @@ def _write_template(inputfilename, data, values, outfile):
         return
 
     def _get_system_checkout_path(self):
-        return os.environ.get("ATOMIC_OSTREE_CHECKOUT_PATH") or \
-            self.get_atomic_config_item(["checkout_path"]) or \
-            "/var/lib/containers/atomic"
+        if os.environ.get("ATOMIC_OSTREE_CHECKOUT_PATH"):
+            return os.environ.get("ATOMIC_OSTREE_CHECKOUT_PATH")
+        if self.get_atomic_config_item(["checkout_path"]):
+            self.get_atomic_config_item(["checkout_path"])
```
rhatdan: Should this be a return?  Should this only happen if this is not self.user?

giuseppe: yes, thanks, I will fix it

giuseppe:
I addressed the reported issues in the last version

Atomic/atomic.py
```diff
@@ -808,8 +808,11 @@ def _check_if_image_present(self):
             self.inspect = self._inspect_image()
 
     def install(self):
-        if self._container_exists(self.name):
-            raise ValueError("A container '%s' is already present" % self.name)
+        try:
+            if self._container_exists(self.name):
+                raise ValueError("A container '%s' is already present" % self.name)
+        except NoDockerDaemon:
+            pass
```
rhatdan: My previous patch set should fix this. Could you verify you are up2date.  

https://github.com/projectatomic/atomic/pull/505

docs/atomic-install.1.md
```diff
@@ -77,6 +77,10 @@ Installing a system container consists of checking it the image by
 default under /var/lib/containers/atomic/ and generating the
 configuration files for runc and systemd.
 
+**--user**
+If running as non-root, specify to install the image from the current
+OSTree repository and manage it through systemd and bubblewrap.
```
rhatdan: Why not do the imply --user if running as non root?  Are you intent on treating run differently then install?

Atomic/atomic.py
```diff
@@ -806,10 +811,14 @@ def install(self):
         if self._container_exists(self.name):
             raise ValueError("A container '%s' is already present" % self.name)
 
-        if self.system:
+        if self.user:
+            if not util.is_user_mode():
+                raise ValueError("--user does not work for privileged user")
+            return self.syscontainers.install_user_container(self.image, self.name)
```
rhatdan: If we did this up in the parser code, it would be a lot cleaner.

Atomic/atomic.py
```diff
@@ -806,10 +811,14 @@ def install(self):
...
+        elif self.system:
             return self.syscontainers.install_system_container(self.image, self.name)
         elif self.args.setvalues:
-            raise ValueError("--set is valid only when used with --system")
+            raise ValueError("--set is valid only when used with --system or --user")
```
rhatdan: Same we should do this in the parser code.  Can we check at that point if we are doing --system

docs/atomic-install.1.md
giuseppe: I think it is better that for the `install`command the user is explicit on what kind of container is going to be installed.  We will be able in future to support other kind of containers as well.

Atomic/atomic.py
giuseppe: this exception is raised from `Atomic._inspect_image`

rhatdan: Can we do it at that layer?

giuseppe:
I've changed `_inspect_container` and `_inspect_image` to not raise `NoDockerDaemon`

rhatdan:
Looks good, can you rebase everything and we can merge

giuseppe:
there is still stuff left to do, but it can be added on top of this PR.  The patches are already rebased onto latest master

rhatdan:
Ok lets bring this in.

@rh-atomic-bot r+

PR #525, opened 2016-08-09
[merged] Atomic/diff.py: Fix --json output/formatting

baude:
With the tty changes, we were sending a str to the
util.output_json method which expects a dict|json.


rhatdan:
@rh-atomic-bot r+

PR #526, opened 2016-08-09
[merged] Atomic/scan.py Output results as JSON

baude:
The IT team working with the cert program requested that they
be able to output the scan results as json.


rhatdan:
Didn't the dbus stuff already do this?

baude:
Nope, which is why I did.

rhatdan:
@rh-atomic-bot r+

PR #527, opened 2016-08-09
[merged] syscontainers: add --remote to atomic install

yuqi-zhang:
Add --remote=REMOTE to atomic install, which allows users to specify
an existing exploded container's location as REMOTE. The remote
container's rootfs will be used as a read-only rootfs for the new
container to be installed, and therefore the new container would
only contain configuration files. (Note: currently all system
containers must have a read-only rootfs by default).

A use case for this would be using existing container's rootfs
through NFS, etc to serve as the rootfs for many containers running
the same image. This avoids duplicate storage of having a container
on each machine in a cluster.

A user can update only the remote container's config file parameters
with 'atomic update', and uninstalling a remote container does not
affect the container specified in REMOTE.

@giuseppe PTAL


rhatdan:
I am not crazy about the name remote, should this just be rootfs to point to a container rootfs?

yuqi-zhang:
The original idea was to use an existing expanded container (could just be mounted at a location), although we currently really just use the rootfs for the container. Either way is ok for me, what do you think @giuseppe

Atomic/syscontainers.py
```diff
@@ -288,8 +309,13 @@ def _write_template(inputfilename, data, values, outfile):
                 missing = {x[1] for x in template.pattern.findall(data, template.flags) if len(x[1]) > 0 and x[1] not in values} # pylint: disable=no-member
                 raise ValueError("The template file %s still contains unreplaced values for: %s" % \
                                  (inputfilename, ", ".join(missing)))
-
-            outfile.write(result)
+            if self.args.remote and "config" in inputfilename:
+                config = json.loads(result)
+                config['root']['remote'] = True
```
giuseppe: is ``config['root']['remote']`` used by runc?  Can we drop it?

docs/atomic-install.1.md
```diff
@@ -60,6 +61,11 @@ If --display is not specified the install command will execute.
  Use this name for creating installed content for the container.
  NAME will default to the IMAGENAME if it is not specified.
 
+**--remote=REMOTE**
+Specify a REMOTE folder, which should be an existing, expanded
+container. The existing container's rootfs will be used as the
+new system container's rootfs. It is only used by --system.
```
giuseppe: it should work with `--user` as well

Atomic/syscontainers.py
```diff
@@ -174,7 +174,11 @@ def _generate_default_oci_configuration(self, destination):
         with open(conf_path, 'r') as conf:
             configuration = json.loads(conf.read())
         configuration['root']['readonly'] = True
-        configuration['root']['path'] = 'rootfs'
+        if self.args.remote:
+            configuration['root']['path'] = self.args.remote + "/rootfs"
```
giuseppe: the `_generate_default_oci_configuration` function is used only for the images that have not a configuration file under the `/exports` directory.

We need to set ``configuration['root']['path']`` after the configuration file has been generated, either the default one as covered here, or if it is copied directly or through the template file.

My suggestion is that this is done just before `_check_oci_configuration_file` is called.

Later we will need to refactor this part though, as the file is opened multiple times, but this can be done out of this patch.

Atomic/syscontainers.py
```diff
@@ -786,6 +821,7 @@ def extract_system_container(self, img, destination):
         repo = self._get_ostree_repo()
         if not repo:
             return False
+        self.args.remote = None
```
giuseppe: I would store this information in the `info` file as well, it could be used by `update_system_container` without checking for the existence of "roots"

docs/atomic-install.1.md
giuseppe: I actually tested it and it works (after I manually changed the rootfs in the configuration file, the image I've used had a template conf file), so the change in the documentation is enough.

Atomic/syscontainers.py
```diff
@@ -232,28 +249,32 @@ def _checkout_system_container(self, repo, name, img, deployment, upgrade, value
...
+                    layers = SystemContainers.get_layers_from_manifest(json.loads(manifest))
+                    for layer in layers:
+                        rev_layer = repo.resolve_rev("%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer.replace("sha256:", "")), False)[1]
+                        self._checkout_layer(repo, rootfs_fd, rootfs, rev_layer)
+                self._do_syncfs(rootfs, rootfs_fd)
+            finally:
+                if rootfs_fd:
+                    os.close(rootfs_fd)
+
+            if extract_only:
```
giuseppe: the `extract_system_container()` function is used my mount so that the content can be inspected.

Remote containers should honor `extract_only` as well, let's just create a symlink for now, we can fix mount later.



Perhaps we must create a symlink to the actual rootfs?

Atomic/syscontainers.py
```diff
@@ -197,7 +201,11 @@ def _checkout_system_container(self, repo, name, img, deployment, upgrade, value
         imagebranch = SystemContainers._get_ostree_image_branch(img)
 
         destination = destination or "%s/%s.%d" % (self._get_system_checkout_path(), name, deployment)
-        exports = os.path.join(destination, "rootfs/exports")
+        if self.args.remote:
```
giuseppe: instead of using `self.args.remote` in different places, I would instead set `self.remote` as:

`self.remote = self._resolve_remote_path(self.args.remote)`

For now the `_resolve_remote_path` can just resolve symlinks, but we will change later to resolve images exploded from the Atomic Registry.

You can then drop the control in `update_system_container` as the rootfs might change as well, if the `_resolve_remote_path` gives a different result (e.g. the symlink points to a different version).

Atomic/syscontainers.py
giuseppe: the end result should be that `atomic mount` will create a ro bind mount for these containers

yuqi-zhang:
Will work on, thanks!

yuqi-zhang:
Refactored:

- dropped unused config
- updated docs slightly
- added "remote" to info file
- added "_resolve_remote_path" function
- allowed user to select a rootfs directly, as well as a container location to use as the remote
- fixed error when mounting

Atomic/syscontainers.py
```diff
@@ -398,12 +434,18 @@ def _get_ostree_repo(self):
...
         repo = self._get_ostree_repo()
         if not repo:
             raise ValueError("Cannot find a configured OSTree repo")
 
         path = os.path.join(self._get_system_checkout_path(), name)
+        with open(os.path.join(path, "info"), 'r') as info_file:
+            info = json.loads(info_file.read())
+            self.args.remote = info['remote']
+            if self.args.remote:
+                util.write_out("Updating a container with a remote rootfs. Only changes to config will be applied.")
```
giuseppe: if we use the original path used for the installation and it resolves to a different location, i.e. the symlink points somewhere else, then the container is also updated.  Should we remove this warning?

giuseppe:
great work Jerry.  I have reviewed and tested it.  If everyone agrees on the name `--remote`, I think it is ready to be merged

rhatdan:
Please give me a good description of what you guys believe --remote means.  I am still thinking we should use a different named option.

giuseppe:
this option allows us to use an already exploded to disk image.  Most likely this will happen with images that are mounted from a network file system.  My initial name for this feature was `detached` mode, as the configuration file is detached from the rootfs of the image.  What about `--rootfs=PATH` or `--image-path=PATH`?

Atomic/syscontainers.py
yuqi-zhang: Hm so one of two things could happen:

1. the user runs --update on an existing container which HAS a rootfs, and that rootfs is also in use by other containers as a remote rootfs. Then the rootfs gets updated, BUT the remote ones still point to the old (since it resolved the true path instead of the symlink). i.e. the container goes to container.1 but the remotes still use container.0

2. the user runs update DIRECTLY on a container with a remote rootfs, which cannot modify the remote rootfs, but can update the config.

In hindsight this is probably not a good way to do it. I think what I will have it do is: if the original container is updated, all the remote automatically use the new rootfs (by just saving the symlink) unless the user specifies not to.

yuqi-zhang:
--remote was originally intended to mean a "remote folder" where an existing container or checked out image exists, so a container can use the rootfs "remotely" and not have one itself. I guess an alternative would be "rootfs-location" for better clarification?

rhatdan:
--remote is not a good name for that.  Since I believe remote would indicate that the content is on a different system.  --rootfs would be a better indication.  Or something like --root.

rpm --root is a good example.

yuqi-zhang:
Refractored the command to --rootfs, and made minor fix to info file.

Atomic/syscontainers.py
giuseppe: I think it is safer this way, so that we can control when the container is updated, we should just improve the way versions are managed somehow and keep track of them in the container info.

For containers with an already exploded rootfs, updating the configuration file to point to another version is an update, so I am still wondering if we could drop this warning as it looks to me that we are just recreating the config file (i.e. just set new values if specified via `--set`) but no update is really taking place.

giuseppe:
@rhatdan if you are fine with the name `--rootfs`, I would merge this patch

rhatdan:
LGTM

giuseppe:
@rh-atomic-bot r+ 9bac7a3

PR #529, opened 2016-08-10
[merged] Improve error messages system containers

giuseppe:
Fail early if runc is not installed when installing a system container, or either if bwrap-oci is not installed or systemctl doesn't support --user when installing an user container.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -48,7 +48,8 @@ WorkingDirectory=$DESTDIR
 [Install]
 WantedBy=multi-user.target
 """
-BWRAP_OCI_PATH="/usr/bin/bwrap-oci"
+BWRAP_OCI_PATH = "/usr/bin/bwrap-oci"
+RUNC_PATH = "/bin/runc"
 
 class SystemContainers(object):
 
@@ -141,6 +142,14 @@ class SystemContainers(object):
         return
 
     def install_user_container(self, image, name):
+        try:
+            util.check_call([BWRAP_OCI_PATH, "--version"], stdout=DEVNULL)
+        except util.FileNotFound:
+            raise ValueError("Cannot install the container: bwrap-oci is needed to run user containers")
+
+        if not "--user" in str(util.check_output(["systemctl", "--help"], stdin=DEVNULL, stderr=DEVNULL)):
+            raise ValueError("Cannot install the container: systemctl does not support --user")
+
         # Same entrypoint
         return self.install_system_container(image, name)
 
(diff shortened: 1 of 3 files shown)


Atomic/syscontainers.py
```diff
@@ -141,6 +142,18 @@ def pull_image(self):
...
 
     def install_user_container(self, image, name):
+        bwrap_oci_found = True
+        try:
+            util.check_call([BWRAP_OCI_PATH, "--version"], stdout=DEVNULL)
+        except util.FileNotFound:
+            bwrap_oci_found = False
+
+        if not bwrap_oci_found:
+            raise ValueError("Cannot install the container: bwrap-oci is needed to run user containers")
```
rhatdan: Why node raise this in the exception?

Atomic/syscontainers.py
```diff
@@ -141,6 +142,18 @@ def pull_image(self):
...
+
+        if not "--user" in str(util.check_output(["systemctl", "--help"], stdin=DEVNULL, stderr=DEVNULL)):
+            raise ValueError("Cannot install the container: systemctl does not support --user")
+
```
rhatdan: I think this checking should be done in the argparser.

Atomic/syscontainers.py
```diff
@@ -149,6 +162,16 @@ def install_system_container(self, image, name):
...
+        if not self.user:
+            runc_found = True
+        try:
+            util.check_call([RUNC_PATH, "--version"], stdout=DEVNULL)
+        except util.FileNotFound:
+            runc_found = False
+
+        if not runc_found:
+            raise ValueError("Cannot install the container: runc is needed to run system containers")
+
```
rhatdan: Why not raise this in the exception.  Don't see the need for runc_found.

giuseppe: Python3 complains when you raise an exception while handling another one

Atomic/syscontainers.py
giuseppe: the advantage of doing it here is that Atomic.syscontainers can be used as a library from other applications and this check would still be done

Atomic/syscontainers.py
rhatdan: Does the following work?
```
 except util.FileNotFound as e:
         raise ValueError("Cannot install the container: runc is needed to run system containers") from e
```

giuseppe: thanks, I have tested that if the exception is caught then Python3 doesn't error out, I have updated and rebased the patches.

rhatdan:
@rh-atomic-bot r+

PR #530, opened 2016-08-10
[merged] Atomic/scan.py: Handle when default scan type is not valid

baude:
In the /etc/atomic.d/scanner file, if the default scan type is
not valid, we didn't handle the error and debugging was
confusing.  This makes a clear error message.


rhatdan:
@rh-atomic-bot r+

PR #531, opened 2016-08-10
[merged] verify: validate layers of system image stored on disk

stephen679:
Integrated the [gomtree filesystem verification tool](https://github.com/vbatts/go-mtree) into `atomic verify -V <system_image>`. 

The idea behind this integration of a filesystem verification tool is so that when a user pulls an image (for now only system images) using `atomic pull ...`, a manifest that details a layer (i.e, file ordering, file attributes like mode, size, sha256digest, etc.) will be created and stored in `/var/lib/atomic/gomtree-manifests`. Then, when a user performs `atomic verify -V <system_image>`, these manifests will be validated against the local image the user has on disk.

Ideally, we would be able to do this for the rest of the images as well, and implement it in a similar manner to `syscontainers._pull_image_to_ostree()`

Since `gomtree` is still being developed, we need to maintain updated versions of the `gomtree` binary. I wasn't sure if a `go get...`, `go install...` was something we wanted to do in the Makefile. For now, I created a local build of the `gomtree` tool and added it as part of this commit.

For testing, would the program be able to write to a `/var/lib/atomic` directory? Because that would be the location where these "gomtree manifests" would be stored. If you want to test it by hand, you can go into your ostree repo and change some content of an image you're not using (although this might affect other images as well, so be careful) and then perform `verify` with the -V flag.

Signed-off-by: Stephen Chung <schung@redhat.com>


stephen679:
@vbatts here is an integration of `gomtree` into the atomic CLI

Atomic/verify.py
```diff
@@ -343,6 +350,50 @@ def assemble_nvr(self, image, image_name=None):
...
+        missing_manifests = []
+        for layer in layers:
+            layer = layer.replace("sha256:","")
+            manifestpath = self.get_gomtree_manifest(layer)
+            if not manifestpath:
+                missing_manifests.append(layer)
+                continue
+            tmp_dir = tempfile.mkdtemp()
+            ref = os.path.join("ociimage", layer)
+            cmd = ['ostree','checkout','--union','--repo=%s' % "/ostree/repo",ref,tmp_dir]
```
stephen679: can we make `syscontainers._get_repo_location()` a non protected function?

Atomic/syscontainers.py
```diff
@@ -858,3 +866,38 @@ def _pull_dockertar_layers(self, repo, imagebranch, temp_dir, input_layers):
...
+        it's layer. Defaults to placing the manifest in a folder in /var folder.
+        :param layer:
+        :param tar: may or may not be gzip compressed
+        :return: None
+        """
+        if not os.path.exists(root):
+            os.makedirs(root)
+        manifestname = os.path.join(root, "%s.mtree" % layer)
+        if os.path.isfile(manifestname):
+            util.write_out("validation manifest for layer %s alredy created" % layer)
```
giuseppe: small typo: already

Atomic/syscontainers.py
```diff
@@ -764,25 +765,32 @@ def _check_system_oci_image(self, repo, img, upgrade):
         manifest = self._skopeo_get_manifest(img)
         layers = SystemContainers.get_layers_from_manifest(manifest)
         missing_layers = []
+        missing_gomtree_manifests = []
         for i in layers:
             layer = i.replace("sha256:", "")
             if not repo.resolve_rev("%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer), True)[1]:
```
giuseppe: could we combine here the two conditions and not require `missing_gomtree_manifests`?
```
has_layer = repo.resolve_rev("%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer), True)[1]
has_gomtree_manifest = os.path.isfile(os.path.join(ATOMIC_VAR, "gomtree-manifests/%s.mtree" % layer)):
if not has_layer or not has_gomtree_manifest:
    missing_layers.append(layer)
    util.write_out("Missing layer %s" % layer)
```

giuseppe:
in `prune_ostree_images` we should check that for each manifest file the layer still exists.  If the layer doesn't exist, then remove the manifest file.

For images stored in an OSTree repository, how does this compare to `ostree fsck`?

Atomic/verify.py
giuseppe: sure

rhatdan:
@mitr @runcom PTAL

@mitr at some point we would like to get this to be covered by the signing spec.

Atomic/syscontainers.py
```diff
@@ -858,3 +866,38 @@ def _pull_dockertar_layers(self, repo, imagebranch, temp_dir, input_layers):
...
+            util.write_out("validation manifest for layer %s alredy created" % layer)
+        else:
+            gomtree = util.generate_validation_manifest(img_tar=tar,
+                                                        keywords="type,uid,gid,mode,size,sha256digest")
+            if gomtree.return_code != 0:
+                decompressed = tar.replace(".tar",".tgz")
+                os.rename(tar,tar.replace(".tar",".tgz"))
+                cmd = ['gunzip', decompressed]
+                d = util.subp(cmd)
+                if d.return_code == 0:
```
rhatdan: Could we use go bindings for gzip?
https://golang.org/pkg/compress/gzip/

Atomic/util.py
```diff
@@ -296,7 +297,6 @@ def skopeo_layers(image, args=None, layers=None):
     finally:
         if not success:
             shutil.rmtree(temp_dir)
-
     return temp_dir
```
rhatdan: Intentional?

Atomic/syscontainers.py
```diff
@@ -858,3 +866,38 @@ def _pull_dockertar_layers(self, repo, imagebranch, temp_dir, input_layers):
...
+                    gomtree = util.generate_validation_manifest(img_tar=tar,
+                                                                keywords="type,uid,gid,mode,size,sha256digest")
+                    util.subp(['gzip',decompressed])
+                    os.rename(tar, tar.replace(".tar.gz",".tar"))
```
rhatdan: Use python bindings for gzip, don't exec unless you have to.

Atomic/verify.py
```diff
@@ -343,6 +350,50 @@ def assemble_nvr(self, image, image_name=None):
...
+            r = util.subp(cmd)
+            if r.return_code != 0:
+                util.write_out(r.stderr)
+                continue
+            r = util.validate_manifest(manifestpath, img_rootfs=tmp_dir,keywords="type,uid,gid,mode,size,sha256digest")
+            util.write_out("validation output for layer %s:\n" % layer)
+            if r.return_code == 0:
+                util.write_out("\t(no changes detected)\n")
+            else:
```
rhatdan: Don't output on success.  If the tool does not find anything then dont' write output.

Atomic/verify.py
```diff
@@ -343,6 +350,50 @@ def assemble_nvr(self, image, image_name=None):
...
+                util.write_out(r.stdout)
+            if r.return_code > 1:
+                util.write_out(r.stderr)
+            shutil.rmtree(tmp_dir)
+        if len(missing_manifests):
+            util.write_out("validation manifests for the following layers do not exist:\n")
+            for layer in missing_manifests:
```
rhatdan: Should this go to stderr?

Makefile
```diff
@@ -15,7 +15,6 @@ VERSION=$(shell $(PYTHON) setup.py --version)
 all: python-build docs pylint-check dockertar-sha256-helper
 
 .PHONY: test-python3-pylint
-test-python3-pylint: 
 	$(PYTHON3_PYLINT) --disable=all --enable=E --enable=W --additional-builtins=_ *.py atomic Atomic tests/unit/*.py -d=no-absolute-import,print-statement,no-absolute-import,bad-builtin
```
rhatdan: Why?

atomic
```diff
@@ -683,6 +683,9 @@ def create_parser(help_text):
     verifyp.add_argument("-v", "--verbose", default=False,
                           action="store_true",
                           help=_("Report status of each layer"))
+    verifyp.add_argument("-V", "--validate", default=False,
+                                action="store_true",
+                                help=_("validate the state of an image"))
     return parser
```
rhatdan: Should this default to true?

docs/atomic-verify.1.md
```diff
@@ -13,7 +13,9 @@ IMAGE
 # DESCRIPTION
 **atomic verify** checks whether there is a newer image available and scans
 through all layers to see if any of the layers, which are base images themselves, have a new version available.
-If the tool finds an out of date image, it will report as such.
+If the tool finds an out of date image, it will report as such. **atomic verify** also allows for
+the state of a system image stored on disk to be validated, in that if the image has changed locally,
+details about those changes will be reported.
```
rhatdan: I think we want to switch this definition to include all of the things that validate will attempt to do.  Do them by default and then add options to disable features.

rhatdan:
@baude PTAL

Makefile
stephen679: This is a mistake, probably due to rebasing. I'll fix it

atomic
stephen679: If we want the default behavior to do this then yes. Since there was other functionality that `verify` does, I thought doing the system image validation would just be an option for `verify`

rhatdan: Right, I just think this might be more important functionality then what we had before.  I would like to see verify grow to 
* verify the image is the lates,
* verify the images is signed accoding to policy
* verify the image has not been tampered with (This patch)

Atomic/verify.py
stephen679: it should, yeah. I'll change it to `write_err`

stephen679:
@giuseppe `ostree fsck` checks for the overall integrity of the image by evaluating only the checksum. If we wanted more details about what exactly about that image has changed, then integrating `gomtree` would help us do that. Also, being able to have a validation manifest allows us to incorporate this 'consistency check' when we distribute and sign images.

stephen679:
pushed revisions based off feedback. reomved the -V flag, and make the system validation default behavior.

stephen679:
revised this PR so that now on an `atomic verify -g <non-system image>`, we can create a validation manifest. Then, when we run `atomic verify <non-system image>`, the validation manifest will be checked against the docker image (through temporarily mounting the image)

Atomic/syscontainers.py
```diff
@@ -31,6 +32,7 @@
     DEVNULL = open(os.devnull, 'wb')
 
 ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
+ATOMIC_VAR = '/var/lib/atomic'
```
rhatdan: Should this be under /var/lib/containers?  As used below?

Makefile
```diff
@@ -15,7 +15,7 @@ VERSION=$(shell $(PYTHON) setup.py --version)
 all: python-build docs pylint-check dockertar-sha256-helper
 
 .PHONY: test-python3-pylint
-test-python3-pylint: 
+-test-python3-pylint:
```
rhatdan: What is this doing?

rhatdan:
We need some tests to make sure this works.

Atomic/syscontainers.py
stephen679: ok I will change this to `/var/lib/containers/atomic`

giuseppe:
when are these files deleted?  Shouldn't the manifest be deleted on prune if the layer is not present anymore?

stephen679:
added a simple test case to check if `atomic pull` on a system image will create gomtree manifests. will need to figure out how to change a consistent file in a system image so that we can test the output of gomtree validation

rhatdan:
Ok we can merge this, but @lsm5 is going to have to add gomtree to the atomic build, or we need to get this accepted as a new package into Fedora.

@rh-atomic-bot r+

atomic
```diff
@@ -683,6 +682,12 @@ def create_parser(help_text):
     verifyp.add_argument("-v", "--verbose", default=False,
                           action="store_true",
                           help=_("Report status of each layer"))
+    verifyp.add_argument("-V", "--validate", default=True,
+                                action="store_false",
+                                help=_("disable validating system images"))
+    verifyp.add_argument("-g", "--generate", default=False,
```
cgwalters: It seems weird to have this option under `verify`.  I'd expect it to be a system-wide setting one can turn on by default.

Also...I'm a bit concerned about the performance of all of this.  Our intent with OSTree BTW is to drive immutability down into the storage stack - then you don't really need verify since the files are truly immutable.  See discussion in [http://www.spinics.net/lists/linux-fsdevel/msg75085.html](O_OBJECT).

rhatdan: Maybe we should move this to another verb
atomic images generate
Which would generate new checksums.

rhatdan: The immutable bit, could be a problem where a user could not remove a file that a hacker had placed on a file system.

cgwalters: That sort of thing is why I'm not using the immutable bit in ostree today.  The semantics are annoying.  My proposal for O_OBJECT *does* allow root to unlink the file (and to create new links to it).  I.e. the inode can change, but not the contents.

cgwalters: Also, the immutable bit can be removed.  With O_OBJECT at least you'd have to have compromised the kernel (or have CAP_SYS_ADMIN to write to the raw disk).

In the bigger picture of course if someone has `CAP_SYS_ADMIN` on your machine, they can inject code into the kernel to lie to whatever userspace validation tools you have, or things like [inject malware into the hard drive controller](http://hackaday.com/2015/06/08/hard-drive-rootkit-is-frighteningly-persistent/).

That said, if we did have the mtree data in OCI images we built, one could use them to more efficiently perform offline forensics too.  Mount a disk with libguestfs or whatever, download just the mtree metadata from a layer rather than the whole layer tar.

PR #533, opened 2016-08-11
[merged] Fix wrong filtered variable in test_ps

yuqi-zhang:
"type" is not a filterable variable. "runtime" is what the test is meaning to filter for. Perhaps in the future, "type" should be supported as an alternative for "runtime", but for now this is a quick fix.

On another note, if the user doesn't actually have docker/atomic containers set up, this test will always pass (since both atomic ps and docker ps gives the same output).


giuseppe:
@rh-atomic-bot r+

PR #534, opened 2016-08-15
[merged] Move generate from atomic verify to atomic images

rhatdan:
Also breakout images handling into a separate python file.

Atomic/images.py

I have switched atomic images generate to generate an images
mtree file for each image in the system.


rhatdan:
This patch cleans up some of the problems in https://github.com/projectatomic/atomic/pull/531/

giuseppe:
@rh-atomic-bot r+ 345faad

PR #535, opened 2016-08-15
[merged]  Fix up many failures when running atomic with overlay backend

rhatdan:
Overlay fields are being returned from docker as Capatalized

LowerDir, UpperDir, WorkDir

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -495,9 +495,9 @@ class DockerMount(Mount):
 
         ld, ud, wd = '', '', ''
         try:
-            ld = cinfo['GraphDriver']['Data']['lowerDir']
-            ud = cinfo['GraphDriver']['Data']['upperDir']
-            wd = cinfo['GraphDriver']['Data']['workDir']
+            ld = cinfo['GraphDriver']['Data']['LowerDir']
+            ud = cinfo['GraphDriver']['Data']['UpperDir']
+            wd = cinfo['GraphDriver']['Data']['WorkDir']
         except KeyError:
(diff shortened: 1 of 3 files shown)


rhatdan:
@baude PTAL

Atomic/mount.py
```diff
@@ -636,6 +643,9 @@ def _clean_temp_container_by_path(self, path):
             self.d.remove_container(short_cid)
         self._clean_tmp_image()
 
+    def get_driver(self):
```
baude: Is this function needed?

Atomic/mount.py
```diff
@@ -617,7 +617,14 @@ def _get_overlay_mount_cid(self):
...
             raise MountError('The device mounted at that location is not a '
                              'docker container.')
-        return cdir.replace('/var/lib/docker/overlay/', '')
+
+        for c in self._get_all_cids():
+            graph = self.d.inspect_container(c)["GraphDriver"]
+            if graph['Data']['UpperDir'].startswith(cdir):
+                return c
+
+        raise MountError('The device mounted at that location is not a '
```
baude: Should "that location" be cdir or something specific?

Atomic/mount.py
rhatdan: I was using it but figured an easier way.  But I thought it was a worth while function to add.

Atomic/mount.py
rhatdan: Fixed the error message.

baude:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #536, opened 2016-08-16
[merged] Syscontainers disable gomtree by default

giuseppe:


rhatdan:
Why do you want to disable the mtree by default rather then allow users to disable it?

giuseppe:
I am not yet convinced that this should be the default for images stored in OSTree.  OSTree already checksums every file stored and have the possibility to verify them (ostree fsck).

rhatdan:
Ok, then could we make atomic verify use this by default then?

giuseppe:
yes, I think we could do that programmatically.  AFAIK, it is not is not possible to do it directly from the CLI for a single branch/layer, `ostree fsck` verifies each object in the repository.

Adding Colin to the discussion @cgwalters

cgwalters:
Yeah, ostree could learn how to do that.

giuseppe:
I've changed what this PR does.  I've added the validation of an OCI layer via OSTree.  There are a few more fixes and improvements for system containers.

rhatdan:
LGTM
@cgwalters PTAL

Atomic/syscontainers.py
```diff
@@ -218,13 +218,37 @@ def _resolve_remote_path(self, remote_path):
...
+            return self._do_checkout_system_container(repo, name, img, upgrade, values, destination, unitfileout, extract_only, remote)
+        except (ValueError, OSError) as e:
+            try:
+                shutil.rmtree(destination)
+            except OSError:
+                pass
+            try:
+                if not upgrade:
+                    shutil.rmtree(unitfileout)
+            except OSError:
```
cgwalters: Might be worth hoisting this out as `util.rmtree_nofail()` or something.

Atomic/syscontainers.py
```diff
@@ -922,34 +919,44 @@ def _pull_dockertar_layers(self, repo, imagebranch, temp_dir, input_layers):
...
+    def validate_layer(self, layer):
+        ret = []
+        layer = layer.replace("sha256:", "")
+        repo = self._get_ostree_repo()
+        if not repo:
+            return ret
+
+        def validate_ostree_file(csum):
+            _, inputfile, file_info, xattrs = repo.load_file(csum)
+            # images are imported from layer tarballs, without any xattr.  Don't use xattr to compute
```
cgwalters: Well...it's messier than that.  The problem is that SELinux will assign a label if we don't give it one, so what we really need to do is somehow keep track of whether or not the imported object was labeled (i.e. is it a system-ostree or a container-ostree).  But we still want to do dedup between them...

Also we have to consider whether or not we do `security.capability` or not.

Atomic/syscontainers.py
```diff
@@ -922,34 +919,44 @@ def _pull_dockertar_layers(self, repo, imagebranch, temp_dir, input_layers):
...
+            return OSTree.checksum_from_bytes(checksum_v)
+
+        def traverse(it):
+            while True:
+                res = it.next()  # pylint: disable=next-method-called
+                if res == OSTree.RepoCommitIterResult.DIR:
+                    dir_checksum = it.get_dir().out_content_checksum
+                    dir_it = OSTree.RepoCommitTraverseIter()
+                    dirtree = repo.load_variant(OSTree.ObjectType.DIR_TREE, dir_checksum)
+                    dir_it.init_dirtree(repo, dirtree[1], OSTree.RepoCommitTraverseFlags.REPO_COMMIT_TRAVERSE_FLAG_NONE)
```
cgwalters: We aren't validating the dirtree object here, but that's OK.  Just a note.

Atomic/syscontainers.py
```diff
@@ -922,34 +919,44 @@ def _pull_dockertar_layers(self, repo, imagebranch, temp_dir, input_layers):
...
+
+        def validate_ostree_file(csum):
+            _, inputfile, file_info, xattrs = repo.load_file(csum)
+            # images are imported from layer tarballs, without any xattr.  Don't use xattr to compute
+            # the OSTree object checksum.
+            xattrs = GLib.Variant("a(ayay)", [])
+            _, checksum_v = OSTree.checksum_file_from_input(file_info, xattrs, inputfile, OSTree.ObjectType.FILE)
+            return OSTree.checksum_from_bytes(checksum_v)
+
+        def traverse(it):
```
cgwalters: I think this code is all OK, but we should consider pushing it down into OSTree at some point.

cgwalters:
@rh-atomic-bot r+ 5c26acb

PR #537, opened 2016-08-16
[merged] Additional fixes to make atomic images generate work.

rhatdan:
Need to ignore <none> images.
Need to change systemcontainers _checkout_system_container to take a remote
flag, rather then expect it in the self.args structure.  Since this
function is being called in mount, there is no guarantee that self.args.remote
exists.


rhatdan:
@giuseppe PTAL

giuseppe:
should the system containers be ignored completely by `generate` or how are the layers tarball accessed? I have a related PR, to disable ithe manifest generation by default on pull:

https://github.com/projectatomic/atomic/pull/536

rhatdan:
@giuseppe I would like to have them generated mtree also.

giuseppe:
from what I see in the code, it generates the manifest for each layer, and we have the layers tarballs only during a pull, as the tarballs are deleted once imported into OSTree.  We should add the support to mount a single imported oci layer first if we want to support manifest generation in a second moment, in the same way as it is done for Docker images.

Personally I would just skip gomtree manifest generation for system containers until these things are sorted out.

rhatdan:
Sounds good to me.

giuseppe:
Atomic/images.py looks fine to me.  Can the changes in Atomic/syscontainers.py be reverted now?

rhatdan:
I actually like the changes to systemcontainers but I guess they are less needed for this patch.

Atomic/syscontainers.py
```diff
@@ -172,7 +172,7 @@ def install_system_container(self, image, name):
             util.write_out("%s already present" % (name))
             return
 
-        return self._checkout_system_container(repo, name, image, 0, False)
+        return self._checkout_system_container(repo, name, image, 0, False, self.args.remote)
```
giuseppe: must be `remote=self.args.remote`, otherwise it gets assigned to `values`.  I have not verified, but this should be enough to fix the failing test.

There is another call to `_checkout_system_container` in the `update_system_container` function that needs the new argument `remote=self.args.remote` as well

rhatdan: Fixed the first one, but the update_system_container was just setting the remote=None so since that is the default, I don't think we need to change it.

Atomic/syscontainers.py
```diff
@@ -847,7 +846,6 @@ def extract_system_container(self, img, destination):
         repo = self._get_ostree_repo()
         if not repo:
             return False
-        self.args.remote = None
         return self._checkout_system_container(repo, img, img, 0, False, destination=destination, extract_only=True)
```
giuseppe: the `_checkout_system_container` call here misses the argument `remote=self.args.remote`.

Everything else LGTM

rhatdan: Since you were hard coding it to None and the default is None, then we don't need to pass it.

giuseppe: the function `update_system_container` sets `self.args.remote` before calling `_checkout_system_container`, this information is read from the `info` file.  In this way, if the container was installed with another rootfs, then we use the same rootfs for updating the container.

rhatdan:
```
-        self.args.remote = None
         return self._checkout_system_container(repo, img, img, 0, False, destination=destination, extract_only=True)
```
Since you are setting remote to None `self.args.remote = None`, I don't see how this is any different?

giuseppe: sorry I placed the comment on the wrong line (how can I comment on a line that is not part of the patch?), the change must be done in the `update_system_container` function.  At line 448 we read the value of self.arg.remote which now is not honored in the call to `_checkout_system_container` at line 467.

The change remote=self.args.remote should go at the line 448

rhatdan: Ok I added this, PTAL

giuseppe:
@rh-atomic-bot r+ d720f4d

PR #538, opened 2016-08-16
Add missing PyYAML for requiremets.txt so setup.py will work

rhatdan:


ashcrow:
@rhatdan good news, it seems to satisfy deps.

I do see another error though that is related to dependencies.

```
$ python3 setup.py develop                                          
  File "setup.py", line 5, in <module>
    import Atomic as _Atomic
  File "/tmp/atomic/Atomic/__init__.py", line 3, in <module>
    from .atomic import Atomic
  File "/tmp/atomic/Atomic/atomic.py", line 8, in <module>
    from .syscontainers import SystemContainers
  File "/tmp/atomic/Atomic/syscontainers.py", line 21, in <module>
    gi.require_version('OSTree', '1.0')
AttributeError: module 'gi' has no attribute 'require_version'
$
$
$ grep -r 'require_ver' /tmp/t/lib/python3.5/site-packages/gi/
$
$ pip freeze | grep gi=
gi==1.2
$
```

I **think** ``gi`` in requirements may be selecting [this library](https://pypi.python.org/pypi/gi/) instead of what is expected as the ``__init__`` contains:

```
GIST_API_URL = 'https://api.github.com/gists'
```

miabbott:
@ashcrow I believe `gi` in this case is actually `pygobject`

ashcrow:
@miabbott That makes a lot more sense. Unfortunately, ``gi`` in PyPi is not the ``gi`` that Atomic is looking for.

rhatdan:
Could this be gobject-base?

# rpm -q --requires atomic | grep python
/usr/bin/python3
python(abi) = 3.5
python3-PyYAML
python3-dateutil
python3-dbus
python3-docker-py >= 1.7.2
python3-gobject-base
python3-requests >= 2.4.3
python3-setuptools
python3-six >= 1.3.0
python3-websocket-client >= 0.11.0
rpm-python3

ashcrow:
@rhatdan I think you are correct but it is not available in pypi AFAICT. The closest I found was a pretty old [PyGObject](https://pypi.python.org/pypi/PyGObject/2.28.3) library. :frowning:.

I think removing ``gi`` and adding ``PyYaml`` is the best that can be done without removing the use of the expected ``gi`` and replacing it with something in pypi. With that said, ```python setup.py develop``` and ```pip install -r requirements.txt`` will not be fully usable on their own and will require manual installation for the extra libraries when installing from source.

cgwalters:
This was covered before:
https://github.com/projectatomic/atomic/pull/113
https://github.com/projectatomic/atomic/pull/180

atomic depends on native C code as well as Docker for that matter, it's not a pure Python program and hence one needs to use other tools (e.g. Docker, `yum-builddep` etc. that support hybrid models).

Having things in `requirements.txt` is useful though for tools that know how to parse it and inject it into multi-language build/distribution systems like RPM spec files, bitbake recipes etc.  It will just never be exhaustive and we can't support `python setup.py develop`.

rhatdan:
Ok I will just merge this patch then.

PR #539, opened 2016-08-17
[merged] Introduce signing

baude:
Allow atomic users to sign an images that has been pulled
from a repository.  This uses GPG, skopeo, and friends and
creates a local signature file for a image that has been
pulled locally.  Signatures are stored in the dir:

/etc/pki/containers/registry/image_name@sha256:image_id/

Individual signature files are then stored as:

  ../signature-(INTEGER)

where INTEGER is incremented each time a signature
is added.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -66,6 +66,7 @@ class Atomic(object):
         self.useTTY = True
         self.syscontainers = SystemContainers()
         self.run_opts = None
+        self.atomic_config = util.get_atomic_config()
 
     def __enter__(self):
         return self
--- a/Atomic/sign.py
+++ b/Atomic/sign.py
@@ -1,0 +1,129 @@
+from . import util
+from . import Atomic
+import os
+import tempfile
+from .atomic import AtomicError
+import re
+
+
+ATOMIC_CONFIG = util.get_atomic_config()
+
+def cli(subparser):
+    # atomic sign
+    signer = ATOMIC_CONFIG.get('default_signer', None)
+    signp = subparser.add_parser("sign",
+                                 help="Sign an image",
+                                 epilog="Create a signature for an image which can be "
+                                        "used later to verify it.")
+    signp.set_defaults(_class=Sign, func="sign")
+    signp.add_argument("images", nargs="*", help=_("images to sign"))
+    signp.add_argument("--signed-by", dest="sign_key", default=signer,
+                       help=_("Name of the signing key. Currently %s, "
+                              "default can be defined in /etc/atomic.conf" % signer))
+    signp.add_argument("-o", dest="output", default=None,
+                       help=_("The filename of the signature"))
+
+
+class Sign(Atomic):
+    def sign(self):
+        # TODO
+        # Atomic is run as sudo. Should we work around that?
+
+        if self.args.debug:
+            util.write_out(str(self.args))
+
+        signature_path = util.get_local_signature_path(self.atomic_config)
+
+        signer = self.args.sign_key
+
+        for sign_image in self.args.images:
+            remote_inspect_info = util.skopeo_inspect("docker://{}".format(sign_image))
+            manifest = util.skopeo_inspect('docker://{}'.format(sign_image), args=['--raw'], return_json=False)
+            try:
+                manifest_file = tempfile.NamedTemporaryFile(mode="wb", delete=False)
+                manifest_file.write(manifest)
+                manifest_file.close()
+                manifest_hash = str(util.skopeo_manifest_digest(manifest_file.name))
+
+                expanded_image_name = str(remote_inspect_info['Name'])
+                sigstore_path = "{}/{}/{}@{}".format(signature_path, os.path.dirname(expanded_image_name),
+                                                     os.path.basename(expanded_image_name), manifest_hash)
+                self.make_sig_dirs(sigstore_path)
+                sig_name = self.args.output if self.args.output is not None else self.get_sig_name(sigstore_path)
+                fq_sig_path = os.path.join(sigstore_path, sig_name)
+                if os.path.exists(fq_sig_path):
+                    raise ValueError("The signature {} already exists.  If you wish to "
+                                     "overwrite it, please delete this file first")
+
+                util.skopeo_standalone_sign(expanded_image_name, manifest_file.name,
+                                            self.get_fingerprint(signer), fq_sig_path)
+                util.write_out("Created: {}".format(fq_sig_path))
+
+            finally:
+                os.remove(manifest_file.name)
+
+    def check_input_validity(self):
+        try:
+            for image in self.args.images:
+                self._is_image(image)
+        except AtomicError:
+            raise ValueError("{} is not a valid image".format(image))
+
+    @staticmethod
+    def get_fingerprint(signer):
+        cmd = ['gpg', '--no-permission-warning', '--with-colons', '--fingerprint', signer]
+        return_code, stdout, stderr = util.subp(cmd, newline=True)
+        if return_code is not 0:
+            raise ValueError(stderr)
+        for line in stdout.splitlines():
+            if line.startswith('fpr:'):
+                return line.split(":")[9]
+
+    @staticmethod
+    def make_sig_dirs(sig_path):
+        if not os.path.exists(sig_path):
+            # TODO
+            # perhaps revisit directory permissions
+            # when complete use-cases are known
+            os.makedirs(sig_path, '0777')
+
+    @staticmethod
+    def get_sig_name2(sig_path):
+        def missing_ints(aoi):
+            # Returns a list of integers in range
+            start, end = 1, max(aoi) + 1
+            if start == end and start is not 1:
+                start = 1
+            _diff = sorted(set(range(start, end)).difference(aoi))
+            if len(_diff) == 0:
+                return end
+            else:
+                return min(_diff)
+
+        sigs = []
+        for sig in os.listdir(sig_path):
+            if re.match(r"signature-\b[0-9]+\b(?!\.[0-9])", sig):
+                sigs.append(int(sig.replace("signature-", "")))
+
+        sigs.sort()
+        if len(sigs) == 0:
+            return "signature-1"
+        # In the event signature-0 exists
+        if sigs[0] == 0:
+            del sigs[0]
+        missing = missing_ints(sigs)
+        if missing == 0:
+            sig_int = max(sigs) + 1
+        else:
+            sig_int = missing
+        return "signature-{}".format(sig_int)
+
+    @staticmethod
+    def get_sig_name(sig_path):
+        sig_files = set(os.listdir(sig_path))
+        sig_int = 1
+        while True:
+            name = "signature-{}".format(sig_int)
+            if name not in sig_files:
+                return name
+            sig_int += 1
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -98,12 +98,13 @@ def image_by_name(img_name, images=None):
     return valid_images
 
 
-def subp(cmd, cwd=None):
+def subp(cmd, cwd=None, newline=False):
     # Run a command as a subprocess.
     # Return a triple of return code, standard out, standard err.
     proc = subprocess.Popen(cmd, cwd=cwd,
                             stdout=subprocess.PIPE,
-                            stderr=subprocess.PIPE, close_fds=True)
+                            stderr=subprocess.PIPE, close_fds=True,
+                            universal_newlines=newline)
     out, err = proc.communicate()
     return ReturnTuple(proc.returncode, stdout=out, stderr=err)
 
@@ -246,7 +247,7 @@ def urllib3_disable_warnings():
             if hasattr(urllib3, 'disable_warnings'):
                 urllib3.disable_warnings()
 
-def skopeo_inspect(image, args=None):
+def skopeo_inspect(image, args=None, return_json=True, newline=False):
     if not args:
         args=[]
 
@@ -257,7 +258,7 @@ def skopeo_inspect(image, args=None):
 
     cmd = ['skopeo', 'inspect'] + args + [image]
     try:
-        results = subp(cmd)
+        results = subp(cmd, newline=newline)
     except OSError:
         raise ValueError("skopeo must be installed to perform remote inspections")
     if results.return_code is not 0:
@@ -265,7 +266,10 @@ def skopeo_inspect(image, args=None):
         check_v1_registry(image)
         raise ValueError("Unable to interact with this registry: {}".format(results.stderr))
     else:
-        return json.loads(results.stdout.decode('utf-8'))
+        if return_json:
+            return json.loads(results.stdout.decode('utf-8'))
+        else:
+            return results.stdout
 
 
 def skopeo_delete(image, args=None):
@@ -318,6 +322,23 @@ def skopeo_layers(image, args=None, layers=None):
             shutil.rmtree(temp_dir)
     return temp_dir
 
+def skopeo_standalone_sign(image, manifest_file_name, fingerprint, signature_path):
+    cmd = ['skopeo', 'standalone-sign', manifest_file_name,
+                       image, fingerprint, "-o", signature_path]
+    try:
+        results = subp(cmd)
+    except Exception as e: # pylint: disable=broad-except
+        raise ValueError(e)
+    if results.return_code is not 0:
+        raise ValueError(results.stderr)
+
+def skopeo_manifest_digest(manifest_file):
+    cmd = ['skopeo', 'manifest-digest', manifest_file]
+    try:
+        results = subp(cmd)
+    except Exception: #pylint: disable=broad-except
+        raise ValueError(results.stderr)
+    return results.stdout.rstrip().decode()
 
 def check_v1_registry(image):
     # Skopeo cannot interact with a v1 registry
@@ -354,7 +375,12 @@ def get_atomic_config_item(config_items, atomic_config=None):
     Lookup and return the atomic configuration file value
     for a given structure. Returns None if the option
     cannot be found.
+
+    ** config_items must be a list!
     """
+
+    assert isinstance(config_items, list)
+
     def _recursive_get(atomic_config, items):
         yaml_struct = atomic_config
         try:
@@ -498,3 +524,11 @@ def validate_manifest(spec, img_rootfs=None, img_tar=None, keywords=""):
     if keywords:
         cmd += ['-k',keywords]
     return subp(cmd)
+
+def get_local_signature_path(atomic_conf):
+    # If signature path is defined, get it; else return default
+    signature_path = get_atomic_config_item(['default-sigstore-path'], atomic_config=atomic_conf)
+    if signature_path is None:
+        signature_path = '/var/lib/atomic/sigstore'
+    return signature_path
+
--- a/atomic
+++ b/atomic
@@ -47,6 +47,7 @@ from Atomic import push
 from Atomic import stop
 from Atomic import uninstall
 from Atomic import update
+from Atomic import sign
 from Atomic.util import write_err, NoDockerDaemon
 import traceback
 from Atomic.mount import MountError
@@ -144,6 +145,7 @@ def create_parser(help_text):
     pull.cli(subparser)
     run.cli(subparser)
     scan.cli(subparser)
+    sign.cli(subparser)
     stop.cli(subparser)
     storage.cli(subparser)
     top.cli(subparser)
--- a/docs/atomic-sign.1.md
+++ b/docs/atomic-sign.1.md
@@ -1,0 +1,55 @@
+% ATOMIC(1) Atomic Man Pages
+% Brent Baude
+% August 2016
+# NAME
+atomic-sign- Create a signature for an image
+
+**WARNING**
+
+Only use **atomic sign** if you trust the remote registry which contains the image
+(preferably by being the only administrator of it).
+
+
+# SYNOPSIS
+**atomic sign**
+[**-h**|**--help**]
+
+[**-o**, **--output**]
+[**--sign_key**]
+[ image ... ]
+
+# DESCRIPTION
+**atomic sign** will create a local signature for one or more local images that have 
+been pulled from a registry. Unless overridden, the signature will end up in the 
+the default storage location (/var/lib/atomic/containers) for signatures.  A different
+default location can be defined in /etc/atomic.conf with the key **default-sigstore-path**.
+
+# OPTIONS
+**-h** **--help**
+  Print usage statement.
+
+**-o** **--output**
+  Assign a specific signature file name; otherwise, the file name is generated.  
+ 
+
+**--signed-by**
+  Override the default identity of the signature. You can define a default in /etc/atomic.conf
+  with the key **default_signer**.
+
+
+# EXAMPLES
+Sign the foobar image from privateregistry.example.com
+
+    atomic sign privateregistry.example.com/foobar
+    
+Sign the foobar image with a specific signature name.
+
+    atomic sign -o foobar.sig privateregistry.example.com
+
+Sign the busybox image with the identify of foo@bar.com
+
+    atomic --signed-by foo@bar.com privateregistry.example.com
+
+
+# HISTORY
+Initial revision by Brent Baude (bbaude at redhat dot com) August 2016


rhatdan:
Please resync your pull .

Atomic/atomic.py
```diff
@@ -99,6 +99,7 @@ class Atomic(object):
     results = '/var/lib/atomic'
     skull = (u"\u2620").encode('utf-8')
 
+
```
rhatdan: Remove

Atomic/atomic.py
```diff
@@ -1142,6 +1144,32 @@ def get_vulnerable_ids(self):
         except IOError:
             return []
 
+
+#    def sign(self):
```
rhatdan: Can we put this in sign.py.

Atomic/atomic.py
```diff
@@ -1142,6 +1144,32 @@ def get_vulnerable_ids(self):
...
+#
+#        # Make sure the signature path exists
+#        if not os.path.exists(signature_path):
+#            raise ValueError("The path signature path {} "
+#                             "does not exist".format(signature_path))
+#
+#        for sign_image in self.args.sign_targets:
+#            print(util.skopeo_inspect('docker://{}'.format(sign_image),args=['--raw']))
+#
+#        sys.exit()
```
rhatdan: Don't think you want this here.

docs/sign.md
```diff
@@ -0,0 +1,43 @@
+% ATOMIC(1) Atomic Man Pages
+% Brent Baude
+% August 2016
+# NAME
+atomic-sign- Create a signature for a local image
+# SYNOPSIS
+**atomic diff**
```
rhatdan: atomic sign

Atomic/atomic.py
baude: Fixed, left over turds.

docs/sign.md
baude: fixed.

Atomic/atomic.py
baude: fixed, leftovers

Atomic/client.py
```diff
@@ -110,3 +112,11 @@ def iter_subs(tree, key=None):
         for i in range(len(tree)):
             tree[i] = iter_subs(tree[i])
     return tree
+
+
+def ids_same(id1, id2):
+    if id1 == id2:
+        return True
+    if no_shaw(id1) == no_shaw(id2):
```
mtrmac: Why is this discarding the algorithm identifier? At a first glance I can’t see any reason for this, and it could allow attacks using a weaker algorithm against a stronger one.

Atomic/sign.py
```diff
@@ -0,0 +1,89 @@
...
+import os
+import tempfile
+from .atomic import AtomicError
+from .client import ids_same, add_shaw
+
+
+class Sign(Atomic):
+    def sign(self):
+        # TODO
+        # If an alternate sign key is provided, should we verify that it is valid?
```
mtrmac: Valid by what measure?

Atomic/sign.py
```diff
@@ -0,0 +1,89 @@
...
+        # Should we check to see if there are valid GPG signatures?
```
mtrmac: Adding more signatures should never hurt. Adding a completely redundant one might merit a warning, sure.

Atomic/util.py
```diff
@@ -472,3 +486,11 @@ def validate_manifest(spec, img_rootfs=None, img_tar=None, keywords=""):
     if keywords:
         cmd += ['-k',keywords]
     return subp(cmd)
+
+def get_signature_path(atomic_conf):
+    # If signature path is defined, get it; else return default
+    signature_path = get_atomic_config_item(['default-sigstore-path'], atomic_config=atomic_conf)
+    if signature_path is None:
+        signature_path = '/etc/pki/containers/'
+    return signature_path
```
mtrmac: We need to make this consistent with https://github.com/mtrmac/image/blob/docker-lookaside/docker/lookaside.go#L102 : 

 - Using URLs
 - `"registries".""` instead of `default-sigstore-path` (though that might be by changing `containers/image` instead)
 - `sigstore-write` values
 - No `/etc/pki/containers` default I think… A single sigstore for all of the internet really isn’t likely to be useful. For any external public repositories, there would be a read-only sigstore via HTTP. Read-write would be only the locally created/managed repositories (`mydepartment.mycompany.com/registry`) or something.

Atomic/sign.py
```diff
@@ -0,0 +1,89 @@
...
+        signer = util.get_atomic_config_item(['default_signer']) if \
+            self.args.sign_key is None else self.args.sign_key
+
+        # Make sure the signature path exists
+        if not os.path.exists(signature_path):
+            raise ValueError("The path signature path {} "
+                             "does not exist".format(signature_path))
+
+        for sign_image in self.args.sign_targets:
+            local_inspect_info = self._inspect_image(image=sign_image)
```
mtrmac: This seems to be looking for a locally downloaded image. Why?

Atomic/sign.py
```diff
@@ -0,0 +1,89 @@
...
+            manifest = util.skopeo_inspect('docker://{}'.format(sign_image), args=['--raw'])
+
+            # Make sure the local image is the same as the image on the
+            # registry
+
+            if not ids_same(sign_image_id, manifest['config']['digest']):
+                raise ValueError("The local image {} differs from the "
+                                 "version on the registry".format(sign_image))
+            with tempfile.NamedTemporaryFile(mode="w", delete=False) as manifest_file:
+                manifest_file.write(str(manifest))
```
mtrmac: This causes the manifest to be JSON-decoded and JSON-reencoded, potentially modifying it. It needs to use the original raw bytes without any modificaiton.

Atomic/util.py
mtrmac: … and the per-registry/per-namespace/per-repository lookup from `lookaside.go` is necessary as well; we need various ISVs who publish to Docker hub to be able to maintain their own sigstore servers.

Atomic/sign.py
```diff
@@ -0,0 +1,89 @@
...
+                reg, image, tag = util.decompose(self.get_fq_name(local_inspect_info)) # pylint: disable=unused-variable
```
mtrmac: Pylint can be silenced in a more readable way (and being more explicit about which variable is intentionally unused) by naming the variable `_` or `unused_`$anything IIRC.

mtrmac: This doesn’t work AFAICS; local inspect gives you _one_ of the local tags of that image, with no guarantee that it will be the right one if there are several.

AFAICS this should not be requiring a local image at all, and only use the identity from the command line = used by the repository.  Yes, it will need some canonicalization to the effect of https://github.com/docker/docker/blob/master/reference/reference.go#L192 (`normalize`) and https://github.com/docker/docker/blob/master/reference/reference.go#L140 (`WithDefaultTag`). I’m not sure whether there already is a Python library doing this.

Atomic/sign.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+                if self.args.output is None:
+                    sigstore_path = os.path.join(signature_path,
+                                                 "{}/{}@{}".format(reg,
+                                                                   image,
+                                                                   add_shaw(sign_image_id)))
```
mtrmac: This is the `config.digest` from the manifest? That won’t match, it must be a digest of the manifest itself.

mtrmac: `./skopeo manifest-digest manifest.json`, per the WIP email.

Atomic/sign.py
```diff
@@ -0,0 +1,89 @@
...
+                    self.make_sig_dirs(sigstore_path)
+                    signature_int = self.get_sig_int(sigstore_path)
+                    fq_sig_path = os.path.join(sigstore_path, "signature-{}".format(signature_int))
+
+                else:
+                    fq_sig_path = os.path.join(self.args.output)
+                    self.make_sig_dirs(os.path.dirname(fq_sig_path))
```
mtrmac: I don’t think it’s customary to automatically create directories if the user explicitly specifies a path (OTOH I really don't care).

Atomic/sign.py
```diff
@@ -0,0 +1,89 @@
...
+                return line.split(":")[9]
+
+    @staticmethod
+    def make_sig_dirs(sig_path):
+        if not os.path.exists(sig_path):
+            os.makedirs(sig_path)
+
+    @staticmethod
+    def get_sig_int(sig_path):
+        return len(os.listdir(sig_path)) + 1
```
mtrmac: This feels brittle. An aborted operation might leave high-numbered signature files around, which are ignored by https://github.com/mtrmac/image/blob/docker-lookaside/docker/docker_image_dest.go (which stops at the lowest unused number). Or there might be any random files in there (OS X Finder metadata, whatever).

https://github.com/mtrmac/image/blob/docker-lookaside/docker/docker_image_dest.go enumerates 1, 2, … and stops at first nonexistent file. In there it is especially motivated by HTTP support, where there isn’t a universally easily supported way to list the directory anyway, AFAIK.

Atomic/client.py
baude: In atomic, the image/container ids are exposed without the sha256 pre-pended.  There was confusion about what sha sum was being used in the local filesystem.  We thought it was the image ID but if it is the manifest, this is no longer needed.

Atomic/sign.py
baude: Only that the path provided is actually a key.  In looking at this later today, I dont think this is really a concern.  The error message gpg gives is quite clear and I simply expose that from stderr.

Atomic/sign.py
baude: One of the working assumptions for creating a local signature was that the image would also have to be local.

Atomic/sign.py
baude: Yeah I could have gone both ways on that.  Given you saw it, I'm with you.  Ill error out instead.

Atomic/sign.py
```diff
@@ -0,0 +1,89 @@
...
+                                                                   add_shaw(sign_image_id)))
+                    self.make_sig_dirs(sigstore_path)
+                    signature_int = self.get_sig_int(sigstore_path)
+                    fq_sig_path = os.path.join(sigstore_path, "signature-{}".format(signature_int))
+
+                else:
+                    fq_sig_path = os.path.join(self.args.output)
+                    self.make_sig_dirs(os.path.dirname(fq_sig_path))
+
+                util.skopeo_standalone_sign(sign_image, manifest_file.name,
```
mtrmac: Does `sign_image` here really work, always, when verifying against a policy? AFAICS it would be necessary to do the equivalent of `reference.WithDefaultTag`.

Atomic/sign.py
mtrmac: What is the purpose of that?

atomic
```diff
@@ -513,6 +514,16 @@ def create_parser(help_text):
                      help=_("Only display container IDs"))
     add_opt(pss)
 
+    # atomic sign
+    signp = subparser.add_parser("sign", help="Sign an image",
+                                 epilog="Create a signature for an image which can be used later to verify it.")
+    signp.set_defaults(_class=Sign, func="sign")
+    signp.add_argument("sign_targets", nargs="*", help=_("images to sign"))
```
rhatdan: Why not images?

atomic
```diff
@@ -513,6 +514,16 @@ def create_parser(help_text):
...
+    signp.add_argument("--sign_key", dest="sign_key", default=None,
```
rhatdan: Should default be read out of the config file?

atomic
```diff
@@ -513,6 +514,16 @@ def create_parser(help_text):
...
+                       help=_("Name of the signing key"))
```
rhatdan: Maybe mention that the default can be set in /etc/atomic.conf or wherever it is set.

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,43 @@
...
+**-h** **--help**
+  Print usage statement.
+
+**-o** **--output**
+  Used to override the default signature location.  It must be a fully qualified 
+  path name (including file name) for where to store the signature.
+
+**--sign_key**
+  Override the default signature and use the defined signature instead.
+
```
rhatdan: Where is the default signature defined?

rhatdan: Really need a couple of examples here.

atomic
baude: Ok changed.

atomic
baude: I dont think any other sub function does that.  Do you want this one to?

atomic
baude: Done

docs/atomic-sign.1.md
baude: Added location information and information around defaults.  Added two additional signing examples.

rhatdan:
If you had that it would eliminate this code, and the --help could show you the default.  I think we should do this for all keys that are defined in /etc/atomic.conf,  I think it would make it more user friendly.
        signer = util.get_atomic_config_item(['default_signer']) if \
+            self.args.sign_key is None else self.args.sign_key

baude:
@rhatdan Is this more what you had in mind?

atomic
```diff
@@ -514,6 +518,18 @@ def create_parser(help_text):
                      help=_("Only display container IDs"))
     add_opt(pss)
 
+    # atomic sign
+    signer = None if ATOMIC_CONFIG['default_signer'] is None else ATOMIC_CONFIG['default_signer']
```
rhatdan: Isn't this the same as 

signer = ATOMIC_CONFIG['default_signer']

atomic
```diff
@@ -514,6 +518,18 @@ def create_parser(help_text):
...
+    sig_store = "/var/lib/atomic/containers" if ATOMIC_CONFIG['default-sigstore-path'] is None else ATOMIC_CONFIG['default-sigstore-path']
+    signp = subparser.add_parser("sign", help="Sign an image",
+                                 epilog="Create a signature for an image which can be used later to verify it.")
+    signp.set_defaults(_class=Sign, func="sign")
+    signp.add_argument("images", nargs="*", help=_("images to sign"))
+    signp.add_argument("--sign_key", dest="sign_key", default=signer,
+                       help=_("Name of the signing key. A default can be defined in /etc/atomic.conf"))
```
rhatdan: help=_("Name of the signing key. Currently %s, default can be defined in /etc/atomic.conf" % signer))

rhatdan:
@baude close, I added a couple of comments.  Looking better, but now we need to get @mtrmac and @aweiteka to review it.

baude:
@mtrmac please review at your convenience

atomic
baude: Fixed.

atomic
baude: Nope but changed to:

signer = ATOMIC_CONFIG.get('default_signer', None)

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,55 @@
...
+The default storage location for signatures is /var/lib/atomic/containers.  A different
+default location can be defined in /etc/atomic.conf with the key **default-sigstore-path**.
+
+# OPTIONS
+**-h** **--help**
+  Print usage statement.
+
+**-o** **--output**
+  Used to override the default signature location.  It must be a fully qualified 
+  path name (including file name) for where to store the signature.
```
rhatdan: +**-o** **--output**
+  File path for storing the signature, overriding the default signature location, /var/lib/atomic/containers.

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,55 @@
...
+
+**--sign_key**
+  Override the default signature and use the defined signature instead. You can define
```
rhatdan: Should be signer not signature in both places.

baude: Fixed.

docs/atomic-sign.1.md
baude: Fixed

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+import os
+import tempfile
+from .atomic import AtomicError
+import re
+
+
+class Sign(Atomic):
+    def sign(self):
+        # TODO
+        # Atomic is run as sudo. Should we work around that?
```
mtrmac: I don’t understand what this comment is trying to say. Is `atomic(1)` _always_ run through sudo? (Probably not.) Does `atomic sign` always need to be run through sudo? (That would be surprising to me.)

Assuming “as sudo” means “as root”, this comment should say why is that needed so that it is clearer what to work around.

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+
+        if self.args.debug:
+            util.write_out(str(self.args))
+
+        # Verify all the images provided are valid
+        self.check_input_validity()
```
mtrmac: This still seems to be validating against local images; why should we require any the remote images to be present locally?

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+
+        signature_path = util.get_signature_path(self.atomic_config)
+        use_def_sig_path = True if signature_path == self.args.output else False
```
mtrmac: This is confusing. The sigpath mechanism (creating a directory structure based on reference) and the `-o` mechanism (writing to the named file) use the input string in a completely different way to arrive at the final output location.

`-o /var/lib/atomic/containers` should not behave as if the option is not there (it should probably fail with “trying to overwrite a directory with a file”).

AFAICS this variable should be removed, and its user replaced by simple `if self.args.output is None` (which would, in turn require removing the default value in the argument parser.).

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+
+        signer = self.args.sign_key
+
+        for sign_image in self.args.images:
+            remote_inspect_info = util.skopeo_inspect("docker://" + sign_image)
+            manifest = util.skopeo_inspect('docker://{}'.format(sign_image), args=['--raw'], return_json=False)
```
mtrmac: One line uses `+`, the line immediately below uses `'{}'.format`. That doesn’t really _hurt_ but having both consistent would make it clearer that nothing is going on here.

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+
+            # Make sure the local image is the same as the image on the
+            # registry
```
mtrmac: AFAICS nothing is doing that, nor should it.

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+            manifest_file = tempfile.NamedTemporaryFile(mode="wb", delete=False)
```
mtrmac: Use `try:`/`finally:` to minimize the risk of leaving files around?

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+            manifest_file.write(manifest)
+            manifest_file.close()
+            manifest_hash = str(util.skopeo_manifest_digest(manifest_file.name))
+
+            if use_def_sig_path:
+                _image_path = str(remote_inspect_info['Name'])
```
mtrmac: This would still benefit from a dedicated “format as full reference” helper in skopeo.

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+                image_name = os.path.basename(_image_path)
+
+                _sig_path = [signature_path, os.path.dirname(_image_path), "{}@{}".format(image_name, manifest_hash)]
+                sigstore_path = str(os.path.join(*_sig_path))
```
mtrmac: If this is supposed to add support for Windows paths, `_image_path` must be split not only to dir/base, but into all components; and the splitting should be done using the `'/'` character, not using OS-specific `os.path`.

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+                start = 1
+            _diff = sorted(set(range(start, end)).difference(aoi))
+            if len(_diff) == 0:
+                return end
+            else:
+                return min(_diff)
+
+        sigs = []
+        for sig in os.listdir(sig_path):
+            if re.match(r"signature-\b[0-9]+\b(?!\.[0-9])", sig):
```
mtrmac: What is this supposed to match? This seems like an attempt to anchor the end of the string, but what is that `\.` doing in there at all?

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+    def make_sig_dirs(sig_path):
+        if not os.path.exists(sig_path):
+            os.makedirs(sig_path)
+
+    @staticmethod
+    def get_sig_name(sig_path):
+        def missing_ints(aoi):
+            # Returns a list of integers in range
+            start, end = 1, max(aoi) + 1
+            if start == end and start is not 1:
```
mtrmac: `start` is always 1 per the line above.

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+                start = 1
+            _diff = sorted(set(range(start, end)).difference(aoi))
+            if len(_diff) == 0:
```
mtrmac: This can never be true because `end==max(aoi)+1` is not a member of `aoi`.

mtrmac: Sorry, `end` is exclusive. Never mind.

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+                sigs.append(int(sig.replace("signature-", "")))
+
+        sigs.sort()
+        if len(sigs) == 0:
+            return "signature-1"
+        # In the event signature-0 exists
+        if sigs[0] == 0:
+            del sigs[0]
+        missing = missing_ints(sigs)
+        if missing == 0:
```
mtrmac: AFAICS this can never happen.

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+            if line.startswith('fpr:'):
+                return line.split(":")[9]
+
+    @staticmethod
+    def make_sig_dirs(sig_path):
+        if not os.path.exists(sig_path):
+            os.makedirs(sig_path)
+
+    @staticmethod
+    def get_sig_name(sig_path):
```
mtrmac: All of this seems to be an awfully confusing way to say something like (untested!)

```python
existing_files = set(os.listdir(sig_path))
i = 1
while True:
    name = "signature-{}".format(i)
    if name not in existing_files:
        return name
    i += 1
```

(I didn’t spend time to prove exact equivalence, but a simple loop counting up from 1 is what `containers/image` does. Simply formatting numbers and looking up strings in a hash table should be simpler than using regexps to parse file names—and the above hard-codes the file name formatting convention in exactly one place.)

Atomic/util.py
```diff
@@ -241,15 +242,18 @@ def skopeo_inspect(image, args=None):
...
     if results.return_code is not 0:
         # Need to check if we are dealing with a v1 registry
         check_v1_registry(image)
         raise ValueError("Unable to interact with this registry: {}".format(results.stderr))
     else:
-        return json.loads(results.stdout.decode('utf-8'))
+        if return_json:
+            return json.loads(results.stdout.decode('utf-8'))
+        else:
+            return results.stdout
```
mtrmac: A single function returning two so wildly different types?  If anybody ever tried to do static analysis of Python, this would throw if off completely.

shrug* not my codebase, but personally I’d prefer a separate `skopeo_inspect_raw` function (which could in turn hardcode much more). Or, perhaps, just inline the `util.subp` call into `sign.py`.

Atomic/sign.py
```diff
@@ -0,0 +1,107 @@
...
+                self.make_sig_dirs(sigstore_path)
+                fq_sig_path = os.path.join(sigstore_path, self.get_sig_name(sigstore_path))
+
+            else:
+                fq_sig_path = self.args.output
+                if not os.path.exists(os.path.dirname(fq_sig_path)):
+                    raise ValueError("The path {} does not exist.".format(
+                        os.path.dirname(fq_sig_path)))
+
+            util.skopeo_standalone_sign(sign_image, manifest_file.name,
```
mtrmac: `_image_path` as computed above would be acceptable here; that adds a default tag if missing. (It also reformats to the fully explicit format, but AFAICS that doesn't matter. In any case, that would eventually benefit from a test that a signature created using the fully expanded name is accepted when pulling using the minimized name, and vice versa.)

Atomic/util.py
```diff
@@ -474,3 +500,11 @@ def validate_manifest(spec, img_rootfs=None, img_tar=None, keywords=""):
     if keywords:
         cmd += ['-k',keywords]
     return subp(cmd)
+
+def get_signature_path(atomic_conf):
```
mtrmac: Per https://github.com/projectatomic/atomic/pull/539#discussion_r75361262 :

---
We need to make this consistent with https://github.com/mtrmac/image/blob/docker-lookaside/docker/lookaside.go#L102 : 

 - Using URLs
 - `"registries".""` instead of `default-sigstore-path` (though that might be by changing `containers/image` instead)
 - `sigstore-write` values
 - No `/etc/pki/containers` default I think… A single sigstore for all of the internet really isn’t likely to be useful. For any external public repositories, there would be a read-only sigstore via HTTP. Read-write would be only the locally created/managed repositories (`mydepartment.mycompany.com/registry`) or something.

---
… and the per-registry/per-namespace/per-repository lookup from `lookaside.go` is necessary as well; we need various ISVs who publish to Docker hub to be able to maintain their own sigstore servers.

atomic
```diff
@@ -514,6 +518,18 @@ def create_parser(help_text):
...
 
+    # atomic sign
+    signer = ATOMIC_CONFIG.get('default_signer', None)
+    sig_store = ATOMIC_CONFIG.get('default-sigstore-path', "/var/lib/atomic/containers")
+    signp = subparser.add_parser("sign", help="Sign an image",
+                                 epilog="Create a signature for an image which can be used later to verify it.")
+    signp.set_defaults(_class=Sign, func="sign")
+    signp.add_argument("images", nargs="*", help=_("images to sign"))
+    signp.add_argument("--sign_key", dest="sign_key", default=signer,
+                       help=_("Name of the signing key. Currently %s, default can be defined in /etc/atomic.conf" % signer))
```
mtrmac: `--sign_key` reads a bit like “sign the specified key”. `--signing_key`, `--sign_by`?

Also, other options seem to use `-`, not `_`, to separate words.

atomic
```diff
@@ -514,6 +518,18 @@ def create_parser(help_text):
...
+    signp.add_argument("-o", "--output", dest="output", default=sig_store,
+                       help=_("Path to store the signature ({})".format(sig_store)))
```
mtrmac: See the discussion around users of this option, `default-sigstore-path` and `--output` are not doing the same thing.  (The option help could still mention that default, but using something like `(instead of creating a file within the directory structure under {})`—if it isn’t too long for `--help`.)

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,55 @@
+% ATOMIC(1) Atomic Man Pages
+% Brent Baude
+% August 2016
+# NAME
+atomic-sign- Create a signature for a local image
```
mtrmac: This isn’t touching the local image (other than currently verifying that an image with such a name exists, but that shouldn’t happen I think)

mtrmac: `Create a signature file for an image in a registry`?

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,55 @@
...
+[**-o**, **--output**]
+[**--sign_key**]
+[ image ... ]
+
+# DESCRIPTION
+**atomic sign** will create a local signature for one or more local images that have 
+been pulled from a registry. Unless overridden, the signature will end up in a directory
+as follows:
+
+ /var/lib/atomic/containers/registry_name/image_name@sha256:manifest/
```
mtrmac: `registry_name/image_name` does not describe the full hierarchy (Docker references can be arbitrarily namespaced), and I’m not sure that this implementation needs to be spelled out at all; instead, tell the user to publish the directory structure rooted at $top on the internet and give users configuration for consuming $public_top_url.

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,55 @@
...
+
+The signatures are named 'signature-INTEGER' where the integer is increased by 1 each
+time.
+
+The default storage location for signatures is /var/lib/atomic/containers.  A different
+default location can be defined in /etc/atomic.conf with the key **default-sigstore-path**.
```
mtrmac: This will need to be expanded for the various more strictly namespaced paths, per `lookaside.go` or similar.

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,55 @@
...
+
+# OPTIONS
+**-h** **--help**
+  Print usage statement.
+
+**-o** **--output**
+  File path for storing the signature, overriding the default signature
+  location, /var/lib/atomic/containers.
```
mtrmac: … and not creating the directory structure.

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,55 @@
...
+
+**--sign_key**
+  Override the default signer and use the defined signer instead. You can define
+  a default in /etc/atomic.conf with the key **default_signer**.
```
mtrmac: “signer” is, to my knowledge, not a common way to refer to a signing private key.

(And all variables, option names and config item names should use the same terminology; `sign_key` and `default_signer` doesn’t match.)

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,55 @@
...
+[**-h**|**--help**]
+
+[**-o**, **--output**]
+[**--sign_key**]
+[ image ... ]
+
+# DESCRIPTION
+**atomic sign** will create a local signature for one or more local images that have 
+been pulled from a registry. Unless overridden, the signature will end up in a directory
+as follows:
```
mtrmac: Somewhere near the top, this needs a

```md
# WARNING

Only use **atomic sign** if you trust the remote registry which contains the image
(preferably by being the only administrator of it);
if the registry serves you a malicious image, **atomic sign** will obliviously sign it.
```

docs/atomic-sign.1.md
```diff
@@ -0,0 +1,55 @@
...
+    atomic sign docker.io/busybox
+    
+Sign the busybox image and store the signature in /tmp
+
+    atomic sign -o /tmp/busybox.sig docker.io/busybox
+
+Sign the busybox image with the identify of foo@bar.com
+
+    atomic --sign-key foo@bar.com docker.io/busybox
+
```
mtrmac: Per the **WARNING** suggested above, all of this should use `privateregistry.example.com` instead of `docker.io`.

Atomic/sign.py
mtrmac: (And this is only validating that a tag with that name exists, the local and remote image may have nothing in common.)

Atomic/sign.py
mtrmac: …splitting of `_image_path`, that is.

Atomic/sign.py
rhatdan: Yes I think the comment should say if atomic is run as root, nothing to do with sudo.  If run as root we could try to figure out the homedir of the user running the command.  SUDO_USER and /proc/self/login_uid, are good indicators of the users UID.

docs/atomic-sign.1.md
rhatdan: Create a signature file for a container image?

rhatdan:
@baude sorry doing some shuffling of code around in atomic command, so this needs a rebase.

Atomic/sign.py
baude: @rhatdan would you be comfy tabling that part and circling back to it after the fact?

Atomic/sign.py
baude: Resolved.  This is leftover from the previous assumption that the image would be pulled locally first.  Good catch.

Atomic/sign.py
baude: This is something @rhatdan asked me to specifically to do.  It was the way you described previously.

aweiteka: I see the issue is the way signature_path is being calculated. Use a single var, `signature_path`. If self.args.output overrides default, then update signature_path. Default should be in one place (you have it defined in two places). I think that means get rid of use_def_sig_path boolean and calculate the path in one place.

Atomic/util.py
```diff
@@ -474,3 +500,11 @@ def validate_manifest(spec, img_rootfs=None, img_tar=None, keywords=""):
...
+    # If signature path is defined, get it; else return default
+    signature_path = get_atomic_config_item(['default-sigstore-path'], atomic_config=atomic_conf)
+    if signature_path is None:
+        signature_path = '/var/lib/atomic/containers'
```
aweiteka: This is redundant.

atomic
```diff
@@ -514,6 +518,18 @@ def create_parser(help_text):
                      help=_("Only display container IDs"))
     add_opt(pss)
 
+    # atomic sign
+    signer = ATOMIC_CONFIG.get('default_signer', None)
+    sig_store = ATOMIC_CONFIG.get('default-sigstore-path', "/var/lib/atomic/containers")
```
aweiteka: Here's the other default. Pick one.

Atomic/sign.py
mtrmac: No, `signature_path` is only the _base_ of a `signature_path/$reponame@$digest/$siquence`; the `-o` option does not override `signature_path` but the final file name, what the code calls `fq_sig_path`.

The location of the default should perhaps be in a utility function, sure, but that won’t resolve my concern that we are mixing strings with two different semantics, especially in this comparison.

(Of course one way to resolve this is to redefine `-o` to specify the base, not the final file.)

Atomic/util.py
mtrmac: (See above; I don’t think there even should be a default location to be used for the whole internet, and then the issue of redundancy between the signing code and the argument parsing + help disappears.)

Atomic/sign.py
baude: Done.

Atomic/sign.py
baude: yup, just leftover

Atomic/sign.py
baude: good idea, done

Atomic/sign.py
baude: agreed, broke this off into a sep issue https://github.com/projectatomic/atomic/issues/578

Atomic/sign.py
baude: I do not believe there is any requirement to support windows.

mtrmac: In that case I don’t understand why the `dirname`/`basename`/`join` sequence is needed, instead of simpler `"{}/{}@{}".format(signature_path, _image_path, manifest_hash)`.

Atomic/sign.py
baude: altered some variable names, _image_path now is expanded_image_name and we use that over sign_image.  Good point.

Atomic/sign.py
mtrmac: (Apart from mere aesthetics, using `os.path` on Docker references is semantically incorrect due to the `'/'` vs. `'\'` conventions.)

Atomic/sign.py
baude: it matches signature-[positive_integer]

Atomic/sign.py
baude: Perhaps a case of overthinking the problem. I like this.  I have added a slightly altered version of this and tested it.  Thanks!

Atomic/sign.py
baude: resolved by a different solution.

Atomic/sign.py
baude: resolved by a different solution.

Atomic/sign.py
baude: resolved by a different solution.

Atomic/sign.py
baude: resolved by a different solution.

Atomic/sign.py
baude: as you wish.

Atomic/util.py
baude: As atomic sign is for only creating a local signature, i have renamed this function to get_local_signature_path.  It is only used for figuring out the local path for storing signatures.  The points raised above do apply to the pull side of things.

atomic
baude: i liked --sign-by.

docs/atomic-sign.1.md
baude: fixed.

atomic
rhatdan: Use --sign-by

atomic
```diff
@@ -514,6 +518,18 @@ def create_parser(help_text):
...
+    signp = subparser.add_parser("sign", help="Sign an image",
```
rhatdan: Can you move all of this handling to sign.py under a function call cli, then only change to atomic is
sign.cli(subparser)

Just like we have done for all of the other CLI Commands.

Atomic/sign.py
rhatdan: Yes table it.

docs/atomic-sign.1.md
baude: done

docs/atomic-sign.1.md
baude: done

docs/atomic-sign.1.md
baude: good idea

atomic
baude: done

baude:
@rh-atomic-bot r+

PR #541, opened 2016-08-18
[merged] Add support for LABEL RUN_OPTS_FILE and ${RUN_OPTS}

adelton:
The main use is for containers that might want to determine some `docker run` option during `atomic install`, in their `install.sh`. For example, `/var/lib/${NAME}/docker-run-opts` can be populated with `--hostname=ipa.example.test`, making the subsequent `atomic run` invoke the `docker run` with container-specific option.


rhatdan:
bot, add author to whitelist

Atomic/run.py
```diff
@@ -38,6 +39,17 @@ def run(self):
...
                 args += self.command
+                opts_file = self._get_args("RUN_OPTS_FILE")
+                if opts_file:
+                    opts_file = self.sub_env_strings("".join(opts_file))
+                    if opts_file.startswith("/"):
+                        try:
+                            run_opts = open(opts_file, "r").read()
+                            os.environ["RUN_OPTS"] = run_opts
+                        except IOError:
+                            util.write_out("Failed to read RUN_OPTS_FILE %s" % opts_file)
```
rhatdan: Raise ValueError with these.  We don't want to continue to run the container if this fails.

Atomic/run.py
```diff
@@ -38,6 +39,17 @@ def run(self):
             args = self._get_args("RUN")
             if args:
                 args += self.command
+                opts_file = self._get_args("RUN_OPTS_FILE")
+                if opts_file:
+                    opts_file = self.sub_env_strings("".join(opts_file))
+                    if opts_file.startswith("/"):
+                        try:
+                            run_opts = open(opts_file, "r").read()
+                            os.environ["RUN_OPTS"] = run_opts
```
rhatdan: I would prefer to not modify the current environ.  We are moving away from this in another pull request.  Could we add this to self.run_opts and set it in cmd_env?

rhatdan: https://github.com/projectatomic/atomic/pull/521

Atomic/run.py
```diff
@@ -38,6 +39,17 @@ def run(self):
...
+                        except IOError:
+                            util.write_out("Failed to read RUN_OPTS_FILE %s" % opts_file)
+                    else:
+                        util.write_out("Failed to read RUN_OPTS_FILE %s: not absolute path" % opts_file)
```
rhatdan: Raise exception.

docs/atomic-run.1.md
```diff
@@ -49,6 +49,9 @@ These defaults are suggested values for your container images.
 **SUDO_GID**
   The `SUDO_GID` environment variable.  This is useful with the docker `-u` option for user space tools.  If the environment variable is not available, the default GID of the value for `SUDO_UID` is used.  If this value is not available, the value of `/proc/self/loginuid` is used.
 
+**RUN_OPTS**
```
rhatdan: How about we just call this OPTS_FILE and then we can use it for INSTALL, RUN, UNINSTALL. Any reason to have multiple?

Atomic/run.py
adelton: What if the file just does not exist -- do you want an error as well?

Atomic/run.py
adelton: Will try.

docs/atomic-run.1.md
adelton: OK, the LABEL can be OPTS_FILE. What should the "environment variable" in the `docker run` command instead of `${RUN_OPTS}` be called? `${OPTS}`? `${OPTS_FILE_CONTENT}`? Calling it just `${OPTS_FILE}` seems weird, as it won't be expanded to file name but to content of that file.

adelton: > How about we just call this OPTS_FILE and then we can use it for INSTALL, RUN, UNINSTALL. Any reason to have multiple?

Truth be told, I am able to imagine just the `RUN` use case, when in `INSTALL` you set the file to be consumed in `RUN`. I'm a bit afraid that using one file for all three (`INSTALL`, `RUN`, and `UNINSTALL`) might cause clashes because `RUN` might need some other options than `UNINSTALL`.

Do you have some examples of uses of this mechanism for `INSTALL` and `UNINSTALL`?

Or do you envision the `LABEL`s being different for each phase, just the "variable" where it is exposed being the same (`OPTS_FILE`)?

Atomic/run.py
rhatdan: If you don't think this is an error then don't state anything.  If we want to report a problem it should be a write_err() anyways.

docs/atomic-run.1.md
rhatdan: I am just thinking that once we have it for one option people will want it for others.
Lets keep it as RUN_OPTS_FILE and ignore the others for now, and we will create them if needed.

Atomic/run.py
adelton: > #521

The https://github.com/projectatomic/atomic/pull/521 as of ac83a85115cc35f467b8c35802c3be0517ab1877 seems to break the tests.

adelton: Bisecting the history, it was already 0cba782a03d110a154550fa5f1ad5a9e7cfacbd5 which introduced the regression.

adelton:
Take fc775bbdd02ab326a9d994159b6fa1e488025239 addresses all comments except for the `os.environ` one. I will update this commit one https://github.com/projectatomic/atomic/pull/521 is merged.

adelton:
Alternatively, if this PR looks otherwise fine, it could be merged and https://github.com/projectatomic/atomic/pull/521 could adapt the `os.environ["RUN_OPTS"]` changes in the future.

rhatdan:
@rh-atomic-bot r+

You need to update the containerlablels standards patch

adelton:
> You need to update the containerlablels standards patch

Where does that live?

rhatdan:
@adelton https://github.com/projectatomic/ContainerApplicationGenericLabels

adelton:
> @adelton https://github.com/projectatomic/ContainerApplicationGenericLabels

Thanks, file https://github.com/projectatomic/ContainerApplicationGenericLabels/pull/72.

PR #542, opened 2016-08-18
[merged] Atomic migrate fixes

shishir-a412ed:


shishir-a412ed:
ping @giuseppe 

/cc @rhatdan 

Shishir

tests/integration/test_migrate.sh
```diff
@@ -25,7 +25,12 @@ if [ "$init" != "systemd" ];then
     exit 0
 fi
 
-dockerPid=$(ps -C docker -o pid=|xargs)
+if ! systemctl is-active docker >/dev/null; then
+     echo "Docker daemon is not running"
+     exit 1
```
giuseppe: I think it is better to skip the test instead of failing.  We can skip it with "exit 77"

giuseppe:
the second commit should rm the existing binary blob for gotar, and add the rules to the Makefile to build it instead, this should be enough:

```
gotar: gotar.go
	$(GO) build -o $@ $<
```

and add 'gotar' to the `all` target as well

tests/integration/test_migrate.sh
shishir-a412ed: `test_migrate.sh` is already skipping the test. This code fixes a problem in capturing `docker PID` the right way, just in case someone decides to run the test.

shishir-a412ed: https://github.com/projectatomic/atomic/blob/master/tests/integration/test_migrate.sh#L5-L10

We can change the message if you like.

giuseppe: ok thanks.  I think the message is fine and can be left as it.

shishir-a412ed:
@giuseppe PTAL.

Also, I just tested `atomic migrate export` and `import` with these commits (2) and it's working fine.
I tested it with a few images, containers and local volumes.

Test environment: Fedora 23 with docker 1.13.0-dev

Shishir

giuseppe:
LGTM @rh-atomic-bot r+ d529da1

Issue #540, opened 2016-08-18
There seems no way to pass dollar sign to atomic run via --opt1


adelton:
When string with dollar sign is passed as value of `--opt1` to `atomic run`, the string disappears.

Consider Dockerfile

```
FROM fedora
LABEL RUN 'docker run --rm ${IMAGE} echo the value: ${OPT1}'
```

For "normal" values it works:

```
# atomic run --opt1=jezek test-double-sub_env_strings
docker run --rm test-double-sub_env_strings echo the value: jezek
the value: jezek
```

But for values with dollar sign it does not:

```
# atomic run --opt1='$jezek' test-double-sub_env_strings
docker run --rm test-double-sub_env_strings echo the value:
the value:
```

I suspect the reason is

```
        # Perform variable subs
        in_string = os.path.expandvars(in_string)

        # Replace undefined variables with blank
        in_string = re.sub(r'\$\{?\S*\}?', '', in_string)
```

in `sub_env_strings` -- the `re.sub` replaces not just undefined variables, but also results of previous successful `expandvars` which put the `$jezek` there instead of `${OPT1}`.

adelton:
On related note, I'm also concerned by a double expansion attempts caused by

```
        cmd = self.gen_cmd(args)
        cmd = self.sub_env_strings(cmd)
        self.display(cmd)
        if self.args.display:
            return

        if not self.args.quiet:
            self.check_args(cmd)
        if not self.args.display:
            util.check_call(self.sub_env_strings(cmd), env=self.cmd_env())
```

in https://github.com/projectatomic/atomic/blob/master/Atomic/run.py#L50. Note that there is
`cmd = self.sub_env_strings(cmd)`, followed by `util.check_call(self.sub_env_strings(cmd), ...` -- the `sub_env_strings` is called twice.

rhatdan:
@adelton Are we ok on this one?  Can I close it or do you think we still need to change?

adelton:
I'm not sure what the expected behaviour and handling of dollars should be.

I have atomic-1.13.1-1.el7.x86_64 and I see

```
# docker build -t issue-540 .
Sending build context to Docker daemon 2.048 kB
Step 1 : FROM fedora
 ---> 11a5107645d4
Step 2 : LABEL RUN 'docker run --rm ${IMAGE} echo the value: ${OPT1}'
 ---> Using cache
 ---> 084b61e41724
Successfully built 084b61e41724
# atomic run --opt1=jezek issue-540
[...]

PR #545, opened 2016-08-19
[merged] Don't call sub_env_strings twice

rhatdan:
Also remove useless check on display

### DIFF
--- a/Atomic/run.py
+++ b/Atomic/run.py
@@ -55,8 +55,7 @@ class Run(Atomic):
 
         if not self.args.quiet:
             self.check_args(cmd)
-        if not self.args.display:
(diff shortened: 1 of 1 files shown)


rhatdan:
This fixes https://github.com/projectatomic/atomic/issues/540

rhatdan:
@adelton PTAL

adelton:
Let me try to find some reproducer, so that we could also add tests for that.

adelton:
This change LGTM and useful but it does not really fix https://github.com/projectatomic/atomic/issues/540.

rhatdan:
@rh-atomic-bot r+

PR #546, opened 2016-08-19
[merged] Cleanup: No need for ifcheck

rhatdan:
driver_unmount_fn defaults to path=None


rhatdan:
@baude PTAL

giuseppe:
it looks like `_unsupported_backend` needs to accept `path`

giuseppe:
@rh-atomic-bot r+ b22471a

PR #547, opened 2016-08-19
[merged] build: Stop trying to build gomtree

cgwalters:
Doing things this way isn't "reproducible" - it won't work inside a
build system requires offline sources only, like current CAHC does.

gomtree needs to be packaged separately.


rhatdan:
@rh-atomic-bot r+

PR #548, opened 2016-08-20
[merged] images-list: Add --json

cgwalters:
When trying to script `atomic` via Ansible, in order to implement
idempotence we need the ability to introspect the current state
in a machine-readable way.

`ps` already has `--json`, so teach `images list` about it too;
the implementation is trivial.


rhatdan:
@rh-atomic-bot r+

PR #549, opened 2016-08-21
[merged] Add image container version to json output

giuseppe:
enhance the `ps --json` and `images list --json` output with the image id.

It allows to check what version of an image a container is using, and what is the latest version available (locally) of the image.

A couple of bug fixes are included as well.

@cgwalters

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -958,6 +958,7 @@ class Atomic(object):
             images = self.d.images(all=get_all)
             for i in images:
                 i["ImageType"] = "Docker"
+                i["ImageId"] = i["Id"]
         except requests.exceptions.ConnectionError:
             raise NoDockerDaemon()
         return images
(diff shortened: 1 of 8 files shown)


rhatdan:
Why is it referred to as ImageID in one place and Version in the other?

giuseppe:
I named it Version as there is already an ID in the output and I thought it was clearer to name it Version.  I can rename it if having both Id and ImageID makes no confusion

rhatdan:
Would probably be better to separate out a couple of these pull requests.

rhatdan:
I just think they should be consistent between images and ps command.

giuseppe:
I have updated it to have the same name "image_id" both for `images list --json` and `ps --json`

cgwalters:
@rh-atomic-bot r+ 643e50f

PR #551, opened 2016-08-22
syscontainers: Don't start on install

cgwalters:
Interactive users or deployment systems may want to perform
configuration before starting.  For example, I may want to configure
the flannel data in etcd before starting flannel.

This topic comes up a lot in the RPM/deb worlds too; most RPM
distributions are inconsistent about this, whereas Debian basically
never starts a service on package install.

I think also it's better to teach admins to use systemd for service
management directly.

(This commit also changes things so we only `enable` on initial install,
 which is a minor tweak)

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -407,11 +407,10 @@ class SystemContainers(object):
             os.unlink(sym)
         os.symlink(destination, sym)
 
-        self._systemctl_command("enable", name)
         if upgrade:
             self._systemctl_command("restart", name)
         else:
-            self._systemctl_command("start", name)
+            self._systemctl_command("enable", name)
 
     def _get_system_checkout_path(self):
         if os.environ.get("ATOMIC_OSTREE_CHECKOUT_PATH"):


giuseppe:
should we have a `--start-service` flag as well for install so that the container can be started without guessing the name?  Just asking here as part of the discussion, we can add this later.

rhatdan:
Why don't we tell them how to start the service on initial install.

```
         if upgrade:
             self._systemctl_command("restart", name)
         else:
            self._systemctl_command("enable", name)
            util.write_out("Enable this service by executing 'systemctl start %s'" % name)
```

giuseppe:
I was thinking of the non interactive case, and even this can be fixed assigning the name to the container.  Anyway, we can probably leave this for later, if it turns to be an issue.

yuqi-zhang:
There's 2 installs in the test_system_containers that directly check for the running service. Adding a "systemctl start ${NAME}" after them should fix the tests.

rhatdan:
@cgwalters Needs a rebase.

giuseppe:
I've rebased this change in https://github.com/projectatomic/atomic/pull/584 plus fixing the tests case.

rhatdan:
Since https://github.com/projectatomic/atomic/pull/584 has been merged can we close this?

PR #552, opened 2016-08-22
syscontainers: Suppress stderr from runc status attempt

cgwalters:
I noticed this when starting the flannel syscontainer failed, and I
got `open /run/runc/flannel/state.json: no such file or directory`
when running `atomic ps`.


cgwalters:
This duplicates https://github.com/projectatomic/atomic/pull/549

PR #553, opened 2016-08-22
[merged] Canononicalize system containers as type: `system`

cgwalters:
Currently for system containers, in `images list --json`, we have:
`"type": "System"`, and in `ps --json`, it shows up as `"type": "systemcontainer"`.

The extra `"container"` suffixing the type seems redundant, so let's
kill it, and canonicalize on lowercase, hence both cases simply become
`"system"`.

This matters because I am parsing the JSON in Ansible and consistency
is good.

### DIFF
--- a/Atomic/images.py
+++ b/Atomic/images.py
@@ -146,7 +146,7 @@ class Images(Atomic):
             if not image["RepoTags"]:
                 continue
             iid = image["RepoTags"][0]
-            if image["ImageType"] == "System":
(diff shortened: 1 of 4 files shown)


cgwalters:
(Note, not really tested a lot yet)

giuseppe:
@rh-atomic-bot r+ db778fa

rhatdan:
@rh-atomic-bot r+

PR #555, opened 2016-08-23
syscontainers: create the bind source path if it doesn't exist

giuseppe:
and if the user namespace is not used, chown the source bind path to
the current user.

This will allow us to bind only /var/lib/etcd/$NAME.etcd in the etcd
container, instead of /var/lib as it will be created automatically if
it doesn't exist.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -179,6 +179,20 @@ class SystemContainers(object):
             raise ValueError("Invalid configuration file.  Only readonly images are supported")
         if configuration['root']['path'] != 'rootfs' and not remote:
             raise ValueError("Invalid configuration file.  Path must be 'rootfs'")
+        # Ensure that the source path specified in bind/rbind exists and that the
+        # user owns it.
+        if "mounts" in configuration:
+            user_namespace = ("linux" in configuration and "namespaces" in configuration["linux"]
+                              and "user" in configuration["linux"]["namespaces"])
+            for mount in configuration["mounts"]:
+                if not "type" in mount:
+                    continue
+                if "source" in mount and "bind" in mount["type"]:
+                    source = mount["source"]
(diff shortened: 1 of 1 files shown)


cgwalters:
Hum...a lot going on here.  One approach is to require the container to install tmpfiles.d units.  Though then we'd need to track the installation of that file in addition to the systemd unit.  Also thinking...if systemd had `StateDirectory=` (like `RuntimeDirectory=`) that'd be exactly what we want. 

I've been thinking we should actually inject some metadata into the unit file to say that it's a syscontainer - it'd help us track the removal of files we installed.  That was something that always bothered me about the previous `atomic install` - we required users to track files they installed, unlike how rpm/dpkg etc. do it.

giuseppe:
Should the state directory be removed on uninstall?  I think it should not be removed by atomic, unless explicitly specified somehow.

I have used `/var/lib/etcd` for the etcd storage only because it was  already the location used by the etcd rpm and also the Docker etcd container.  Should we instead use something that facilitate the association storage->container, as `/var/lib/atomic/$NAME` or `/var/lib/containers-storage/$NAME`?

giuseppe:
this PR is related: https://github.com/projectatomic/atomic/pull/557

giuseppe:
I had a look at using systemd-tmpfiles to manage these directories, is it fine to assume `systemd-tmpfiles` is used only on `atomic install --user` and `atomic-uninstall` for user containers, while for system containers the conf file can be placed under '/etc/tmpfiles.d'?

Another thing to figure out is how it will work with `update --container`.  We should probably use stop + start the service instead of restart so we have time to cleanup and install the new version of the configuration file.

giuseppe:
I have implemented the system-tmpfiles integration here: https://github.com/projectatomic/atomic/pull/559

PR #556, opened 2016-08-23
[merged] tests: Start a libtest.sh like model for syscontainers test

cgwalters:
I'm trying to debug things here, and it's nicer to have more
info, like `libtest.sh` in ostree does.


giuseppe:
@rh-atomic-bot r+ 2e52e41

PR #557, opened 2016-08-25
[merged] syscontainers: define RUN_DIRECTORY and STATE_DIRECTORY

giuseppe:
instead of hardcoding /var/lib or /run in the config.json of the
containers, use these two variables.  They also change value when run as
not root user so to support both modes.

Differently from other defined variables, it is possible to override
these via --set.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -316,6 +316,14 @@ class SystemContainers(object):
         if extract_only:
             return
 
+        if self.user:
+            home = os.path.expanduser("~")
+            values["RUN_DIRECTORY"] = "/run/user/%s" % (os.getuid())
+            values["STATE_DIRECTORY"] = "%s/.data" % (home)
+        else:
+            values["RUN_DIRECTORY"] = "/run"
+            values["STATE_DIRECTORY"] = "/var/lib"
+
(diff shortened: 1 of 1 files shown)


Atomic/syscontainers.py
```diff
@@ -316,6 +316,14 @@ def _do_checkout_system_container(self, repo, name, img, upgrade, values, destin
         if extract_only:
             return
 
+        if self.user:
+            home = os.path.expanduser("~")
+            values["RUN_DIRECTORY"] = "/run/user/%s" % (os.getuid())
```
rhatdan: Should this check for 
XDG_RUNTIME_DIR=/run/user/3267
?

rhatdan:
Are you adding a patch where you will actually use these?

Atomic/syscontainers.py
giuseppe: yes, it should check that.  I will amend it.

giuseppe:
I would like to start using them in etcd and flannel, so that etcd can run both as `atomic install --system gscrivano/etcd` and `atomic install --user gscrivano/etcd` without any other change (I have not tested it through bwrap-oci yet though)

rhatdan:
@rh-atomic-bot r+

PR #558, opened 2016-08-25
[merged] Fix template error check

giuseppe:
fix the check for missing values to apply to the template file

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -156,6 +156,12 @@ class SystemContainers(object):
         if not repo:
             raise ValueError("Cannot find a configured OSTree repo")
 
+        if self.args.user and self.args.system:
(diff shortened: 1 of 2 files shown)


Atomic/syscontainers.py
```diff
@@ -156,6 +156,12 @@ def install_system_container(self, image, name):
         if not repo:
             raise ValueError("Cannot find a configured OSTree repo")
 
+        if self.args.user and self.args.system:
```
rhatdan: Can't you set these up as mutually exclusive in argparse?

giuseppe: cool, I didn't know it was possible, I have pushed a fixup patch

rhatdan:
@rh-atomic-bot r+

PR #559, opened 2016-08-26
[merged] Use systemd-tmpfiles to track files/directories created on the host

giuseppe:
se systemd-tmpfiles to track files/directories created on the host

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -426,6 +426,13 @@ class SystemContainers(object):
         else:
             tmpfiles_template = ""
 
+        # When upgrading, stop the service and remove previously installed
+        # tmpfiles, before restarting the service.
+        if upgrade:
+            self._systemctl_command("stop", name)
+            if (tmpfiles_template):
+                self._systemd_tmpfiles("--remove", tmpfilesout)
+
         _write_template(unitfile, systemd_template, values, unitfileout)
         if (tmpfiles_template):
             _write_template(unitfile, tmpfiles_template, values, tmpfilesout)
@@ -438,11 +445,9 @@ class SystemContainers(object):
         if (tmpfiles_template):
             self._systemd_tmpfiles("--create", tmpfilesout)
 
-        self._systemctl_command("enable", name)
-        if upgrade:
-            self._systemctl_command("restart", name)
-        else:
-            self._systemctl_command("start", name)
+        if not upgrade:
+            self._systemctl_command("enable", name)
+        self._systemctl_command("start", name)
 
     def _get_system_checkout_path(self):
         if os.environ.get("ATOMIC_OSTREE_CHECKOUT_PATH"):
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -32,10 +32,12 @@ except ImportError:
 
 ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
 ATOMIC_VAR = '/var/lib/containers/atomic'
-
+ATOMIC_VAR_USER = "%s/.containers/atomic" # Must be expanded
 OSTREE_OCIIMAGE_PREFIX = "ociimage/"
 SYSTEMD_UNIT_FILES_DEST = "/etc/systemd/system"
 SYSTEMD_UNIT_FILES_DEST_USER = "%s/.config/systemd/user" # Must be expanded
+SYSTEMD_TMPFILES_DEST = "/etc/tmpfiles.d"
+SYSTEMD_TMPFILES_DEST_USER = "%s/.containers/tmpfiles" # Must be expanded
 SYSTEMD_UNIT_FILE_DEFAULT_TEMPLATE = """
 [Unit]
 Description=$NAME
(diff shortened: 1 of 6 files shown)


Atomic/syscontainers.py
```diff
@@ -183,6 +185,19 @@ def _check_oci_configuration_file(self, conf_path, remote=None):
         if configuration['root']['path'] != 'rootfs' and not remote:
             raise ValueError("Invalid configuration file.  Path must be 'rootfs'")
 
+        missing_source_paths = []
+        # Ensure that the source path specified in bind/rbind exists and that the
+        # user owns it.
```
rhatdan: Where are you ensuring that the user owns it?

giuseppe: ops, sorry, wrong comment

Atomic/syscontainers.py
```diff
@@ -32,10 +32,12 @@
 
 ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
 ATOMIC_VAR = '/var/lib/containers/atomic'
-
+ATOMIC_VAR_USER = "%s/.containers/atomic" # Must be expanded
```
rhatdan: Why not
```
ATOMIC_VAR_USER = "%s/.containers/atomic" % os.path.expanduser("~")
```

Atomic/syscontainers.py
```diff
@@ -32,10 +32,12 @@
...
 OSTREE_OCIIMAGE_PREFIX = "ociimage/"
 SYSTEMD_UNIT_FILES_DEST = "/etc/systemd/system"
 SYSTEMD_UNIT_FILES_DEST_USER = "%s/.config/systemd/user" # Must be expanded
+SYSTEMD_TMPFILES_DEST = "/etc/tmpfiles.d"
+SYSTEMD_TMPFILES_DEST_USER = "%s/.containers/tmpfiles" # Must be expanded
```
rhatdan: Same as above.

Atomic/syscontainers.py
```diff
@@ -32,10 +32,12 @@
 
 ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
 ATOMIC_VAR = '/var/lib/containers/atomic'
-
+ATOMIC_VAR_USER = "%s/.containers/atomic" # Must be expanded
 OSTREE_OCIIMAGE_PREFIX = "ociimage/"
 SYSTEMD_UNIT_FILES_DEST = "/etc/systemd/system"
 SYSTEMD_UNIT_FILES_DEST_USER = "%s/.config/systemd/user" # Must be expanded
```
rhatdan: Same as above.

Atomic/syscontainers.py
giuseppe: in general I would prefer that code is not executed on an import, but this is small enough that I can fix in the patch and adjust the users of the variable

giuseppe:
as the change is related to already existing code which is not added by this PR, I have added a new patch to address the expansion of *_USER variables

rhatdan:
LGTM
@cgwalters PTAL

yuqi-zhang:
I tested a bit with etcd and this resolves the bind path issues, so we can probably drop https://github.com/projectatomic/atomic/pull/555 if we merge this.

giuseppe:
yes, this is an improved version of #555 so there would not be need of it once this is merged

giuseppe:
one issue that I am aware of is for --user containers, and how will `/var/run/$ID` be recreated, as the tmpfiles.conf file is processed only on install and uninstall.  Do you have any suggestions on how to handle this case?

giuseppe:
I have tried it, and `RuntimeDirectory=` in the systemd configuration file will be enough to cover this case, users should not probably need other temporary directories.

rhatdan:
Awesome.

giuseppe:
once this gets included, I will update the gscrivano/etcd and gscrivano/flannel containers to use this new feature

Atomic/syscontainers.py
```diff
@@ -183,6 +187,18 @@ def _check_oci_configuration_file(self, conf_path, remote=None):
         if configuration['root']['path'] != 'rootfs' and not remote:
             raise ValueError("Invalid configuration file.  Path must be 'rootfs'")
 
+        missing_source_paths = []
+        # Ensure that the source path specified in bind/rbind exists
+        if "mounts" in configuration:
+            for mount in configuration["mounts"]:
+                if not "type" in mount:
+                    continue
+                if "source" in mount and "bind" in mount["type"]:
```
rhatdan: how about:
```
if "type" in mount and "source" in mount and "bind" in mount["type"]:
```

giuseppe: I wanted to prepare it in case we will have to consider other types of mount.  A missing type should always be skipped.  I can add a fixup path if you prefer the shorter version

rhatdan: No this is fine

giuseppe:
does this look fine?

rhatdan:
Can you merge some of these together?

giuseppe:
I usually try to keep git patches as small as possible with a single logical change as it makes things much easier with "git blame", "git revert" if there is need, and more important with "git bisect" to find regressions.  With a big patch you will need to manually "sub bisect" it, as probably you won't spot immediately what change introduces the regression.  Anyway, if you prefer I can try to squash some of them together (e.g. "syscontainers: use daemon-reload when files are changed" with "syscontainers: during an upgrade block the service")

rhatdan:
I am looking at autogenerate and fixup autogenerate.  I don't mind you keeping them separate but a patch that modifies another patch should be merged.

cgwalters:
`fixup!` is special in git; see https://robots.thoughtbot.com/autosquashing-git-commits .  Homu will squash them.

rhatdan:
Ok thanks
@rh-atomic-bot r+

PR #560, opened 2016-08-26
[merged] pull: change output for pulled layers

yuqi-zhang:
Change "Missing layer" to "Pulling layer" to make it look less like
an error and more like an action during "atomic pull".


cgwalters:
@rh-atomic-bot r+ 58a1db7

PR #562, opened 2016-08-26
Pass additional args down to ostree commands

rhatdan:
Need this to fix
https://bugzilla.redhat.com/show_bug.cgi?id=1356539


cgwalters:
The bugzilla here was invalid :no_good: ...though this won't help anything right now AFAIK, it also doesn't seem to break anything.  So...I dunno.

rhatdan:
The current functions do not pass additional args to the ostree tools, which was the purpose of adding these args.  I believe it does fix that.  It does fix people putting the args in the wrong order

```
  args               Additional arguments appended to the upgrade method. Use
                     `-- --OPTION=VAL` if you want to pass additional
                     unsupported arguments to rpm-ostree.
```

rhatdan:
@cgwalters @giuseppe PTAL.  If this is not needed I can close this pull request.

rhatdan:
Doesn't seem like people want this so closing.

PR #565, opened 2016-08-27
[merged] Atomic/util.py Catch/Ignore errors in tempfiles /etc/atomic.d

baude:
When the util.get_scanners method was run, if there was a tempfile
or swa file existent in /etc/atomic.d/, atomic scan would throw
an unhandled exception.  We now handle the exception with a try/except
and print an error message but do not fail.

This should fix bug 1346921


rhatdan:
@rh-atomic-bot r+

PR #567, opened 2016-08-29
[merged] Check for gomtree and bwrap-oci before showing "images generate" and "install --user"

giuseppe:


Atomic/syscontainers.py
```diff
@@ -52,6 +52,12 @@
 BWRAP_OCI_PATH = "/usr/bin/bwrap-oci"
 RUNC_PATH = "/bin/runc"
 
+def system_containers_are_supported():
```
rhatdan: Lets put these into utils.
utils.runc_available()
utils.bwrap_oci_available()

giuseppe:
updated :arrow_up:  but the tests timed out

atomic
```diff
@@ -299,11 +299,14 @@ def create_parser(help_text):
     installp.add_argument("image", help=_("container image"))
     if Atomic.syscontainers.OSTREE_PRESENT:
         system_xor_user = installp.add_mutually_exclusive_group()
```
rhatdan: What happens if both are not available?  Does this work correctly and just ignore the exclusive_group?

giuseppe: ops, yes, it fails.  I have added a fixup commit to deal with this issue

rhatdan:
I will merge but do we have something in the man pages that explains why these fields are not there?

@rh-atomic-bot r+

giuseppe:
yes, I have added the documentation to the man page saying that `OSTree and (runc|bwrap-oci) are required for this feature to be available`

giuseppe:
@rh-atomic-bot retry

Issue #327, opened 2016-03-17
Bad usage for: atomic run --help 


jlebon:
```
# atomic run --help
usage: atomic run [-h] [-n NAME] [--spc] [--display] image ...

positional arguments:
  image                 container image
  command               command to execute within the container. If container
                        is not running, command is appendedto the image run
                        method

optional arguments:
  -h, --help            show this help message and exit
  -n NAME, --name NAME  name of container
  --spc                 use super privileged container mode: '/usr/bin/docker
                        run -t -i --rm --privileged -v /:/host -v /run:/run -v
                        /etc/localtime:/etc/localtime --net=host --ipc=host
                        --pid=host -e HOST=/host -e NAME=${NAME} -e
                        IMAGE=${IMAGE} ${IMAGE}'
  --display             preview the command that /bin/atomic would execute

atomic run defaults to the following command, if image does not specify LABEL
run '/usr/bin/docker create -t -i --name ${NAME} ${IMAGE}'
```

The positional arguments list `command` but the usage string doesn't have the word `command` anywhere. To someone unfamiliar with `atomic`, it's not obvious that `command` refers to the `...` at the end (in fact it makes it seem like `atomic run` can start multiple images at once).

I'd suggest something like:

```
usage: atomic run [-h] [-n NAME] [--spc] [--display] image [command] [arg...]

positional arguments:
  image                 container image
  command               command to execute within the container. If container
                        is not running, command is appendedto the image run
                        method
  arg                   arguments to pass to command
```

I'm guessing this is automatically output by argparse, so it will probably require reorganizing the run parser.

rhatdan:
Yes I would think this would require a code change, if possible.

rhatdan:
Fixed in atomic-1.13

PR #568, opened 2016-08-29
Added command as optional argument parameter for atomic run

Akasurde:
Fixes: https://github.com/projectatomic/atomic/issues/327

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>


rhatdan:
What is the difference here?  Does command end up having the wrong data if their is no command added?

rhatdan:
@Akasurde Ping?

PR #570, opened 2016-08-30
[merged] This series of pull requests moves the CLI option handling out of atomic into submodules.

rhatdan:
Simplify/shorten atomic and Atomic/atomic.py by moving each atomic `verb` to its own submodule.

PR #572, opened 2016-08-30
[merged] Add --fstrim to atomic ps command

shishir-a412ed:

### DIFF
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -1,9 +1,16 @@
 import json
+import os
 
 from . import util
 from . import Atomic
+from .client import AtomicDocker
 import datetime
 from dateutil.parser import parse as dateparse
+
+try:
+    from subprocess import DEVNULL  # pylint: disable=no-name-in-module
+except ImportError:
+    DEVNULL = open(os.devnull, 'wb')
 
 def cli(subparser):
     # atomic containers
@@ -11,6 +18,7 @@ def cli(subparser):
     containers_subparser = c.add_subparsers(title='images subcommands',
                                             description="operate on images",
                                             help='additional help')
+    # atomic containers list
     pss = containers_subparser.add_parser("list",
                                           help=_("list the containers"),
                                           epilog="By default this shows only the running containers.")
@@ -28,11 +36,24 @@ def cli(subparser):
                      help=_("Don't truncate output"))
     pss.add_argument("-q", "--quiet", action='store_true', dest="quiet", default=False,
                      help=_("Only display container IDs"))
-    util.add_opt(pss)
-
+    # atomic containers trim
+    trimp = containers_subparser.add_parser("trim",
+                                            help=_("discard unused blocks (fstrim) on running containers"),
+                                            epilog="Discard unused blocks (fstrim) on rootfs of running containers.")
+    trimp.set_defaults(_class=Containers, func='fstrim')
 
 class Containers(Atomic):
 
+    def fstrim(self):
+        with AtomicDocker() as client:
+            for container in client.containers():
+                containerId = container["Id"]
+                ret = self._inspect_container(name=containerId)
+                pid = ret["State"]["Pid"]
+                mp = "/proc/%d/root" % (pid)
+                util.write_out("Trimming container id {0}".format(containerId[0:12]))
+                util.check_call(["/usr/sbin/fstrim", "-v", mp], stdout=DEVNULL)
+        return
     def ps_tty(self):
         all_container_info = self.ps()
         all_containers = []
--- a/docs/atomic-containers.1.md
+++ b/docs/atomic-containers.1.md
@@ -14,14 +14,9 @@ atomic containers allows the user to view and operate on containers
 
 list containers on your system.
 
-# list OPTIONS
-[**-h|--help**]
-[**-a|--all**]
-[**-f|--filter**]
-[**--json**]
-[**-n|--noheading**]
-[**--no-trunc**]
-[**-q|--quiet**]
+**trim**
+
+discard unused blocks (fstrim) on running containers.
 
 # DESCRIPTION
 **atomic containers list**, by default, will list all running containers on your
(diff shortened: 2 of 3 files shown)


shishir-a412ed:
ping @rhvgoyal @rhatdan 

/cc @giuseppe 

Shishir

Atomic/ps.py
```diff
@@ -83,6 +96,17 @@ def ps_tty(self):
         if not all_containers:
             return
 
+        if self.args.fstrim:
+            for info in all_containers:
+                container = info["container"]
+                ret = self._inspect_container(name=container)
```
giuseppe: _inspect_container works only for Docker containers.  So this loop should skip any container where `info["type"] != "docker"`

Atomic/ps.py
```diff
@@ -24,6 +30,8 @@ def cli(subparser):
                      help=_("Don't truncate output"))
     pss.add_argument("-q", "--quiet", action='store_true', dest="quiet", default=False,
                      help=_("Only display container IDs"))
+    pss.add_argument("--fstrim", action='store_true', dest="fstrim", default=False,
```
giuseppe: this should be added to the documentation and bash autocompletion as well.  There is a `test_ps.sh` file that you could use to add a test as well.

Atomic/ps.py
```diff
@@ -33,6 +41,11 @@ def ps_tty(self):
         skull = (u"\u2620").encode('utf-8')
         all_container_info = self.ps()
         all_containers = []
+
+        if self.args.all and self.args.fstrim:
+            util.write_out("fstrim can only be used for running containers. Use --fstrim without -a/--all option.")
```
rhatdan: Can we make these mutually exclusive options in the argparser?

rhatdan:
I am not crazy about this being an option on atomic ps.  Might be better to add a new category.  

```
atomic containers trim
```

And then have something like
```
atoimc containers ps 
```
To mimic atomic images.

rhatdan:
Or better yet

```
atomic containers list
```

shishir-a412ed:
@rhatdan so you are saying remove `atomic ps` completely.
And we should have something like:

1) `atomic containers list` which would have all the options like `--all`, `--no-trunc`, `--quiet` which currently `atomic ps` has. 
2) `atomic containers fstrim` which would discard unused blocks on the container rootfs.

Shishir

Atomic/ps.py
shishir-a412ed: Fixed.

rhatdan:
We could maybe leave `atomic ps` as a deprecated interface and make it link to `atomic containers list`

@cgwalters @baude What do you think?

Adding additional commands to atomic ps that has nothing to do with listing processes makes no sense.

giuseppe:
we didn't keep it around for long enough that we can probably just rename it and not have to worry of a deprecated interface

rhatdan:
Ok @shishir-a412ed Please change pull request to move to atomic containers trim,  Then we can get a different pull request to change ps to atomic containers list.

shishir-a412ed:
@rhatdan Since `atomic containers trim` needs to list containers (ID's) and `fstrim` their rootfs one by one. It would be better to open PR for `atomic containers list` first. And once that is merged.
We can add the subcommand `trim` to `atomic containers`. 

Based on @giuseppe comment, no need to deprecate `ps` right? Just create (rename ps) a new command `atomic containers list`.

rhatdan:
Yes I am preparing an `atomic containers list` pull request.  Just waiting for my previous pull request to get merged.
https://github.com/projectatomic/atomic/pull/570

rhatdan:
@shishir-a412ed My atomic containers list patch has been merged, so lets get this one rebased on top of it.
`atomic containers trim`
@rhvgoyal WDYT

rhvgoyal:
@rhatdan `atomic containers trim` sounds good to me. It is simple and intuitive.

shishir-a412ed:
@giuseppe do we need `atomic containers trim` for both type of containers {system containers, docker containers} ?

giuseppe:
@shishir-a412ed no, we don't need fstrim for system containers

shishir-a412ed:
@giuseppe thanks for the quick response !

Shishir

shishir-a412ed:
@rhatdan @giuseppe PTAL.

Shishir

giuseppe:
LGTM

rhvgoyal:
What is "system container" and why don't we need trimming capabilities for those. Is it because of the fact that system containers run outside of docker and they don't use thin volume?

giuseppe:
@rhvgoyal exactly they don't use thin volumes.  Images are stored in OSTree and containers are exploded to the file system

docs/atomic-containers.1.md
```diff
@@ -29,10 +35,20 @@ system.
 
 Using --all will list all the installed containers.
 
+**atomic containers trim**, By default this will only discard unused blocks (fstrim)
```
rhatdan: Discard unused blocks (fstrim) on all running containers.

docs/atomic-containers.1.md
```diff
@@ -14,8 +14,14 @@ atomic containers allows the user to view and operate on containers
...
 
-# list OPTIONS
+**trim**
+
+discard unused blocks (fstrim) on running containers.
+
+# OPTIONS
 [**-h|--help**]
+
+# list OPTIONS
```
rhatdan: Discard this section, only use the list options section below.

Atomic/containers.py
```diff
@@ -29,10 +37,25 @@ def cli(subparser):
...
+    # atomic containers trim
+    trimp = containers_subparser.add_parser("trim",
+                                            help=_("discard unused blocks (fstrim) on running containers"),
+                                            epilog="By default this will only discard unused blocks (fstrim) on rootfs of running containers.")
+    trimp.set_defaults(_class=Containers, func='ps_fstrim')
+    util.add_opt(trimp)
 
 class Containers(Atomic):
 
+    def ps_fstrim(self):
```
rhatdan: Lets drop the ps_ and just call this fstrim

Atomic/containers.py
```diff
@@ -29,10 +37,25 @@ def cli(subparser):
     pss.add_argument("-q", "--quiet", action='store_true', dest="quiet", default=False,
                      help=_("Only display container IDs"))
     util.add_opt(pss)
-
+    # atomic containers trim
+    trimp = containers_subparser.add_parser("trim",
+                                            help=_("discard unused blocks (fstrim) on running containers"),
+                                            epilog="By default this will only discard unused blocks (fstrim) on rootfs of running containers.")
+    trimp.set_defaults(_class=Containers, func='ps_fstrim')
+    util.add_opt(trimp)
```
rhatdan: Don't want this

shishir-a412ed:
@rhatdan PTAL.

rhatdan:
@rh-atomic-bot r+

PR #573, opened 2016-08-31
[merged] add option to support automatically answer yes for all questions

chuanchang:
sometimes, I wanna delete importing directory automatically w/o interaction. 

### DIFF
--- a/Atomic/Import.py
+++ b/Atomic/Import.py
@@ -9,7 +9,7 @@ from . import util
 
 ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
 
-def import_docker(graph, import_location):
+def import_docker(graph, import_location, assumeyes):
     """
     This is a wrapper function for importing docker images, containers
     and volumes.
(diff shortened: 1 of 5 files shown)


rhatdan:
I would prefer that we do like dnf and make a flag be assume yes.

```       -y, --assumeyes
              Automatically answer yes for all questions
```

That way we can handle any future questions.

rhatdan:
bot, add author to whitelist

chuanchang:
@rhatdan , I simply added '--assumeyes' option into atomic import command, it had better we have a global class/function to manage command/command option like dnf, in order to implement this, we probably need to re-factoring codes, maybe, we can do this later.

rhatdan:
We could add --assumeyes at the level of -v, --version and --debug to the atomic command.  Then you could do 

atomic -y storage export ...

chuanchang:
@rhatdan, done, thanks for your review.

rhatdan:
bot, add author to whitelist

rhatdan:
Now I need updated man page and bash completion.  While you are fixing the man page can you add documentation for --debug and --version.  Thanks, then I can merge.

chuanchang:
@rhatdan , done.

rhatdan:
Awesome thanks.
@rh-atomic-bot r+

PR #579, opened 2016-09-02
[merged] Move atomic ps to atomic containers

rhatdan:
As we add more commands related to containers, IE Trim, we need to make
containers be a subcommand and add verbs like list, and trim


rhatdan:
@shishir-a412ed @giuseppe PTAL

giuseppe:
@rh-atomic-bot r+ aa01dbb

PR #583, opened 2016-09-03
[merged] Change options to the agreed syntax.

rhatdan:
We agreed on using --sign-by for specifying the signer, and --directory
to specify the output directory for images.

### DIFF
--- a/Atomic/sign.py
+++ b/Atomic/sign.py
@@ -11,18 +11,23 @@ ATOMIC_CONFIG = util.get_atomic_config()
 def cli(subparser):
     # atomic sign
     signer = ATOMIC_CONFIG.get('default_signer', None)
+    signature_path = util.get_atomic_config_item(['default-sigstore-path'], atomic_config=ATOMIC_CONFIG)
+    if signature_path is None:
+        signature_path = util.ATOMIC_VAR_LIB + '/sigstore'
+
     signp = subparser.add_parser("sign",
                                  help="Sign an image",
                                  epilog="Create a signature for an image which can be "
                                         "used later to verify it.")
     signp.set_defaults(_class=Sign, func="sign")
     signp.add_argument("images", nargs="*", help=_("images to sign"))
-    signp.add_argument("--signed-by", dest="sign_key", default=signer,
+    signp.add_argument("--sign-by", dest="sign_by", default=signer,
                        help=_("Name of the signing key. Currently %s, "
                               "default can be defined in /etc/atomic.conf" % signer))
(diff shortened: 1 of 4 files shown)


rhatdan:
@baude @mtrmac @aweiteka PTAL

mtrmac:
ACK.

(I still have concerns about the configuration capabilities and the default, but those are unchanged by this PR and I will raise them separately.)

baude:
The reason I deviated from the "agreed" options was because after further consideration, it seemed the options were still insufficient.  In this implementation, while you can influence the directory the signature lands in, you cannot influence the signature's name.  So it will be signature-integer.

In the example we used, you had liked the idea of:
```
atomic sign -o busybox.sig -d /tmp
```
resulting in /tmp/busybox.sig.

If this approach is your preference, I'm ok with it.

rhatdan:
My problem with the -n busybox.sig is, what does the user do with it.  Since the atomic tool will have know way of finding it.  I believe @aweiteka pointed out that you could just do 

```
atomic sign -d /tmp/sigs
cd /tmp/sigs
tar zcvf /tmp/sigs.tgz .
scp /tmp/sigs.tgz root@signature.example.com
ssh root@signature.example.com sh -c "cd /var/www/html; tar xvf ~/sigs.tgz"
```

mtrmac:
> My problem with the -n busybox.sig is, what does the user do with it.

With `-o` (full path to a single destination file), some possible future equivalent of `koji import-sig`. We have designed this to be served from a static web server but it doesn’t _have_ to be done that way. Of course we don’t have a known user for any other way right now.

For `-n` (use the standard directory structure but a different name for the image), this would allow a workflow where an image is created under a `prerelease` repo and then copied into the `released` repo when QA is done, and `atomic sign -n released prerelease` could be used to prepare the signatures before actually pushing the image into the `released` Docker registry (so that at no moment in time there is a `released` image with missing signatures).

Really, none of `-o`, `-n` and `-d` is essential.

mtrmac:
> The reason I deviated from the "agreed" options was because after further consideration, it seemed the options were still insufficient.

@baude , looking at https://github.com/projectatomic/atomic/pull/539/files#diff-4a7d85c2a2907162beed01286a2149b3R52 , AFAICS the current `-o` implements the “root directory” semantics, not the “output file” semantics, and this PR is only changing the option naming. Am I missing anything?

rhatdan:
Ok I am going to merge and then we can revisit with these options when demand arises. 
@rh-atomic-bot r+

PR #586, opened 2016-09-06
[merged] Atomic/pull.py: Use skopeo to perform "pull" and signature verification

baude:
There is no change to a pull where --storage is ostree.  But for docker,
we now use skopeo copy to "pull" and image from a registry and verify
its signature (should the policy require it).

When docker is the destination for skopeo copy, it requires a tag.  If
no tag is given by the user, we inject "latest".

### DIFF
--- a/Atomic/pull.py
+++ b/Atomic/pull.py
@@ -1,16 +1,33 @@
-from .syscontainers import SystemContainers
+try:
+    from . import Atomic
+except ImportError:
+    from atomic import Atomic  # pylint: disable=relative-import
+from .util import skopeo_copy, get_atomic_config, skopeo_inspect, decompose
+
+ATOMIC_CONFIG = get_atomic_config()
+
 
 def cli(subparser):
     # atomic pull
+    backend = ATOMIC_CONFIG.get('default_storage', "ostree")
     pullp = subparser.add_parser("pull", help=_("pull latest image from a repository"),
                                  epilog="pull the latest specified image from a repository.")
     pullp.set_defaults(_class=Pull, func='pull_image')
-    pullp.add_argument("--storage", dest="backend", help=_("Specify the storage."))
+    pullp.add_argument("--storage", dest="backend", default=backend,
+                       help=_("Specify the storage. Default is currently '%s'.  You can"
+                              "change the default by editing /etc/atomic.conf and changing"
+                              "the 'default_storage' field." % backend))
     pullp.add_argument("image", help=_("image id"))
 
-class Pull(SystemContainers):
-    def __init__(self):
-        super(Pull, self).__init__()
 
+class Pull(Atomic):
     def pull_image(self):
-        super(Pull, self).pull_image()
+        if self.args.backend == 'ostree':
+            return self.syscontainers.pull_image()
+        _, _, tag = decompose(self.args.image)
+        # If no tag is given, we assume "latest"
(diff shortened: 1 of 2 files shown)


Atomic/pull.py
```diff
@@ -1,16 +1,31 @@
...
 
 def cli(subparser):
     # atomic pull
+    backend = ATOMIC_CONFIG.get('default_storage', "docker")
     pullp = subparser.add_parser("pull", help=_("pull latest image from a repository"),
                                  epilog="pull the latest specified image from a repository.")
     pullp.set_defaults(_class=Pull, func='pull_image')
-    pullp.add_argument("--storage", dest="backend", help=_("Specify the storage."))
+    pullp.add_argument("--storage", dest="backend", default=backend,
+                       help=_("Specify the storage."))
```
rhatdan:
```
 help=_("Specify the storage. Default is currently %s" % backend.  You can change the default by editing /etc/atomic.conf and changing the 'default_storage' field.))
```

Atomic/pull.py
```diff
@@ -1,16 +1,31 @@
...
 
-class Pull(SystemContainers):
-    def __init__(self):
-        super(Pull, self).__init__()
 
+class Pull(Atomic):
     def pull_image(self):
-        super(Pull, self).pull_image()
+        if self.args.backend == 'ostree':
+            self.syscontainers.pull_image()
```
rhatdan: Does this work?

rhatdan: Should this be 
```
return self.syscontainers.pull_image()
```

rhatdan:
@giuseppe PTAL

baude:
Changed ostree to the default storage option; all tests now pass.

Atomic/pull.py
```diff
@@ -1,16 +1,33 @@
...
-class Pull(SystemContainers):
-    def __init__(self):
-        super(Pull, self).__init__()
 
+class Pull(Atomic):
     def pull_image(self):
-        super(Pull, self).pull_image()
+        if self.args.backend == 'ostree':
+            return self.syscontainers.pull_image()
+        dummy1, dummy2, tag = decompose(self.args.image)
```
rhatdan: I think you could just do

_, _, tag = decompose(self.args.image)

Atomic/pull.py
```diff
@@ -1,16 +1,33 @@
...
 
-class Pull(SystemContainers):
-    def __init__(self):
-        super(Pull, self).__init__()
 
+class Pull(Atomic):
     def pull_image(self):
-        super(Pull, self).pull_image()
+        if self.args.backend == 'ostree':
+            return self.syscontainers.pull_image()
```
giuseppe: Syscontainers.pull_image at this point has some redundant code as it also checks for the backend.  We should replace Syscontainers.pull_image() to use directly `self._pull_image_to_ostree(self._get_ostree_repo(), self.args.image, True)` and drop all the different handlers logic there

giuseppe:
I had only one minor comment but we can fix it later, as anyway the pull part in Syscontainers will need some refactoring after this change.  The patch LGTM

rhatdan:
Fix the dummy stuff and then I will merge.  We also need to write some integration tests for testing some of the signing.

rhatdan:
@rh-atomic-bot r+

baude:
@rh-atomic-bot r+

PR #588, opened 2016-09-07
[merged] syscontainers: hide "systemctl is-active" output

giuseppe:
it is used only internally and has no useful information for the user.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -650,7 +650,7 @@ class SystemContainers(object):
 
     def _is_service_active(self, name):
         try:
-            return self._systemctl_command("is-active", name).replace("\n", "") == "active"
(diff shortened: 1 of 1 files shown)


Atomic/syscontainers.py
```diff
@@ -650,7 +650,7 @@ def get_system_images(self, get_all=False, repo=None):
 
     def _is_service_active(self, name):
         try:
-            return self._systemctl_command("is-active", name).replace("\n", "") == "active"
+            return self._systemctl_command("is-active", name, print_command=False).replace("\n", "") == "active"
```
rhatdan: How about we do quiet=True

giuseppe:
sure, I have just changed it :arrow_up:

rhatdan:
@rh-atomic-bot r+

jlebon:
This one does seem hung. Investigating...

jlebon:
@rh-atomic-bot retry

PR #590, opened 2016-09-07
[merged] Atomic/sign.py: Fix dir mod declaration

baude:
Convert string mod back to integer.  Disable the warning in pylint.

### DIFF
--- a/Atomic/sign.py
+++ b/Atomic/sign.py
@@ -74,7 +74,7 @@ class Sign(Atomic):
 
     @staticmethod
     def get_fingerprint(signer):
-        cmd = ['gpg', '--no-permission-warning', '--with-colons', '--fingerprint', signer]
+        cmd = ['gpg2', '--no-permission-warning', '--with-colons', '--fingerprint', signer]
         return_code, stdout, stderr = util.subp(cmd, newline=True)
         if return_code is not 0:
             raise ValueError(stderr)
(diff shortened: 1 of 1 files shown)


Atomic/sign.py
```diff
@@ -88,7 +88,7 @@ def make_sig_dirs(sig_path):
             # TODO
             # perhaps revisit directory permissions
             # when complete use-cases are known
-            os.makedirs(sig_path, '0777')
+            os.makedirs(sig_path, 0777) #pylint: disable=old-octal-literal
```
rhatdan: Does this work correctly in both python3 and python2?

baude: Apparently not.  Ill search for a solution.

Atomic/sign.py
```diff
@@ -88,7 +88,7 @@ def make_sig_dirs(sig_path):
             # TODO
             # perhaps revisit directory permissions
             # when complete use-cases are known
-            os.makedirs(sig_path, '0777')
+            os.makedirs(sig_path, mode=0o777)
```
aweiteka: Let's do 664

rhatdan: Don't specify it and it should work in both formats.  755 is probably what you want the permissions to be.

rhatdan:
@rh-atomic-bot r+

PR #591, opened 2016-09-07
sign dir should be integer and reduce permissions

aweiteka:


aweiteka:
@baude

rhatdan:
Needs to work on both python2 and python3.

rhatdan:
https://github.com/projectatomic/atomic/pull/590 is already attempting to deal with this.

PR #595, opened 2016-09-07
[merged] Atomic/sign.py: Add tag to signing image

baude:
Skopeo discards the tag when expanding the docker image
name.  We need the tag included when doing the verification,
so use util.decompose to derive the tag from the input name
and be sure to add it.

Also added some simple debug extensions to be able to see
what skopeo is doing when --debug is provided.


rhatdan:
@rh-atomic-bot r+

PR #597, opened 2016-09-08
Add support for deleting containers

rhatdan:
We want to be able to delete containers from the system, using the
new atomic containers delete call.

atomic containers delete --all

Comes in handy.

### DIFF
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -1,3 +1,4 @@
+import argparse
 import json
 import os
 
@@ -6,6 +7,7 @@ from . import Atomic
 from .client import AtomicDocker
 import datetime
 from dateutil.parser import parse as dateparse
+from docker.errors import NotFound, APIError
 
 try:
     from subprocess import DEVNULL  # pylint: disable=no-name-in-module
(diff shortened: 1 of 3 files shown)


rhatdan:
@giuseppe Should we be deleting syscontainers also?  Is there a difference between a container and an image from a system containers point of view.

giuseppe:
@rhatdan we do support deleting syscontainers, as part of `atomic uninstall $NAME`, if `$NAME` is a system container then it is uninstalled/deleted.  Should this just work for system containers as `atomic uninstall`.
In the system containers world, an image is a rootfs plus some template configuration files, while a container is a rootfs with finalized configuration files.  They share the same rootfs, but a container generates its configuration files from the templates provided by the image (as part of `/exports` in the rootfs).

rhatdan:
Either I document atomic delete does not delete system containers, and then add a plurb that you wrote above to man page, or we cause atomic delete to call atomic uninstall if it is a system container.  

Are we listing system containers with atomic containers list?

giuseppe:
yes, we list system containers with "atomic containers list".

I think it is good if we make "atomic containers delete" an alias for "atomic uninstall" for system containers.

The list returned from `self.get_containers()` must be filtered so that `self.d.remove_container` is used only for Docker containers.  For system containers you can just use `self.syscontainers.uninstall_system_container(name)`

rhatdan:
Have no idea how this got created with this name.  Repushed in https://github.com/projectatomic/atomic/pull/598

PR #598, opened 2016-09-08
[merged] Add support for deleting containers

rhatdan:
We want to be able to delete containers from the system, using the
new atomic containers delete call.

atomic containers delete --all

Comes in handy.

### DIFF
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -1,3 +1,4 @@
+import argparse
 import json
 import os
 
@@ -6,6 +7,7 @@ from . import Atomic
 from .client import AtomicDocker
 import datetime
 from dateutil.parser import parse as dateparse
+from docker.errors import NotFound, APIError
 
 try:
     from subprocess import DEVNULL  # pylint: disable=no-name-in-module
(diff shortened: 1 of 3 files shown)


rhatdan:
@giuseppe PTAL

Atomic/containers.py
```diff
@@ -194,3 +213,38 @@ def _filter_include_container(self, container_info):
...
+                if c["Type"] == "system":
+                    sys_targets.append(c["Id"])
+                else:
+                    docker_targets.append(c["Id"])
+        else:
+            for c in self.args.containers:
+                if self._inspect_container(c):
+                    docker_targets.append(c)
+                else:
+                    sys_targets.append(c)
```
giuseppe: we look up system containers first then try the docker version.  I think here we should do:
```
if self.syscontainers.get_system_container_checkout(self, name):
    sys_targets.append(c)
else:
    docker_targets.append(c)
```

rhatdan: You mean?
```
if self.syscontainers.get_system_container_checkout(c):
    sys_targets.append(c)
else:
    docker_targets.append(c)
```

giuseppe: yes, sorry :)

giuseppe:
@rh-atomic-bot r+ 2fb55c7

PR #603, opened 2016-09-09
[merged] Atomic/sign.py|util.py Support sigstore from conf files

baude:
We now derive the proper sigstore from a series of YAML
configuration files in /etc/containers/registries.d.  These
configuration files can have: sigstore and sigstore-write as
keys and the values are a file path or http|s URL.

When signing an image, as long as the -d override is not used,
we use those values when writing local signatures.

### DIFF
--- a/Atomic/sign.py
+++ b/Atomic/sign.py
@@ -11,10 +11,6 @@ ATOMIC_CONFIG = util.get_atomic_config()
 def cli(subparser):
     # atomic sign
     signer = ATOMIC_CONFIG.get('default_signer', None)
-    signature_path = util.get_atomic_config_item(['default-sigstore-path'], atomic_config=ATOMIC_CONFIG)
-    if signature_path is None:
-        signature_path = util.ATOMIC_VAR_LIB + '/sigstore'
-
     signp = subparser.add_parser("sign",
                                  help="Sign an image",
                                  epilog="Create a signature for an image which can be "
@@ -25,19 +21,19 @@ def cli(subparser):
                        help=_("Name of the signing key. Currently %s, "
                               "default can be defined in /etc/atomic.conf" % signer))
     signp.add_argument("-d", "--directory",
-                       default=signature_path,
+                       default=None,
                        dest="signature_path",
-                       help=_("The directory to store signatures under: Default {}.".format(signature_path) ))
+                       help=_("Define an alternate directory to store signatures"))
 
 class Sign(Atomic):
     def sign(self):
-        # TODO
-        # Atomic is run as sudo. Should we work around that?
 
         if self.args.debug:
             util.write_out(str(self.args))
 
         signer = self.args.sign_by
+        registry_config_path = util.get_atomic_config_item(["registry_confdir"], ATOMIC_CONFIG)
+        registry_configs, default_store = util.get_registry_configs(registry_config_path)
 
         for sign_image in self.args.images:
             remote_inspect_info = util.skopeo_inspect("docker://{}".format(sign_image))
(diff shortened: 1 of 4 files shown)


Atomic/sign.py
```diff
@@ -50,14 +46,46 @@ def sign(self):
...
+                    if not os.path.exists(self.args.signature_path):
+                        raise ValueError("The path {} does not exist".format(self.args.signature_path))
+                    fq_sig_path = os.path.join(self.args.signature_path,
+                                               self.get_sig_name(self.args.signature_path))
+
+                else:
+                    reg, repo, _ = util.decompose(expanded_image_name)
+                    reg_info = util.have_match_registry("{}/{}".format(reg, repo), registry_configs)
+                    # No default and no registry match, exit
+                    if reg_info is None and default_store is None:
```
rhatdan: Couldn't this be simplified to 
```
if not reg_info:
       reg_info = default_store
if not reg_info:
...
```

Atomic/sign.py
```diff
@@ -50,14 +46,46 @@ def sign(self):
...
+                        raise ValueError("Unable to associate {} with "
+                                         "configurations in {} and no 'default-docker' "
+                                         "is defined.".format(sign_image, registry_config_path))
+                    # If no registry match but there is a default defined, use it
+                    elif reg_info is None and default_store is not None:
```
rhatdan: Then this would not be needed.

Atomic/sign.py
```diff
@@ -50,14 +46,46 @@ def sign(self):
...
+                        # Make sure signature path exists
+                        if not os.path.exists(signature_path):
+                            raise ValueError("The signature path {} does not exist".format(signature_path))
+
+                    sigstore_path = "{}/{}/{}@{}".format(signature_path, os.path.dirname(expanded_image_name),
+                                                         os.path.basename(expanded_image_name), manifest_hash)
+                    self.make_sig_dirs(sigstore_path)
+                    sig_name = self.get_sig_name(sigstore_path)
+                    fq_sig_path = os.path.join(sigstore_path, sig_name)
+                    if os.path.exists(fq_sig_path):
```
rhatdan: Should we have done this a lot earlier?

Atomic/util.py
```diff
@@ -589,3 +591,54 @@ def expandvars(path, environ=None):
...
+            try:
+                temp_conf = safe_load(conf_file)
+                if isinstance(temp_conf, dict):
+                    def_store = temp_conf.get('default-docker', None)
+                    if def_store is not None and default_store is not None:
+                        raise ValueError("There are duplicate entries for 'default-docker' in {}.".format(yaml_dir))
+                    elif default_store is None and def_store is not None:
+                        default_store = def_store
+                    registries = temp_conf.get('docker', None)
+                else:
```
rhatdan: Should this be an error or a continue?

Atomic/sign.py
baude: I'm not sure I follow your logic.  The english representation is:

If there is no registry configuration match and no default, then error.

baude: Oh follow now, yeah, will change.

Atomic/sign.py
baude: Yup, fixed.

Atomic/sign.py
baude: Nope, this is the earliest.  This hasn't changed since the first submission.  We need to collection all the information including path and signature name until we can run this check.

Atomic/util.py
baude: It depends on how strict you want to be.  I was trying to allow for a file that wasn't accurate and just move on.  If you want to error out and be more strict, we can do that too.

rhatdan: Since this has potential security concerns, I think we should probably error out.  But I think skopeo would fail on an atomic pull anyways.

The thing I would worry about would be allowing an image to be pull/run that did not pass the "trust" policy.

Atomic/sign.py
rhatdan: I was thinking this should happen before make_sig_dirs, but I guess the sig_dirs would have to be there in order for this to fail

rhatdan:
@rh-atomic-bot r+

PR #605, opened 2016-09-10
[merged] Atomic/util.py: Allow stdin for skopeo calls (BZ 1374851)

baude:
In some skopeo calls like standalone-sign and copy, the system
may need to present a password challenge to the user.  The current
subp calls would not allow this.  We now use util.call and pass
stdin=sys.stdin.

This resolves BZ #1374851.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -324,29 +324,20 @@ def skopeo_layers(image, args=None, layers=None):
             shutil.rmtree(temp_dir)
     return temp_dir
 
-def skopeo_subprocess(cmd):
-    try:
-        results = subp(cmd)
-    except Exception as e: # pylint: disable=broad-except
-        raise ValueError(e)
-    if results.return_code is not 0:
-        raise ValueError(results.stderr)
-    return results
-
 def skopeo_standalone_sign(image, manifest_file_name, fingerprint, signature_path, debug=False):
     cmd = ['skopeo']
     if debug:
         cmd = cmd + ['--debug']
     cmd = cmd + ['standalone-sign', manifest_file_name, image,
                  fingerprint, "-o", signature_path]
-    return skopeo_subprocess(cmd)
+    return check_call(cmd)
 
 def skopeo_manifest_digest(manifest_file, debug=False):
     cmd = ['skopeo']
     if debug:
         cmd = cmd + ['--debug']
     cmd = cmd  + ['manifest-digest', manifest_file]
-    return skopeo_subprocess(cmd).stdout.rstrip().decode()
+    return check_output(cmd).rstrip().decode()
 
 def skopeo_copy(source, destination, debug=False):
     cmd = ['skopeo']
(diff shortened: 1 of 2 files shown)


Atomic/util.py
```diff
@@ -331,13 +331,22 @@ def skopeo_subprocess(cmd):
         raise ValueError(results.stderr)
     return results
 
+def skopeo_call(cmd):
+    try:
+        return_code = call(cmd, stdout=sys.stdout, stderr=sys.stderr, stdin=sys.stdin)
+    except Exception as e: # pylint: disable=broad-except
```
rhatdan: This try block does nothing.  Just remove it.

Atomic/util.py
```diff
@@ -354,7 +363,9 @@ def skopeo_copy(source, destination, debug=False):
     if destination.startswith("docker-daemon"):
         cmd = cmd + ['--remove-signatures']
     cmd = cmd + [source, destination]
-    return skopeo_subprocess(cmd)
+    return_code = call(cmd)
+    if return_code is not 0:
+        raise Exception("The command '{}' failed".format(" ".join(cmd)))
```
rhatdan: Can we just remove subp and always use something from the call chain.

Atomic/util.py
```diff
@@ -331,13 +331,22 @@ def skopeo_subprocess(cmd):
...
+    return return_code
+
 def skopeo_standalone_sign(image, manifest_file_name, fingerprint, signature_path, debug=False):
     cmd = ['skopeo']
     if debug:
         cmd = cmd + ['--debug']
     cmd = cmd + ['standalone-sign', manifest_file_name, image,
                  fingerprint, "-o", signature_path]
-    return skopeo_subprocess(cmd)
+    return_code = call(cmd)
```
rhatdan: If you call check_call, this will raise exceptions automatically

Atomic/util.py
rhatdan: If you call check_call, this will raise exceptions automatically

Atomic/util.py
baude: I was accounting for the case where an exception may not be raised but the command actually failed.  Not worth it?

Atomic/util.py
baude: We can.  I was using subp for cases where no input would be required.  I am happy to alter however.

Atomic/util.py
baude: Done

Atomic/util.py
baude: REviewing and in this case, I need the return stdout.  Check call does not account for that.

Atomic/util.py
baude: Fixed.

rhatdan:
Call check_call and you don't have to look at the return call.
```
 subprocess.check_call(args, *, stdin=None, stdout=None, stderr=None, shell=False)

    Run command with arguments. Wait for command to complete. If the return code was zero then return, otherwise raise CalledProcessError. The CalledProcessError object will have the return code in the returncode attribute.
```

rhatdan:
Were calling CheckCall all over the place, so if we are going to start looking at the caller, we should catch the CalledProcessError exception, and raise.  Also we should not raise Exception.  Something like CalledProcessError seems more appropriate.

rhatdan:
@rh-atomic-bot r+

PR #607, opened 2016-09-12
[merged] Atomic/util.py: Add filename to registries configurations

baude:
Aaron's work in policy and configuration editing would be simplier
if we retained a relationship between the registry configuration
and the filename it came from.  Now we simply add a 'filename'
key into the dictionary being returned by
get_registry_configs().

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -584,6 +584,7 @@ def expandvars(path, environ=None):
     return path
 
 def get_registry_configs(yaml_dir):
+    # Returns a dictionary of registries and a str of the default_store if applicable
     regs = {}
     default_store = None
     # Get list of files that end in .yaml and are in fact files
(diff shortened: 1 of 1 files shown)


baude:
@aweiteka Please review this PR to make sure it is what you wanted.  You can exercise this with the check'd out branch like:

```
# python -c "from Atomic.util import get_registry_configs, have_match_registry; foo, _ = get_registry_configs('/etc/containers/registries.d/'); print(have_match_registry('docker.io/library/hello-world', foo))"
{'sigstore': 'http://172.17.0.2/signatures/', 'sigstore-write': 'file:///var/lib/atomic/sigstore/', 'filename': '/etc/containers/registries.d/foobar.yaml'}
#
```

aweiteka:
@baude works great. Thanks.

rhatdan:
@rh-atomic-bot r+

PR #610, opened 2016-09-13
[merged] Atomic/sign.py: Don't fail for missing /etc/containers/registries.d w…

baude:
…ith -d

This is a fix for bugzilla #1375654 where atomic sign -d fails because
it still tries to parse the registries.d/ YAML files.  Theoretically the
directory should always exist but this small code fix provides
a work-around.


rhatdan:
@rh-atomic-bot r+

PR #612, opened 2016-09-13
[merged] Atomic/install.py: Only check args.setvalues if OSTREE_PRESENT

baude:
Because the setvalues argument is only present when
OSTREE_PRESENT, we should only check for it when
OSTREE_PRESENT as well.  This bug is raised in
bugzilla #1374676.


lsm5:
i picked this patch on top of 1.12.1, so we should be fine even without merge

rhatdan:
@rh-atomic-bot r+

Issue #574, opened 2016-08-31
duplicate scanner entries listed when bad scanner defined in /etc/atomic.d


miabbott:
it looks like PR #565 fixed the problem of handling bad data in `/etc/atomic.d`, but now duplicate scanner entries are printed out when using `atomic scan --list`

```
[root@mastershake atomic]# git rev-parse HEAD
5fec4354e1db5f24919c815d1740f59b84bf30c5
[root@mastershake atomic]# ./atomic install fedora/atomic_scan_openscap
docker run --rm --privileged -v /:/host/ fedora/atomic_scan_openscap sh /root/install.sh

Installing the configuration file 'atomic_scan_openscap' into /etc/atomic.d/.  You can now use this scanner with atomic scan with the --scanner atomic_scan_openscap command-line option.  You can also set 'atomic_scan_openscap' as the default scanner in /etc/atomic.conf.  To list the scanners you have configured for your system, use 'atomic scan --list'.

Saving current config.ini as config.ini.2016-08-31-16:23:52.atomic_save
Updating config.ini with latest configuration
Installation complete. You can customize /etc/oscapd/config.ini as needed.

[root@mastershake atomic]# ls -l /etc/atomic.d
total 4
-rw-r--r--. 1 root root 410 Aug 31 12:23 atomic_scan_openscap
[root@mastershake atomic]# ./atomic scan --list
Scanner: atomic_scan_openscap 
  Image Name: docker.io/fedora/atomic_scan_openscap
     Scan type: cve * 
     Description: Performs a CVE scan based on known CVE data


* denotes defaults
[root@mastershake atomic]# dd if=/dev/urandom of=/etc/atomic.d/bad-openscap bs=1 count=410
410+0 records in
410+0 records out
410 bytes (410 B) copied, 0.00143437 s, 286 kB/s
[root@mastershake atomic]# ls -l /etc/atomic.d
total 8
-rw-r--r--. 1 root root 410 Aug 31 12:23 atomic_scan_openscap
-rw-r--r--. 1 root root 410 Aug 31 12:25 bad-openscap
[root@mastershake atomic]# ./atomic scan --list
Error: Unable to load scannerfile /etc/atomic.d/bad-openscap.  Continuing...
Error: Unable to load scannerfile /etc/atomic.d/bad-openscap.  Continuing...
Scanner: atomic_scan_openscap 
  Image Name: docker.io/fedora/atomic_scan_openscap
     Scan type: cve * 
[...]

PR #613, opened 2016-09-14
[merged] Don't add scanners if they have bad format

rhatdan:
This fixes https://github.com/projectatomic/atomic/issues/574


rhatdan:
@baude PTAL

rhatdan:
I am merging to catch the next release.

@rh-atomic-bot r+

PR #618, opened 2016-09-14
[merged] Atomic/sign.py: Raise error on writing signatures to http

baude:
Atomic cannot write a signature to web(http/https) target. We
now raise an error as such.

sudo ./atomic sign docker.io/library/hello-world

Writing to http://123 is not supported. Use a file:///location instead.

### DIFF
--- a/Atomic/sign.py
+++ b/Atomic/sign.py
@@ -72,6 +72,9 @@ class Sign(Atomic):
                     if signature_path is None:
                         raise ValueError("No write path for {}/{} was "
                                          "found in {}".format(reg, repo, registry_config_path))
+                    elif signature_path.startswith("http"):
+                        raise ValueError("Writing to {} is not supported. Use a "
+                                         "file:///location instead.".format(signature_path))
 
                     # Deal with write path prepends
                     if signature_path.startswith("file://"):
--- a/docs/atomic-sign.1.md
+++ b/docs/atomic-sign.1.md
@@ -63,7 +63,7 @@ docker:
   privateregistry.example.com:
     sigstore: file:///var/lib/atomic/signature
 
-When signing an image preceeded with the registry name 'privateregistrt.example.com',
+When signing an image preceeded with the registry name 'privateregistry.example.com',
 the signature will be written into subdirectories of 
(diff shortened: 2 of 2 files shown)


rhatdan:
@rh-atomic-bot r+

Atomic/sign.py
```diff
@@ -72,6 +72,9 @@ def no_reg_no_default_error(image, registry_path):
                     if signature_path is None:
                         raise ValueError("No write path for {}/{} was "
                                          "found in {}".format(reg, repo, registry_config_path))
+                    elif signature_path.startswith("http"):
+                        raise ValueError("Writing to {} is not supported. Use a "
+                                         "file:///location instead.".format(signature_path))
 
                     # Deal with write path prepends
                     if signature_path.startswith("file://"):
```
mtrmac: This is still pretty brittle; what if I say `ftp:`? This should only accept known URL schemas, not randomly look for specific prefixes with `startswith`.

baude: @mtrmac do you have known URL schemas? I can add them.

mtrmac: That’s not what I mean.
```python
url =  urlparse.urlparse(signature_path) # + error handling
switch url.scheme:
   case 'file': ... # use url.path
   default: ... # fail with "scheme %s not supported for writing signatures" % url.schem
```

or something; parse using the real structure of the string, do not just check for a small subset of possible values.

rhatdan: Why not check if the path exists?  I would check for file:// and remove it and then to an os.path.exists(), if the user hands in a URI other then file it will fail with a path does not exist.

mtrmac: If I understand  correctly, that still seems brittle and semantically incorrect; if the user _did_ for some crazy reason have a local directory tree named `http://example.com/sigstore`, we _still_ wouldn’t want to use that local path.

Use real URL parsing, and extract URL schema first. _If_ it is a `file:`, sure, then we can do an `os.path.exists() ` to be friendly. But we should be refusing to write to a `http` regardless of how reasonable or crazy the local filesystem contents are.

PR #619, opened 2016-09-15
[merged] syscontainers: display better container status

yuqi-zhang:
Change the status in "atomic containers list" to display systemd
state instead of runc state. This way system containers match user
containers. Another reason for the switch is that systemctl
start/stop creates/destroys a runc container every time instead of
using the same container, so systemd gives better indication of
container status.

Also expand the states to be:

 - running: container is running correctly
 - inactive: container is stopped/newly created and ready to run
 - failed: either systemd service or container process failed

instead of just "exited" and "running" as before.

Add tests and fix to test-image originally by Giuseppe.


yuqi-zhang:
running/inactive/failed are probably not optimal names. Also was considering changing status to display time ("running for x min/sec") but that seems a bit verbose, so I left it out of this patch.

rhatdan:
@giuseppe PTAL

Atomic/syscontainers.py
```diff
@@ -549,17 +548,14 @@ def update_system_container(self, name):
...
 
     def get_container_runtime_info(self, container):
-        if self.user:
-            return {'status' : "running" if self._is_service_active(container) else "exited"}
 
-        try:
-            inspect_stdout = util.check_output([util.RUNC_PATH, "state", container], stderr=DEVNULL)
-            ret = json.loads(inspect_stdout.decode())
-            status = ret["status"]
-            created = dateparse(ret['created'])
```
giuseppe: I think we should still keep using `created`, when available and simply discard `status`

yuqi-zhang: Currently we display the "created time" of the container saved in info, which is when we installed the container. I think that is better because if we systemctl stop, and then systemctl start, the runc created time would be refreshed, and it would display LAST run-time instead.

giuseppe:
I have tested it and it works well.  Good job!

@rh-atomic-bot r+ b3b17d8

PR #620, opened 2016-09-15
[merged] [RFC] syscontainers: change ostree tag separator to '/'

giuseppe:
translating the ':' separator for tag to '-' for the ostree branch was
a wrong choice as we are not able to handle tags like 1.2-3.
It is a breaking change with what we are shipping with atomic 1.12,
but better to do this now than carrying this issue.

Fetched images can be converted executing the output of:

```
find /ostree/repo/refs/ -type f | sed -ne "/ociimage.*-/ s|\(.*ociimage/\)\(.*\)-\(.*\)|mkdir -p \1\2; mv \1\2-\3 \1\2/\3|g p"
```

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1017,7 +1017,7 @@ class SystemContainers(object):
         if "ostree:" in img:
             imagebranch = img.replace("ostree:", "")
         else: # assume "oci:" image
-            imagebranch = "%s%s" % (OSTREE_OCIIMAGE_PREFIX, SystemContainers._encode_to_ostree_ref(img))
+            imagebranch = "%s%s" % (OSTREE_OCIIMAGE_PREFIX, SystemContainers._encode_to_ostree_ref(img.replace("sha256:", "")))
         return imagebranch
 
     def has_system_container_image(self, img):
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -610,8 +610,8 @@ class SystemContainers(object):
         commit_rev = repo.resolve_rev(imagebranch, False)[1]
         commit = repo.load_commit(commit_rev)[1]
 
-        branch_id = imagebranch.replace(OSTREE_OCIIMAGE_PREFIX, "")
-        tag = ":".join(branch_id.rsplit('-', 1))
+        branch_id = SystemContainers._decode_from_ostree_ref(imagebranch.replace(OSTREE_OCIIMAGE_PREFIX, ""))
+        tag = ":".join(branch_id.rsplit(':', 1))
         timestamp = OSTree.commit_get_timestamp(commit)
         labels = {}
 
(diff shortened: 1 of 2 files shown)


giuseppe:
@cgwalters could you please review this change?  Is this better than the previous version?

giuseppe:
@aveshagarwal this should fix the issue with 'rhel7/etcd:2.3.7-8'

rhatdan:
I am about to create atomic-1.12.2, if this patch is ok with Colin we should get it into that release before this gets out to RHEL.

cgwalters:
Hm, ideally it's a bidirectional mapping.  If both can contain `/` then it becomes ambiguous, right?  

Also, it seems like we could solve the issue by only translating the *first* `-` to `:` maybe?

Anyways I'm not objecting to this patch exactly, just trying to think through it a bit more.

giuseppe:
yes, it must be bidirectional.  Since there is always a tag in the images ('latest' if not specified), can't we can always use the last component of the OSTree reference `/path` as the tag name since a tag cannot contain '/'?

if we translate only the first `-` to `:` then we will have the opposite problem that an image name cannot have '-' in its name, e.g. `spc-etcd-latest`.

Another question since we are at it, a fully qualified image like `my.docker.registry.com:8888/an/image:tag` is stored as `an/image/tag`.  Should we keep the registry as well?  If yes, how can we translate the registry part (it can also contain a `:` for the port)?

cgwalters:
How about we keep it simple and do something like URL-encoding, I think we could just use `urllib.quote_plus()`?

giuseppe:
ostree doesn't like that either: `Error: g-io-error-quark: Invalid refspec ociimage/gscrivano%2Fetcd (0)`

giuseppe:
I'll try with base64

cgwalters:
Right, no `%` in refs right now.  Base64 will work but sacrifices legibility.  How about copying https://github.com/projectatomic/rpm-ostree/blob/master/src/libpriv/rpmostree-core.c#L748

giuseppe:
I forgot to add a fixup patch, but here it is a version that escapes an image name to use only the characters allowed by ostree :arrow_up:

giuseppe:
should this change get it?

Atomic/syscontainers.py
```diff
@@ -982,12 +981,43 @@ def extract_system_container(self, img, destination):
...
+        except ValueError:
+            return name
+
+    @staticmethod
     def _get_ostree_image_branch(img):
         if "ostree:" in img:
             imagebranch = img.replace("ostree:", "")
         else: # assume "oci:" image
-            _, image, tag = SystemContainers._parse_imagename(img.replace("oci:", "").replace("docker:", ""))
-            imagebranch = "%s%s-%s" % (OSTREE_OCIIMAGE_PREFIX, image.replace("sha256:", ""), tag)
```
giuseppe: Good catch.  I am going to re-add it.  It should not happen but better be safe

rhatdan:
@rh-atomic-bot r+

cgwalters:
@rh-atomic-bot r+ bc8f189

PR #621, opened 2016-09-15
[merged] Atomic/util.py: Deal with tls-verify change in skopeo (BZ #1375891)

baude:
skopeo now defaults to using tls-verify where it had not
prior.  This means interaction with insecure registries
began to fail.  Because we were dealing with insecure
registries prior, we will use --tls-verify-false with only
certain non-signing skopeo functions until we can
implement a way to get a list of insecure registries
from the docker configuration including dealing with
CIDR addressing.

This fixed bugzilla #1375891.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -17,10 +17,6 @@ import tempfile
 import shutil
 import re
 import requests
-try:
-    from urlparse import urlparse #pylint: disable=import-error
-except ImportError:
-    from urllib.parse import urlparse #pylint: disable=no-name-in-module,import-error
 
 # Atomic Utility Module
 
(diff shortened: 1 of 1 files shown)


Atomic/util.py
```diff
@@ -349,13 +346,6 @@ def skopeo_copy(source, destination, debug=False):
     cmd = cmd + [source, destination]
     return check_call(cmd)
 
-def check_v1_registry(image):
```
rhatdan: Why is this being removed?  If correct add comment to patch indicating why?

rhatdan:
@rh-atomic-bot r+

Issue #625, opened 2016-09-16
atomic trust will not create or edit empty policy.json 


ddarrah:
If the file is missing or has no valid json content, tool will not run.  The argument could be made that it should never edit a corrupted file, but it would be fairly straightforward to create a new one if the file is not there.  This would allow a user a simple way to reset by blowing it away and adding new trust with the tool.

aweiteka:
@ddarrah this should be fixed with #628

rhatdan:
Fixed in atomic-1.13

PR #628, opened 2016-09-17
[merged] Create policy file if it does not exists.

rhatdan:
Currently the policy file does not exist the tools fail, and the user
is scrambling to setup a correct policy file.


rhatdan:
@aweiteka PTAL

rhatdan:
Fixes https://github.com/projectatomic/atomic/issues/625

Atomic/trust.py
```diff
@@ -120,8 +123,13 @@ def delete(self):
...
 
     def default(self):
-        with open(self.policy_filename, 'r+') as policy_file:
-            policy = json.load(policy_file)
+        mode = "r+" if os.path.exists(self.policy_filename) else "w+"
+        with open(self.policy_filename, mode) as policy_file:
+            if mode == "r+":
+                policy = json.load(policy_file)
+            else:
+                policy = { "default":[{}] }
```
aweiteka: I think we need the default type. See https://github.com/projectatomic/skopeo/blob/master/default-policy.json

```
{ "default": [ { "type": "insecureAcceptAnything" }] }
```

rhatdan:
@rh-atomic-bot r+

Issue #626, opened 2016-09-16
atomic trust default adds entry into docker transport


ddarrah:
If there is no default section in the policy.json, using the 'atomic trust default' command adds it to the docker transport.  It edits the policy fine if it is in the file.  This is probably low priority, just being thorough.

```
[root@rhel-base openshift-origin-client-tools-v1.3.0-rc1-ac0bb1bf6a629e0c262f04636b8cf2916b16098c-linux-32bit]# atomic trust default accept
'default'
[root@rhel-base openshift-origin-client-tools-v1.3.0-rc1-ac0bb1bf6a629e0c262f04636b8cf2916b16098c-linux-32bit]# cat /etc/containers/policy.json 
{
    "transports": {
        "docker": {
            "default": [
                {
                    "type": "insecureAcceptAnything"
                }
            ]
        }, 
        "atomic": {}, 
        "dir": {
            "registry.example.com/something": [
                {
                    "keyType": "GPGKeys", 
                    "type": "signedBy", 
                    "keyPath": "/etc/containers/pki/ddarrah@redhat.com.key"
                }
            ]
        }
    }
```

rhatdan:
@aweiteka PTAL

aweiteka:
@ddarrah this should be fixed by #630

PR #630, opened 2016-09-19
If default option missing from policy then add it.

rhatdan:


rhatdan:
@aweiteka PTAL

rhatdan:
Trying to replicate https://github.com/projectatomic/atomic/issues/626 found this issue.

PR #631, opened 2016-09-19
[merged] Atomic/push.py: Implement push with signing

baude:
Enabling signing and pushing at the same time.  At the time of
this writing, we cannot verify that the atomic targets are
working but the code is there.  THis is enabled with the
--type atomic switch. Otherwise, signatures are written locally
and the image is pushed.

### DIFF
--- a/Atomic/pull.py
+++ b/Atomic/pull.py
@@ -2,8 +2,9 @@ try:
     from . import Atomic
 except ImportError:
     from atomic import Atomic  # pylint: disable=relative-import
-from .util import skopeo_copy, get_atomic_config, skopeo_inspect, decompose, write_out
 from .trust import Trust
+from .util import skopeo_copy, get_atomic_config, skopeo_inspect,\
+    decompose, write_out, strip_port, is_insecure_registry
 
 ATOMIC_CONFIG = get_atomic_config()
 
@@ -18,20 +19,27 @@ def cli(subparser):
                        help=_("Specify the storage. Default is currently '%s'.  You can"
                               " change the default by editing /etc/atomic.conf and changing"
                               " the 'default_storage' field." % backend))
+    pullp.add_argument("-t", "--type", dest="reg_type", default=None,
+                       help=_("Push to an alternative registry type."))
     pullp.add_argument("image", help=_("image id"))
 
 
 class Pull(Atomic):
     def pull_docker_image(self):
-        _, _, tag = decompose(self.args.image)
+        registry, _, tag = decompose(self.args.image)
+        insecure = True if is_insecure_registry(self.d.info()['RegistryConfig'], strip_port(registry)) else False
         # If no tag is given, we assume "latest"
         tag = tag if tag != "" else "latest"
-        fq_name = skopeo_inspect("docker://{}".format(self.args.image))['Name']
+        if self.args.reg_type == "atomic":
+            pull_uri = 'atomic:'
+        else:
+            pull_uri = 'docker://'
+        fq_name = skopeo_inspect("{}{}".format(pull_uri, self.args.image))['Name']
         image = "docker-daemon:{}:{}".format(fq_name, tag)
         trust = Trust()
         trust.args.assumeyes=self.args.assumeyes
         trust.discover_sigstore(fq_name)
-        skopeo_copy("docker://{}".format(self.args.image), image, debug=self.args.debug)
+        skopeo_copy("docker://{}".format(self.args.image), image, debug=self.args.debug, insecure=insecure)
 
     def pull_image(self):
         handlers = {
--- a/Atomic/push.py
+++ b/Atomic/push.py
@@ -1,6 +1,6 @@
 import getpass
-import json
 
+from .sign import Sign
 from . import util
 from . import pulp
 from . import satellite
@@ -10,7 +10,10 @@ try:
 except ImportError:
     from atomic import Atomic # pylint: disable=relative-import
 
+ATOMIC_CONFIG = util.get_atomic_config()
+
 def cli(subparser):
+    signer = ATOMIC_CONFIG.get('default_signer', None)
     # atomic push
     pushp = subparser.add_parser(
         "push", aliases=['upload'], help=_("push latest image to repository"),
@@ -32,10 +35,6 @@ def cli(subparser):
                          default=None,
                          action="store_true",
                          help=_("flag to verify ssl of registry"))
-    pushp.add_argument("--debug",
-                         default=None,
-                         action="store_true",
-                         help=_("debug mode"))
     pushp.add_argument("-U", "--url",
                          dest="url",
                          default=None,
@@ -57,6 +56,11 @@ def cli(subparser):
                          default=None,
                          dest="repo_id",
                          help=_("Repository ID"))
+    pushp.add_argument("-t", "--type", dest="reg_type", default=None,
+                       help=_("Push to an alternative registry type."))
+    pushp.add_argument("--sign-by", dest="sign_by", default=signer,
+                       help=_("Name of the signing key. Currently %s, "
+                              "default can be defined in /etc/atomic.conf" % signer))
     # pushp.add_argument("--activation_key_name",
     #                      default=None,
     #                      dest="activation_key_name",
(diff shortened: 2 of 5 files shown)


Atomic/push.py
```diff
@@ -133,19 +142,41 @@ def push(self):
...
+
+            # We must push the file to the registry first prior to performing a
+            # local signature because the manifest file must be on the registry
+            return_code = util.skopeo_copy(local_image, remote_image, debug=self.args.debug,
+                                           sign_by=self.args.sign_by if not sign_local else None, insecure=insecure)
+
+            if return_code != 0:
+                raise ValueError("Pushing {} failed.".format(self.image))
+
+            util.write_out("Pushed: {}".format(self.image))
```
rhatdan: No debug statements.

Atomic/push.py
```diff
@@ -133,19 +142,41 @@ def push(self):
...
+            local_image = "docker-daemon:{}".format(self.image)
+            if self.args.reg_type == "atomic":
+                remote_image = "atomic:{}".format(self.image)
+                sign_local = False
+            else:
+                remote_image = "docker://{}".format(self.image)
+
+            sign = True if self.args.sign_by else False
+            if sign:
+                util.write_out("\nSigning with '{}'\n".format(self.args.sign_by))
```
rhatdan: No debug statements.

Atomic/push.py
```diff
@@ -133,19 +142,41 @@ def push(self):
...
-                elif status.startswith("Pushing"):
-                    util.write_out("Pushing: " + bar['id'])
-
-                prevstatus = status
+            reg, _, tag = util.decompose(self.image)
+            if not tag:
+                raise ValueError("The image being pushed must have a tag")
+
+            if self.args.password and self.args.username:
+                print(self.d.login(username=self.args.username, password=self.args.password, registry=reg))
```
rhatdan: No debug messages.

Atomic/sign.py
```diff
@@ -25,24 +30,38 @@ def cli(subparser):
                        help=_("Define an alternate directory to store signatures"))
 
 class Sign(Atomic):
-    def sign(self):
+    def __init__(self):
+        super(Sign, self).__init__()
+
+    def sign(self, debug=None, signer=None, in_signature_path=None, images=None):
```
rhatdan: Make debug a param of __init__, Maybe others

Atomic/push.py
```diff
@@ -133,19 +142,41 @@ def push(self):
...
+            if return_code != 0:
+                raise ValueError("Pushing {} failed.".format(self.image))
+
+            util.write_out("Pushed: {}".format(self.image))
+
+            if sign_local and sign:
+                if self.args.debug:
+                    util.write_out("Need to create a local signature")
+                atomic_sign = Sign()
+                atomic_sign.sign(debug=self.args.debug, signer=self.args.sign_by,
```
rhatdan: We could just do
atomic_sign.args.sign_by=self.args.sign_by
...

Rather then adding all of these parameters.

rhatdan:
Also please fix the test failures.

Atomic/push.py
baude: I dont think that will work because the self.args differ between the commands.

Atomic/push.py
baude: Fixed.

Atomic/push.py
baude: Wrapped with if debug

Atomic/push.py
baude: wrapped with if debug

Atomic/push.py
```diff
@@ -71,12 +75,17 @@ def cli(subparser):
...
     def push(self):
         self.ping()
-        prevstatus = ""
+        if self.args.debug:
+            util.write_out(str(self.args))
+
+        # This allows a user to turn off signing when a global
+        # sign_by has been defined in /etc/atomic.conf and saves
+        # us from having to define something like --no-sign
+        if self.args.sign_by == "None":
```
rhatdan: Who is passing in "None" string?

Atomic/pull.py
```diff
@@ -18,20 +19,27 @@ def cli(subparser):
...
+    pullp.add_argument("-t", "--type", dest="reg_type", default=None,
+                       help=_("Push to an alternative registry type."))
     pullp.add_argument("image", help=_("image id"))
 
 
 class Pull(Atomic):
     def pull_docker_image(self):
-        _, _, tag = decompose(self.args.image)
+        registry, _, tag = decompose(self.args.image)
+        insecure = True if is_insecure_registry(self.d.info()['RegistryConfig'], remove_port(registry)) else False
```
rhatdan: The remove_port() here seems like a strange function name to call?

Atomic/pull.py
```diff
@@ -18,20 +19,27 @@ def cli(subparser):
...
         # If no tag is given, we assume "latest"
         tag = tag if tag != "" else "latest"
-        fq_name = skopeo_inspect("docker://{}".format(self.args.image))['Name']
+        if self.args.reg_type == "atomic":
+            pull_uri = 'atomic:'
```
rhatdan: The atomic does not use the atomic:// path?

Atomic/push.py
rhatdan: You just have to make sure that Sign has them all defined in its init.

Atomic/util.py
```diff
@@ -635,3 +647,64 @@ def get_signature_read_path(reg_info):
     # Return the defined path for where signatures should be read
     # or none if no entry is found
     return reg_info.get('sigstore', None)
+
+def remove_port(_input):
```
rhatdan: Maybe this should be called strip_port_text or something like that, I see remove_port, and I think you are changing something in the configuration.

Atomic/push.py
baude: you can pass None to turn off signing if a default signer is defined in /etc/atomic.conf like the comment above says

Atomic/pull.py
baude: ok def strip_port() now.

Atomic/pull.py
baude: nope, it is defined differently by skopeo than the others.

Atomic/util.py
baude: yeah, took your suggestion.

Atomic/pull.py
rhatdan: @runcom can you fix this?  Should follow the URI format.

Atomic/push.py
rhatdan: Ok, now I get it.

Atomic/sign.py
baude: @rhatdan I will push a new update here shortly.  I have removed all the named args that overlap with self.args between push and sign.  That only leaves images and in_signature_path which are not in push.

Atomic/pull.py
runcom: @rhatdan we stripped the double slash long time ago when designing system containers IIRC, any reason we want it again? @mtrmac do you remember by any chance?

rhatdan:
@rh-atomic-bot r+

Atomic/pull.py
rhatdan: Just seems weird not to have it.

mtrmac: The original motivation was simply that the `//` is too much mandatory=meaningless noise and extra typing.

Strictly speaking URI format is schema:schema-specific something; `//` is used only when the part starts with an “authority” (server name/address), and our original design didn’t have any hostnames in `atomic:`.  Now that `docker:` and `atomic:` use the same docker API endpoint + repo etc. semantics, being consistent about `//` would make more sense…

… but then I’ve been tempted for a long time to make the `//` in `docker://` optional :)

mtrmac: (… actually, the shorthand notation like `docker://busybox` is violating the rules.)

rhatdan: Does this mean it will break?  I don't care as long as they are consistent.

mtrmac: We would just accept but not require the `//`. No existing users should break.

cgwalters:
Note this required http://pkgs.fedoraproject.org/cgit/rpms/atomic.git/commit/?id=099aba986bf53edb082871db7c9519f67710b4c5

PR #633, opened 2016-09-19
[merged] default.yaml: Define a default yaml for signing.

baude:
The default.yaml is useful for out of the box
experience.  It allows the user to begin signing
once atomic is install (and they have the proper
gpg requirements).

Note the .spec file will need to be updated to
handle /etc/containers/registries.d/default.yaml.

### DIFF
--- a/default.yaml
+++ b/default.yaml
@@ -1,0 +1,26 @@
+# This is a default registries.d configuration file.  You may
+# add to this file or create additional files in registries.d/.
+#
+# sigstore: indicates a location that is read and write
+# sigstore-staging: indicates a location that is only for write
+#
+# sigstore and sigstore-staging take a value of the following:
+#   sigstore:  {schema}://location
+#
+# For reading signatures, schema may be http, https, or file.
+# For writing signatures, schema may only be file.
+
+# This is the default signature write location for registries.
+default-docker:
+#  sigstore: file:///var/lib/atomic/sigstore
+  sigstore-staging: file:///var/lib/atomic/sigstore
+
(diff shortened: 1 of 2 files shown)


default.yaml
```diff
@@ -0,0 +1,26 @@
...
+# sigstore and sigstore-staging take a value of the following:
+#   sigstore:  {schema}://location
+#
+# For reading signatures, schema may be http, https, or file.
+$ For writing signatures, schema may only be file.
+
+# This is the default signature write location for registries.
+default-docker:
+#  sigstore: file:///var/lib/atomic/sigstore
+sigstore-staging: file:///var/lib/atomic/sigstore
```
aweiteka: indent

default.yaml
```diff
@@ -0,0 +1,26 @@
...
+# add to this file or create additional files in registries.d/.
+#
+# sigstore: indicates a location that is read and write
+# sigstore-staging: indicates a location that is only for write
+#
+# sigstore and sigstore-staging take a value of the following:
+#   sigstore:  {schema}://location
+#
+# For reading signatures, schema may be http, https, or file.
+$ For writing signatures, schema may only be file.
```
aweiteka: typo: #

baude: good catch.

default.yaml
baude: also good catch

rhatdan:
@aweiteka I think we need the english cleaned up here to be more descriptive.  But I will merge so that we can move forward.

rhatdan:
@rh-atomic-bot r+

PR #635, opened 2016-09-20
[merged] Add additional bash completions for push/pull

rhatdan:


rhatdan:
I am just going to merge now so that we can get out another release for testing.  This patch just effects completions, and a help message.

@rh-atomic-bot r+

PR #637, opened 2016-09-20
[merged] atomic.conf: Set docker as default backend

baude:
Setting docker as the default backend instead of ostree.

### DIFF
--- a/Atomic/pull.py
+++ b/Atomic/pull.py
@@ -11,7 +11,7 @@ ATOMIC_CONFIG = get_atomic_config()
 
 def cli(subparser):
     # atomic pull
-    backend = ATOMIC_CONFIG.get('default_storage', "ostree")
+    backend = ATOMIC_CONFIG.get('default_storage', "docker")
     pullp = subparser.add_parser("pull", help=_("pull latest image from a repository"),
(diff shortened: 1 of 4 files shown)


baude:
@rhatdan please eval and merge

rhatdan:
Can you cleanup the atomic pull man page also.  Looks like it needs fixes.

rhatdan:
Use the --ostree option to store it into the OSTree repository.
should be

       Use the --storage ostree option to store it into the OSTree repository.

rhatdan:
atomic pull [-h|--help] [--storage=[ostree|docker]] IMAGE

rhatdan:
All pull examples should be --storage ostree.  Also needs mention of the conf file.

baude:
Ok and added the -t|--type for atomic registries too.

rhatdan:
I think the systemcontainers tests might break with this changes also.

rhatdan:
We also need to expand some of the man pages to show non ostree pulls.  Perhaps something with trust, breaking.

rhatdan:
Otherwise LGTM.

baude:
updated...

rhatdan:
@rh-atomic-bot r+

PR #642, opened 2016-09-20
[merged] Fix bug in sigstore discovery

aweiteka:
Resolve KeyError if optional `sigstore-type` LABEL doesn't exist


rhatdan:
@rh-atomic-bot r+

PR #643, opened 2016-09-21
[merged] add keyring override config

aweiteka:


aweiteka:
This config value will override the GNUPGHOME env var. I *think* that's what we want. If not I could add a check to use it if defined.

@baude PTAL

mtrmac:
Is “keyring” the right terminology? Looking at `man gpg2`, it seems to use “keyring” to mean “a file containing keys”; `~/.gnupg` is called a "(GnuPG) home directory”, e.g.:
> ```
> --keyring file
>               Add file to the current list of keyrings. If file begins with a tilde and a slash, these are replaced by the $HOME directory. If the filename does not contain a slash, it is assumed  to  be  in
>               the GnuPG home directory ("~/.gnupg" if --homedir or $GNUPGHOME is not used).
>  
>               Note that this adds a keyring to the current list. If the intent is to use the specified keyring alone, use --keyring along with --no-default-keyring.
>  
>        --secret-keyring file
>               This is an obsolete option and ignored.  All secret keys are stored in the ‘private-keys-v1.d’ directory below the GnuPG home directory.
```

**Comment by rhatdan** [comment]
We need to get this right, so lets come to agreement on the option names.

**Comment by aweiteka** [comment]
> Is “keyring” the right terminology?

How about `gnupg_homedir: /home/USER/.gnupg`?

**Comment by mtrmac** [comment]
> How about gnupg_homedir: /home/USER/.gnupg?

Works for me.

**Comment by baude** [comment]
So if a default keyring is set, how does the user override that?  I think there should be a --something.

File: Atomic/sign.py [hunk 1]
```diff
@@ -57,6 +64,17 @@ def no_reg_no_default_error(image, registry_path):
         registry_config_path = '/etc/containers/registries.d' if registry_config_path is None else registry_config_path
         registry_configs, default_store = util.get_registry_configs(registry_config_path)
 
+        # we honor GNUPGHOME if set, override with atomic.conf, arg overrides all
+        gpghomedir =  None
+        if "GNUPGHOME" in os.environ:
+            gpghomedir = os.environ['GNUPGHOME']
```
rhatdan: No reason for this check.

Atomic/sign.py
```diff
@@ -57,6 +64,17 @@ def no_reg_no_default_error(image, registry_path):
...
+        if util.get_atomic_config_item(['gnupg_homedir']):
```
rhatdan: I think you can simplify this to:
```
        if self.args.gnupghome:
               gpghomedir = self.args.gnupghome
        else:
              gpghomedir = util.get_atomic_config_item(['gnupg_homedir'])
```

docs/atomic-sign.1.md
```diff
@@ -30,12 +31,14 @@ in /etc/atomic.conf.
...
 **-d** **--directory**
   Store the signatures in the specified directory.  Default: /var/lib/atomic/signature
- 
 
 **--sign-by**
   Override the default identity of the signature. You can define a default in /etc/atomic.conf
   with the key **default_signer**.
 
+**-g** **--gnupghome**
+  Specify the GNUPGHOME directory to use for signing, e.g. &#126;/.gnupg. This
```
rhatdan: Does this display correctly in the man page with the &#126;?/.gnupg

aweiteka: This one does because it's not indented

docs/atomic-sign.1.md
```diff
@@ -46,9 +49,9 @@ Sign the foobar image and save the signature in /tmp/signatures/.
 
     atomic sign -d /tmp/signatures privateregistry.example.com
 
-Sign the busybox image with the identify of foo@bar.com
+Sign the busybox image with the identify of foo@bar.com with a user's keyring
 
-    atomic sign --sign-by foo@bar.com privateregistry.example.com
+   sudo atomic sign --sign-by foo@bar.com --gnupghome=&#126;/.gnupg privateregistry.example.com
```
aweiteka: @rhatdan this does not display if I indent properly.

docs/atomic-sign.1.md
```diff
@@ -80,6 +80,12 @@ Now any image from the john repository will use the sigstore-staging location of
 means that signatures should be written to that location but read should occur from
 the http URL provided.
 
+The user's keyring will be used during signing. When running as root user this may
+not be desired. Another keyring may be specified gusing environment variable GNUPGHOME
+or set in configuration file atomic.conf. For example:
+
+default_keyring: /home/USER/.gnupg
```
aweiteka: In my experience the `~` only expands to the current user at the shell. If used here it expands to the root users dir, not user who called sudo.

Atomic/sign.py
```diff
@@ -57,6 +57,9 @@ def no_reg_no_default_error(image, registry_path):
         registry_config_path = '/etc/containers/registries.d' if registry_config_path is None else registry_config_path
         registry_configs, default_store = util.get_registry_configs(registry_config_path)
 
+        if util.get_atomic_config_item(['default_keyring']):
+            os.environ['GNUPGHOME'] = util.get_atomic_config_item(['default_keyring'])
+            print("using keyring", os.environ['GNUPGHOME'])
```
aweiteka: Whoops. neither do I good catch.

docs/atomic-sign.1.md
rhatdan: OK this is fine then since it is just an example.

docs/atomic-sign.1.md
rhatdan: If you just put in ~/.gnupg, what happens?

aweiteka: It replaces ~ with space.

aweiteka:
@rhatdan updated per comments.

rhatdan:
LGTM Merge your commits and I will merge.

aweiteka:
@rhatdan done

PR #650, opened 2016-09-21
[merged] Atomic/pull.py: Fix logic error in pull

baude:
When a image without a registry is provided as input, i.e.

atomic pull busybox

We are not able to resolve that reference now because we can
no longer depend on docker's algo's for that as we now
use skopeo. Therefore, we now take the input, see if skopeo
inspect can resolve it; and then decompose the fqdn so we
can check if the registry is secure.  Had to make a small
change on the return types of util.decompose to ensure
it returns str objects or else it will return unicode in
python2.

### DIFF
--- a/Atomic/pull.py
+++ b/Atomic/pull.py
@@ -26,8 +26,7 @@ def cli(subparser):
 
 class Pull(Atomic):
     def pull_docker_image(self):
-        registry, _, tag = decompose(self.args.image)
-        insecure = True if is_insecure_registry(self.d.info()['RegistryConfig'], strip_port(registry)) else False
+        _, _, tag = decompose(self.args.image)
         # If no tag is given, we assume "latest"
         tag = tag if tag != "" else "latest"
         if self.args.reg_type == "atomic":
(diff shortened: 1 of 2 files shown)


baude:
rebased... @rhatdan please review ... no real code changes, just logic.

baude:
@rhatdan hold on this.

baude:
ok should be good.

Atomic/pull.py
```diff
@@ -27,15 +27,16 @@ def cli(subparser):
 class Pull(Atomic):
     def pull_docker_image(self):
         registry, _, tag = decompose(self.args.image)
```
rhatdan: No longer use this registry, should probably be _

Atomic/util.py
```diff
@@ -62,7 +62,7 @@ def decompose(compound_name):
         reg, repo = repo.split('/', 1)
     if ':' in repo:
         repo, tag = repo.rsplit(':', 1)
-    return reg, repo, tag
+    return str(reg), str(repo), str(tag)
```
rhatdan: Why is this needed and will it work on both python2 and python3

baude: in python2,  the skopeo outputs are all unicode.

Atomic/pull.py
baude: indeed, will fix that.

Atomic/util.py
baude: prefer i type convert this prior to calling decompose?

rhatdan: No I just wanted to make sure you tested it on both pythons.  Code like this can sometimes trigger other issues.

rhatdan:
@rh-atomic-bot r+

PR #651, opened 2016-09-21
[merged] Start/stop syscontainers with atomic run/stop

yuqi-zhang:
Wrap atomic run/stop with systemctl start/stop for system
containers. This way the user can directly use the atomic
CLI for the full container cycle. In addition, the equivalent
functionality already exists for docker containers, so it
makes sense to allow system containers to start/stop in a
similar fashion.


Atomic/syscontainers.py
```diff
@@ -677,6 +677,18 @@ def _is_service_failed(self, name):
             else:
                 return False
 
+    def start_service(self, name):
+        try:
+            self._systemctl_command("start", name)
+        except subprocess.CalledProcessError as e:
+            util.write_out(e.output)
```
rhatdan: If this is an error you should raise ValueError(e.output)

Atomic/syscontainers.py
```diff
@@ -677,6 +677,18 @@ def _is_service_failed(self, name):
...
+
+    def stop_service(self, name):
+        try:
+            self._systemctl_command("stop", name)
+        except subprocess.CalledProcessError as e:
+            util.write_out(e.output)
```
rhatdan: If this is an error you should raise ValueError(e.output)

yuqi-zhang:
Will fix the error

Atomic/syscontainers.py
yuqi-zhang: Changed, thanks!

rhatdan:
@rh-atomic-bot r+

PR #654, opened 2016-09-22
[merged] Missing paths are causing failures in running atomic in cron job

rhatdan:
We need to fully path executables in order for atomic command to
work with any $PATH set.

Basically cron jobs does not include /sbin in PATH.


rhatdan:
@baude PTAL

rhatdan:
Should fix https://bugzilla.redhat.com/show_bug.cgi?id=1378091

Atomic/mount.py
```diff
@@ -36,6 +36,11 @@
 
 # Module for mounting and unmounting containerized applications.
 
+DMSETUP_PATH='/usr/sbin/dmsetup'
+LSBLK_PATH='/usr/bin/lsblk'
+MOUNT_PATH='/usr/bin/mount'
+FINDMNT_PATH='/usr/bin/findmnt'
+
```
baude: What  happens when the paths are different?

rhatdan: Where would the paths be different?

baude: I was just thinking of Fedora vs RHEL vs Centos things.

rhatdan: They are in the same location in all of these cases. There could be a problem if other distros picked up the atomic command like debian, maybe.  But for now I think we should fully path these.

rhatdan:
@baude OK To merge?

rhatdan:
@rh-atomic-bot r+

PR #656, opened 2016-09-22
[merged] Update to error message and install

ddarrah:
2 changes:
- allow pull to print a human readable error if the local docker daemon isn't running
- have install create the /var/lib/atomic/sigstore directory so that default signing config will work out of box


rhatdan:
@rh-atomic-bot r+

Issue #653, opened 2016-09-22
Update trust default policy so source type "docker-daemon"  is not rejected


aweiteka:
See discussion here: https://github.com/projectatomic/skopeo/issues/212

rhatdan:
Do we need a pull request?  Isn't this something we are shipping now?

aweiteka:
#658 
Can't merge yet; pending skopeo.

mtrmac:
Isn’t the policy actually shipped in the skopeo package?

aweiteka:
@mtrmac based on comments I think we'd like to ship with an insecure docker-daemon transport type. I would then close this PR and request that from skopeo. Thoughts?

mtrmac:
Yeah, I can put that on the `docker-daemon` skopeo branch for now.

rhatdan:
Looks like this is merged.

Fixed in atomic-1.13

PR #658, opened 2016-09-22
WIP: add docker-daemon accept when default reject

aweiteka:
Pending support in skopeo for docker-daemon type

### DIFF
--- a/Atomic/trust.py
+++ b/Atomic/trust.py
@@ -167,6 +167,11 @@ class Trust(Atomic):
                 policy["default"][0]["type"] = default_type_map[self.args.default_policy]
             else:
                 policy["default"] = [ { "type": default_type_map[self.args.default_policy] }]
+            if default_type_map[self.args.default_policy] == "reject":
+                if not "transports" in policy:
+                    policy["transports"] = {}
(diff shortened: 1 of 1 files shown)


Atomic/trust.py
```diff
@@ -167,6 +167,11 @@ def modify_default(self):
                 policy["default"][0]["type"] = default_type_map[self.args.default_policy]
             else:
                 policy["default"] = [ { "type": default_type_map[self.args.default_policy] }]
+            if default_type_map[self.args.default_policy] == "reject":
+                if not "transports" in policy:
+                    policy["transports"] = {}
+                # we don't want to reject push from docker-daemon
+                policy["transports"]["docker-daemon"] = { "": [{"type":"insecureAcceptAnything"}] }
```
rhatdan: This would override whatever the user added for this.  I think we should just make this the default and put it in our policy.conf file. Then if the user changes the default, it will stay with whatever the user changed it to.

aweiteka: Good point. I think that's a safe default for any case I can imagine.

rhatdan:
Any update on this?

aweiteka:
@rhatdan see #653

PR #662, opened 2016-09-23
[merged] Atomic/push.py: Rely on skopeo to write signatures

baude:
atomic push should always rely on skopeo to write
signatures even when the signatures are being
written locally.


baude:
@rh-atomic-bot r+

PR #663, opened 2016-09-23
[merged] Pass down default policy_filename to skopeo command

rhatdan:


rhatdan:
@aweiteka PTAL

aweiteka:
I think you're going to need to pass in the filename everywhere we call skopeo and add it to the cmd string: skopeo_copy(), skopeo_delete(), skopeo_standalone_sign(), etc.

rhatdan:
Ok I am passing it for skopeo_inspect and skopeo_copy now.

rhatdan:
@aweiteka I added it to skopeo and inspect, but the tests are failing because skopeo --policy is not supported in the version in the test suite.

baude:
What is the background for this? I dont understand why a policy file has anything to do with skopeo inspect.  Can you elaborate?

Atomic/util.py
```diff
@@ -33,6 +33,7 @@
 BWRAP_OCI_PATH = "/usr/bin/bwrap-oci"
 RUNC_PATH = "/bin/runc"
 SKOPEO_PATH = "/usr/bin/skopeo"
+POLICY_FILENAME = "/etc/containers/policy.conf"
```
baude: I think this should be policy.json

rhatdan: The reason we want to pass this down is for testing.  I agree, I am not sure if skopeo inspect requires it either.  I think it is only used for skopeo copy.  Although --policy is an option for skopeo, rather then just an option for skopeo copy,  @runcom and/or @mtrmac do any of the other skopeo calls use the `--policy` option?

runcom: > do any of the other skopeo calls use the --policy option?

@rhatdan nope, those options can be moved under `copy` probably

rhatdan: Ok I now only do it for `skopeo copy`

PTAL

aweiteka: @runcom would 'standalone-sign' touch policy?

runcom: > @runcom would 'standalone-sign' touch policy?

@aweiteka nope, as far as I can tell from looking at skopeo code, @mtrmac ?

mtrmac: `skopeo standalone-sign` doesn’t use `--policy`, and won’t.  I wouldn’t _entirely_ rule out `skopeo inspect`, and especially possible future commands, using it in the future—but that would be definitely discussed and coordinated, not a silent change.

Atomic/util.py
```diff
@@ -341,8 +342,12 @@ def skopeo_manifest_digest(manifest_file, debug=False):
...
     return check_output(cmd).rstrip().decode()
 
-def skopeo_copy(source, destination, debug=False, sign_by=None, insecure=False):
-    cmd = [SKOPEO_PATH]
+def skopeo_copy(source, destination, debug=False, sign_by=None, insecure=False, policy_filename=None):
+
+    if not policy_filename:
+        policy_filename=POLICY_FILENAME
+
+    cmd = [SKOPEO_PATH, "--policy=%s" % policy_filename]
```
mtrmac: I’d be a bit happier if `atomic` were only overriding `--policy` if it had a _non-default_ value; this would minimize the risk of `skopeo` and `atomic` accidentally using a different path (see the `policy.conf` typo which happened earlier in this PR.)

rhatdan: @mtrmac I changed it to only add flag if policy_file !=None.
PTAL

Atomic/util.py
```diff
@@ -33,6 +33,7 @@
 BWRAP_OCI_PATH = "/usr/bin/bwrap-oci"
 RUNC_PATH = "/bin/runc"
 SKOPEO_PATH = "/usr/bin/skopeo"
+POLICY_FILENAME = "/etc/containers/policy.json"
```
mtrmac: This is now unused?

rhatdan:
@mtrmac @aweiteka needs this for some tests where he will use alternate policy files.

aweiteka:
@rhatdan LGTM

Atomic/util.py
```diff
@@ -341,8 +341,12 @@ def skopeo_manifest_digest(manifest_file, debug=False):
     cmd = cmd  + ['manifest-digest', manifest_file]
     return check_output(cmd).rstrip().decode()
 
-def skopeo_copy(source, destination, debug=False, sign_by=None, insecure=False):
+def skopeo_copy(source, destination, debug=False, sign_by=None, insecure=False, policy_filename=None):
+
     cmd = [SKOPEO_PATH]
+    if not policy_filename:
```
mtrmac: Shouldn’t this be `if policy_filename is not None`?

mtrmac:
LGTM except for the `None` check pointed out above.

Atomic/util.py
rhatdan: Is there any difference between those two lines?

rhatdan: Oops should be just 
if policy_filename

rhatdan:
@rh-atomic-bot r+

PR #671, opened 2016-09-30
[merged] Code cleanup for atomic sign

rhatdan:
We need to default the gnukeyring to the users login directory. To
make it easier for the user.

Also add bash completions for -g option


rhatdan:
@baude @aweiteka PTAL

baude:
Whats happens if the gpg is owned by root/sudo?

rhatdan:
The the user could specify it or put it in atomic.conf.  The most common path, I believe is the users gnugpg will be in the users homedir, so we should default to that.

rhatdan:
@rh-atomic-bot r+

PR #673, opened 2016-10-03
[merged] Atomic/sign.py: Input image must have registry name

baude:
Because we only intend atomic sign to create local signatures
of remote images, all input image names must have a registry
or we now fail.


baude:
@aweiteka @rhatdan per our conversation this morning.

rhatdan:
@rh-atomic-bot r+

PR #675, opened 2016-10-04
[merged] Atomic/push.py: Don't prompt for password if token found

baude:
If the user has a token for the registry, we assume the token
is good and do not prompt for a username or password.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -67,6 +67,7 @@ class Atomic(object):
         self.syscontainers = SystemContainers()
         self.run_opts = None
         self.atomic_config = util.get_atomic_config()
+        self.local_tokens = {}
 
     def __enter__(self):
         return self
@@ -677,6 +678,28 @@ class Atomic(object):
             return vuln_ids
         except IOError:
             return []
+
+    def get_local_tokens(self):
+        if len(self.local_tokens) < 1:
+            self.local_tokens = self.load_local_tokens()
+        return self.local_tokens
+
+    @staticmethod
+    def load_local_tokens():
+        tokens = {}
+        token_file_name = os.path.expanduser('~/.docker/config.json')
+        if not os.path.exists(token_file_name):
+            return {}
+        with open(token_file_name) as token_file:
+            token_data = json.load(token_file)
+        try:
+            for registry in token_data['auths']:
+                tokens[registry] = token_data['auths'][registry]['auth']
+        except KeyError:
+            # Just return a blank dict
+            pass
+        return tokens
+
 
 class AtomicError(Exception):
     pass
--- a/Atomic/push.py
+++ b/Atomic/push.py
@@ -89,6 +89,13 @@ class Push(Atomic):
         self.policy_filename=policy_filename
 
     def push(self):
+        def prompt():
+            if not self.args.username:
+                self.args.username = util.input("Registry Username: ")
+
+            if not self.args.password:
+                self.args.password = getpass.getpass("Registry Password: ")
+
         self.ping()
         if self.args.debug:
             util.write_out(str(self.args))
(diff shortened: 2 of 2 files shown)


Atomic/atomic.py
```diff
@@ -678,6 +679,24 @@ def get_vulnerable_ids(self):
...
 
+    def get_local_tokens(self):
+        if len(self.local_tokens) < 1:
+            self.local_tokens = self.load_local_tokens()
+        return self.local_tokens
+
+    @staticmethod
+    def load_local_tokens():
+        tokens = {}
+        token_file_name = os.path.expanduser('~/.docker/config.json')
```
rhatdan: @runcom Does Skopeo just do this, or is it a little more cleaver?  Docker?  Does it look at SUDO_USER?

Atomic/atomic.py
```diff
@@ -678,6 +679,24 @@ def get_vulnerable_ids(self):
...
+        if not os.path.exists(token_file_name):
+            return {}
+        with open(token_file_name) as token_file:
+            token_data = json.load(token_file)
+        for registry in token_data['auths']:
+            tokens[registry] = token_data['auths'][registry]['auth']
```
rhatdan: Should this be in a try block?  Potential errors in the auth.  If the auths and auth do not exists should we return {}

Atomic/atomic.py
```diff
@@ -678,6 +679,24 @@ def get_vulnerable_ids(self):
...
+        return tokens
+
+
 class AtomicError(Exception):
```
rhatdan: Remove extra space

Atomic/push.py
```diff
@@ -122,24 +129,29 @@ def push(self):
...
+
         if self.args.pulp:
             return pulp.push_image_to_pulp(self.image, self.args.url,
                                            self.args.username,
                                            self.args.password,
                                            self.args.verify_ssl,
                                            self.d)
 
-        if self.args.satellite:
+        elif self.args.satellite:
```
rhatdan: Why else? Previous if returned.

Atomic/push.py
```diff
@@ -122,24 +129,29 @@ def push(self):
...
+        sign = True if self.args.sign_by else False
+
+        registry, _, _ = util.decompose(self.image)
+        if sign and not self.args.satellite and not self.args.pulp:
+            # Check if any local tokens exist
+            if registry not in [x for x in self.load_local_tokens()]:
+                # If we find a token for the registry, we don't
+                # prompt for a username or password
+                prompt()
+
```
rhatdan: Why not have this block to after the satellite and pulp checks are run.

Atomic/atomic.py
baude: Added try block for a KeyError exception

Atomic/atomic.py
rhatdan: ok

Atomic/push.py
```diff
@@ -156,6 +160,13 @@ def push(self):
 
         else:
             reg, _, tag = util.decompose(self.image)
+            if sign and not self.args.satellite and not self.args.pulp:
```
rhatdan: No need for satellite or pulp check.
No need for else: above either.

Atomic/push.py
```diff
@@ -156,6 +160,13 @@ def push(self):
...
+                # Check if any local tokens exist
+                if reg not in [x for x in self.load_local_tokens()]:
```
rhatdan: Shouldn't this be get_local_tokens?

Atomic/push.py
baude: Geez, sloppy on my part.  Fixed.

Atomic/push.py
baude: yup!  fixed.

PR #676, opened 2016-10-04
[merged] Remove primary commands and move to images subcommand

rhatdan:
In order to cleanup code and simplify interface, i
move help, info, verify and version to the images subcommand.

Remove man pages for these atomic commands and add info to images subcommand.

Since these have been previously documented we will continue to support
atomic help
atomic info
atomic verify
atomic version

But we will not document them.


rhatdan:
@cgwalters @baude @aweiteka @giuseppe PTAL

baude:
I'm struggling with this a little.  I see the connection between these subcommands and 'image'.  But in that case, why not install, pull, push, sign, etc? else lgtm

rhatdan:
I considered them but they are more fundamental then the four commands I removed.  I am just trying to handle the explosion of primary commands.  These commands are minor and not used by many, so moving them seemed like a good idea. 

atomic sign is also questionable as a primary command since you are signing images.

Bottom line atomic was originally built for atomic install/run/uninstall work flow.  Keeping these primary makes sense.  
In the future we need to pay attention to new primary verbs.

rhatdan:
BTW I also want to consider merging pull and update.  They even have the same description.

rhatdan:
@baude Please review the code. We can debate at a later time whether there are other commands we can remove

baude:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #679, opened 2016-10-04
[merged] version: save and display system image info

yuqi-zhang:
Save labels from docker manifest into ostree, so when a user invokes
'atomic version' on a system image, the metadata info is displayed,
same as docker images.


yuqi-zhang:
Bug: https://bugzilla.redhat.com//show_bug.cgi?id=1377749

Note that due to how the refs are now saved (with "_2" and "_3") in ostree, the bash completion for __atomic_system_containers_images no longer works. I will look into it in a separate patch.

rhatdan:
@giuseppe PTAL

giuseppe:
@rh-atomic-bot r+ 9a60542a3af11cc35919c5d575cbb0a72092a103

PR #680, opened 2016-10-05
[merged] Syscontainers fix scan

giuseppe:
syscontainers: inspect and checkout accept image Ids
    
If the image Id is specified, look it up in the available images metadata.


rhatdan:
LGTM
@rh-atomic-bot r+

PR #681, opened 2016-10-05
[merged] images: fix system images auto-complete

yuqi-zhang:
Image names are stored differently in ostree, and thus cannot be
directly used as we used to.


rhatdan:
@rh-atomic-bot r+

PR #684, opened 2016-10-06
[merged] fix bugs and migrate to .redhat-ci.yml

jlebon:
This patch add a new `.redhat-ci.yml` file. This file is read by the
internal Red Hat CI service to determine testing procedures. For more
information, see:

https://github.com/jlebon/redhat-ci

---

:warning: Do not merge yet. The backend is not yet completely hooked up. Just opening the PR ahead of time to get folks familiar with it. Once this is merged, I will be retiring the current PR tester. :warning:

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -1,0 +1,19 @@
+branches:
+  - master
+
+host:
+  distro: fedora/24/atomic
+
+tests:
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1318547#c7
+  - mount --make-rshared /
+  - systemctl start docker
+  - ostree admin unlock
+  - >
+      docker run --privileged
+      -v /root:/root -v $PWD:/code -v /:/host
+      --workdir /code fedora:24 ./rhci.sh
+  - PYTHON=python3 ./test.sh
+
+artifacts:
+  - tests.log
--- a/Atomic/help.py
+++ b/Atomic/help.py
@@ -38,7 +38,7 @@ class AtomicHelp(Atomic):
 
     def help_tty(self):
         result = self.help()
-        if sys.stdout.isatty():
+        if not sys.stdout.isatty():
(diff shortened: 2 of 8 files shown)


jlebon:
@aweiteka @runcom This could also provide a way to support more recent `skopeo` builds for testing. The YAML file supports an `extra-repos` [key](https://github.com/jlebon/redhat-ci/blob/master/sample.redhat-ci.yml#L30), so e.g. if someone sets up a COPR, it would be trivial to hook up here.

There's also the CAHC skopeo build, but there's the epoch mismatch issue, and I'm not sure if all the test requirements are available on CentOS, though you're free to explore that area as well.

.redhat-ci.yml
```diff
@@ -0,0 +1,43 @@
+host:
+  distro: fedora/24/cloud
```
cgwalters: Same complaint as https://github.com/projectatomic/commissaire/pull/16#pullrequestreview-3143852 :smile:

jlebon:
Bug fixes and full migration to Atomic Host tests. ^

rhatdan:
LGTM

jlebon:
Could we review & merge this soon? I'd like to get it in before things get out of sync and start gating PRs on the AH tests.

cgwalters:
Long term, I think we should split the test suite into both traditional "make check" (uid != 0, uninstalled), and "destructive system tests".  

The groff etc would run under `make check` just fine in a container.  The destructive system tests would run docker.

cgwalters:
@rh-atomic-bot r+ d7024bb

jlebon:
Right, in fact as it stands right now, the whole testsuite could probably be run from a docker container with the socket mounted, except for the migrate and storage tests. Shouldn't be too hard to modify.

I think I'll leave the previous PR tester since it runs on Python 2. At least until we make redhat-ci support multiple testsuites. In the meantime, can we make homu gate on the new context too?

PR #687, opened 2016-10-07
[merged] Introduce registry inspect methods

baude:
Add two classes and several new methods for inspecting
registries and images.  These methods are helpful
for:

* taking user input (i.e. image names, partial image names) and determining
the fully qualified names.
* determining the fq name when the registry is omitted by the user.
* obtaining image manifests
* mimicing skopeo inspect

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -17,6 +17,7 @@ from . import util
 import re
 from .util import NoDockerDaemon, DockerObjectNotFound
 from docker.errors import NotFound
+from .discovery import RegistryInspect
 
 def find_repo_tag(d, Id, image_name):
     def image_in_repotags(image_name, repotags):
@@ -109,9 +110,10 @@ class Atomic(object):
         self.ping()
         if self.force:
             self.force_delete_containers()
-        registry, _, _, _ = util.decompose(self.image)
+        fq_name = self.get_fq_image_name(self.image)
+        registry, _, _, _ = util.decompose(fq_name)
         return util.skopeo_copy("docker://{}".format(self.image),
-                                "docker-daemon:{}".format(self.image),
+                                "docker-daemon:{}".format(fq_name),
                                 util.is_insecure_registry(self.d.info()['RegistryConfig'], util.strip_port(registry)))
 
     def pull(self):
@@ -699,6 +701,16 @@ class Atomic(object):
             # Just return a blank dict
             pass
         return tokens
+
+    def get_fq_image_name(self, input_image):
+        registry, repo, image, tag = util.decompose(input_image)
+        if not image:
+            raise ValueError('Error parsing input: "{}" invalid'.format(input_image))
+        if all([True if x else False for x in [registry, image, tag]]):
+            return input_image
+        if not registry:
+            ri = RegistryInspect(registry, repo, image, tag, debug=self.args.debug, orig_input=self.image)
+            return ri.find_image_on_registry()
 
 
 class AtomicError(Exception):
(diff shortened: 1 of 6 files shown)


Atomic/atomic.py
```diff
@@ -700,6 +702,18 @@ def load_local_tokens():
             pass
         return tokens
 
+    def get_fq_image_name(self, input_image):
+        registry, repo, image, tag = util.decompose(input_image)
+        if not image:
+            raise ValueError('Error parsing input: "{}" invalid'.format(input_image))
+        if all([True if x else False for x in [registry, repo, image, tag]]):
+            return input_image
+        if all([True if x else False for x in [registry, image, tag]]):
```
rhatdan: Why do the first one?  Doing just this block should be enough?

rhatdan:
@runcom PTAL. Since he is doing some of the stuff skopeo does for remote viewing...

rhatdan:
Are you handling the case where there are multiple registires?

atomic pull rhel7tools

Does this work there?

Atomic/atomic.py
baude: Yup! Fixed.

tests/unit/test_discovery.py
```diff
@@ -0,0 +1,36 @@
+import unittest
+from Atomic import util
+from Atomic import discovery
+
+
+class TestAtomicUtil(unittest.TestCase):
+    IMAGE = 'docker.io/library/busybox:latest'
```
runcom: is this working with a digest? e.g. docker.io/library/busybox@sha256:%HEX%

runcom:
patch LGTM

> Are you handling the case where there are multiple registires?
> 
> atomic pull rhel7tools
> 
> Does this work there?

The loop over the available registries in docker should be here https://github.com/projectatomic/atomic/pull/687/files#diff-1e7bfb73126e6f7e89b71bbe478132d3R249 right Brent?

baude:
yeah that is the loop.  I was trying to understand if that is what @rhatdan wanted or something different.  thanks for the review @runcom

rhatdan:
Right as long as rhel7tools ends up being something like 
redhat-registry.redhat.com/rhel7tools:latest

I am fine with this patch.

baude:
Current output for this looks like:

```
[bbaude@localhost atomic (inspect)]$ sudo ./atomic pull rhel7
Image rhel7 is being pulled to docker ...
Trying docker.io/library/rhel7:latest
Trying registry.access.redhat.com/rhel7:latest
Pulling registry.access.redhat.com/rhel7:latest ...
Getting image source manifest
Getting image source configuration
Uploading blob sha256:30cf2e26a24f2a8426cbe8444f8af2ecb7023bd468b05c1b6fd0b2797b0f9ff9
 5.97 MB / ? [-----------------------------
```

Atomic/discovery.py
```diff
@@ -0,0 +1,406 @@
...
+import hashlib
+
+
+try:
+    from urllib.request import parse_http_list, parse_keqv_list
+except ImportError:
+    from urllib2 import parse_http_list, parse_keqv_list #pylint: disable=import-error
+
+
+def jose_base64_urldecode(in_str):
```
cgwalters: Who's "jose"?  :smile:   Did this code come from somewhere else?

baude:
At the risk of merging my own code ... but doing so to unblock CAHC

@rh-atomic-bot r+

Atomic/discovery.py
baude: @cgwalters it mimics some golang code

PR #692, opened 2016-10-10
[merged] atomic.d/openscap: Fix race condition (bz #1368896)

baude:
There is a race condition in oscpd where it sometimes fails to scan
because of a threading issue.  While that is resolved upstream, we
set the max number of threads to 1 to avoid it.

This resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1368896

### DIFF
--- a/atomic.d/openscap
+++ b/atomic.d/openscap
@@ -5,10 +5,10 @@ default_scan: cve
 custom_args: ['-v', '/etc/oscapd:/etc/oscapd:ro']
 scans: [ 
       { name: cve,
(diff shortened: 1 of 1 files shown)


rhatdan:
Will this slow down our scans?  Will we remove it when upstream becomes downstream?

baude:
Yes it will slow down scans of more than one image.  When Martin has oscapd fixed, we can remove the single threading as well.

rhatdan:
@rh-atomic-bot r+

Issue #647, opened 2016-09-21
trust delete should have a "rm sigstore" option


aweiteka:
Deleting trust removes it from the policy.json but leaves the registries.d config around. That's the correct default behavior but we should allow a user to cleanup the environment.

Not sure what to call this boolean flag...
```
$ atomic trust delete POLICY_NAME --sigstore
```

rhatdan:
--remove?
```
man userdel
...
       -r, --remove
           Files in the user's home directory will be removed along with the home directory itself and the user's mail spool. Files located in
           other file systems will have to be searched for and deleted manually.

           The mail spool is defined by the MAIL_DIR variable in the login.defs file.
```
Seems quite similar.

rhatdan:
@aweiteka is this something you are working on, or do you want me to do it.

Where do I find the sigstore data?

aweiteka:
I'm not working on it.

This should remove the matching yaml section from a file in /etc/containers/registries.d/*. A call to util.get_registry_configs() will return the data you need.

I think this should be the **default** behavior, with an option to retain the sigstore URL, a less common use case. In other words, policy and sigstore are typically managed together IMO.

rhatdan:
Ok, just set the --remove to default to yes.  Feel free to steel the stuff I have done so far.

PR #693, opened 2016-10-10
[merged] Allow pull from registry not in docker conf

baude:
We now allow pulls from registries that are not in
the docker configuration file.  This altered our
decompose method a bit.  We now check the registry
in decompose to see if it resolves on the network.
If so, then we use it.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -707,7 +707,11 @@ class Atomic(object):
         if not image:
             raise ValueError('Error parsing input: "{}" invalid'.format(input_image))
         if all([True if x else False for x in [registry, image, tag]]):
-            return input_image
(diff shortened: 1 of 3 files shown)


rhatdan:
Should we update the tests for this?

rhatdan:
@baude Can we add a pull from 
    def test_find_image_on_registry(self):
for this container image.

rhatdan:
The goal of this patch set is to satisfy https://github.com/projectatomic/atomic/issues/647

rhatdan:
Close https://github.com/projectatomic/atomic/issues/647

baude:
@rhatdan updated with tests

baude:
@rh-atomic-bot r+

PR #694, opened 2016-10-11
DONTMERGE: Add atomic trust delete --remove flag

rhatdan:
This flag will remove local signature stores.

### DIFF
--- a/Atomic/trust.py
+++ b/Atomic/trust.py
@@ -58,11 +58,13 @@ def cli(subparser):
     defaultp.set_defaults(_class=Trust, func="modify_default")
     deletep = subparsers.add_parser("delete",
                                     help="Delete a trust policy for a registry")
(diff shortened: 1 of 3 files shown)


rhatdan:
@aweiteka PTAL

I don't really know what I am removing here, but I thought I would throw together something to get this moving.

Atomic/trust.py
```diff
@@ -143,6 +145,8 @@ def delete(self):
         with open(self.policy_filename, 'r+') as policy_file:
             policy = json.load(policy_file)
             try:
+                if self.args.remove:
+                    sigstore = policy["transports"][sstype][self.args.registry]["sigstore"]
```
aweiteka: Should be calling `util.get_registry_configs()` here, then removing the yaml section from the returned filename.

PR #696, opened 2016-10-11
[merged] version: fix atomic version error

yuqi-zhang:
Add extra check to prevent errors for image IDs without repo name.
Before, when a user invokes "atomic version" on such an ID, it
returns "List index out of range" instead of label information
that exists.


Atomic/atomic.py
```diff
@@ -362,7 +362,9 @@ def gen_cmd(self, cargs):
             return " ".join(args)
 
     def get_fq_name(self, image_info):
-        if len(image_info['RepoTags']) > 1:
+        if not image_info['RepoTags']:
+            return None
+        elif len(image_info['RepoTags']) > 1:
```
rhatdan: I would prefer:
```
if image_info['RepoTags'] and len(image_info['RepoTags']) > 1:
```

yuqi-zhang: If it doesn't exist it will default to:
`else:
    return image_info['RepoTags'][0]`
Which will also cause the issue.

rhatdan: Ok no need of the else clauses.

```
        if not image_info['RepoTags']:
            return None

        if len(image_info['RepoTags']) > 1:
            if self.image in image_info['RepoTags']:
                return self.image

            raise ValueError("\n{} is tagged with multiple repositories. "
                             "Please use a repository name instead.\n".format(self.image))

        return image_info['RepoTags'][0]
```

rhatdan: Not a big fan of elif clauses.

yuqi-zhang: Sorry, updated.

rhatdan:
@rh-atomic-bot r+

PR #704, opened 2016-10-12
remove sigstore config with trust delete

### LINKED ISSUE #647: trust delete should have a "rm sigstore" option

aweiteka:
resolves #647 and #694

### DIFF
--- a/Atomic/trust.py
+++ b/Atomic/trust.py
@@ -63,6 +63,8 @@ def cli(subparser):
     deletep.add_argument("--sigstoretype", dest="sigstoretype", default="web",
                          choices=['atomic', 'local', 'web'],
                          help=sigstore_help)
+    deletep.add_argument("--save-sigstore", dest="savesigstore", default=True, action="store_false",
+                         help="Do not remove the registry sigstore configuration")
     deletep.set_defaults(_class=Trust, func="delete")
     showp = subparsers.add_parser("show",
                                   help="Display trust policy for the system")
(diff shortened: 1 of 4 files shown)


docs/atomic-trust.1.md
```diff
@@ -72,6 +72,11 @@ testing.
     **atomic**: OpenShift-based Atomic Registry API
     **local**: Local filesystem path
 
+# delete OPTIONS
+
+**--sigstore**
```
rhatdan: Default is true.

docs/atomic-trust.1.md
```diff
@@ -98,6 +103,10 @@ Remove a trust scope
 
     atomic trust delete docker.io
 
+Remove a trust scope but retain the sigstore configuration
+
+    atomic trust delete docker.io --sigstore
```
rhatdan: I think you would need to say

 atomic trust delete docker.io --sigstore=false

aweiteka: This is a boolean flag so it fails when a value is passed in.

```
$ atomic trust delete --sigstore=false
atomic trust delete: argument --sigstore: ignored explicit argument 'false'
Try 'atomic trust delete --help' for more information.
```

aweiteka: Good idea.

aweiteka:
@rhatdan all set.

Atomic/trust.py
```diff
@@ -63,6 +63,8 @@ def cli(subparser):
     deletep.add_argument("--sigstoretype", dest="sigstoretype", default="web",
                          choices=['atomic', 'local', 'web'],
                          help=sigstore_help)
+    deletep.add_argument("--save-sigstore", dest="savesigstore", default=True, action="store_false",
```
aweiteka: The only way I see to make this work is to use `default=True, action="store_false"`. Any other combination I've tested does not produce the behavior we document and expect.

PR #707, opened 2016-10-13
[merged] delete gomtree blob

jlebon:
The Makefile no longer creates it, nor installs it. We should just add a
runtime req in downstream distros.


cgwalters:
@rh-atomic-bot r+ 7ff9012

PR #712, opened 2016-10-18
[merged] .redhat-ci.yml: enable homu-related branches

jlebon:
Also test auto and try branches.

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -1,5 +1,7 @@
 branches:
   - master
+  - auto
+  - try
 
 host:
   distro: fedora/24/atomic


rhatdan:
@cgwalters PTAL

cgwalters:
@rh-atomic-bot r+ 32ec91b

cgwalters:
RE the CI failure, why are we running pip instead of using package dependencies?

jlebon:
Yeah, will probably do that. The error above is: https://github.com/eliben/pycparser/issues/151, which is apparently breaking a lot of builds.

PR #713, opened 2016-10-18
[merged] rhci.sh: don't install requirements.txt

jlebon:
It's actually unnecessary here since we're already installing all the
packages needed just before.

And anyway, we should be using the packaged equivalents to ensure
compatibility with what we'll actually be running against.


cgwalters:
It looks like you added a `fixup!` for the wrong commit?

jlebon:
Ahh, drats. Fixed up the fix up!

cgwalters:
@rh-atomic-bot r+ 0952ccf

PR #715, opened 2016-10-18
[merged] Atomic/images.py: Enable filter for dangling.

baude:
The dangling filter was being omitted from the filter list
when running atomic images list.


rhatdan:
@rh-atomic-bot r+

PR #717, opened 2016-10-21
[merged] Atomic: Prune unused methods

baude:
Removing unused methods to reduce our code counts which will
come in handy when refactoring occurs.


rhatdan:
Looks like these functions were not used.  LGTM

rhatdan:
@rh-atomic-bot r+

PR #718, opened 2016-10-21
[merged] Atomic/atomic.py: Explore RepoTags for Fq name

baude:
In cases where the input name is not fully-qualified and
the local image is tagged more than once, we iterate repotags
in an attempt to figure out the fq name.  If this fails, we
raise a ValueError like before.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -16,7 +16,7 @@ from . import util
 import re
 from .util import NoDockerDaemon, DockerObjectNotFound
 from docker.errors import NotFound
-from .discovery import RegistryInspect
+from .discovery import RegistryInspect, RegistryInspectError
 
 def find_repo_tag(d, Id, image_name):
     def image_in_repotags(image_name, repotags):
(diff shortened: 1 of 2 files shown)


baude:
looks like this will require a rebase, will do should it pass tests and the previous commit completes merge.

Atomic/atomic.py
```diff
@@ -379,8 +379,18 @@ def get_fq_name(self, image_info):
             if self.image in image_info['RepoTags']:
                 return self.image
 
+            possibles = []
+            for i in image_info['RepoTags']:
+                try:
+                    possibles.append(self.get_fq_image_name(i))
+                except RegistryInspectError:
+                    possibles.append(None)
```
rhatdan: What does this do?  Sholdn't we just continue?

baude: It appends None to the list, so later when I check if they all resolve to the same fq name, it will disqualify it.  I could theoretically raise the ValueError at this point to be quicker, but as of now, the final raise handles both cases.

rhatdan:
I think we should have a test for this.

Atomic/atomic.py
rhatdan: Wouldn't this be a strange case where an Image would throw an exception on get_fq_image_name?  Would something be wrong in their database or a bug in code?  I think we should throw a different exception, so that it would be easier to debug.

rhatdan:
@baude we need to get this merged today, if possible.  Want to cut a final atomic release for rhel7.3.2

rhatdan:
@rh-atomic-bot r+

PR #720, opened 2016-10-21
[merged] Add --storage option to image-related commands

yuqi-zhang:
Add an optional --storage flag to the following commands:
 - images delete
 - info
 - mount
 - verify
 - version

If specified, the command will only look at the specified storage
(ostree/docker) for the image to perform the action. If not
specified, the command will look through both ostree/docker for
the image (as it was before). However, if the storage is not
specified and the image exists in both ostree and docker, the
command will error and prompt the user to specify. Image inspection
also no longer forces the user to delete/rename one or the other.

This is meant to address the duplicate naming issue (where a user
can have an image in both ostree and docker with the same name).


rhatdan:
@baude and @giuseppe are working on a redesign to handle this.  Please work together.

giuseppe:
@rhatdan I think the refactoring must depend on this PR, which is solving a real problem for atomic users.  I am going to review this PR more in detail, but let's get this merged before anything else, I and @baude will adapt the refactoring on top of this.

rhatdan:
Ok I am cutting a  final release for 7.3.1 on Monday.  So please finish review if you want this to get released soon.

giuseppe:
@rh-atomic-bot r+ 427cd00

PR #721, opened 2016-10-22
[merged] syscontainers: Fix docker: and dockertar: installs

petervo:


rhatdan:
@giuseppe PTAL

rhatdan:
@petervo do you have a test where this is failing?

rhatdan:
@petervo It looks like pull_dockter_tar does not return anything?

giuseppe:
@rhatdan the patch LGTM, if you have no objections to it, I will merge it.

petervo:
Seems emails aren't getting through.

@rhatdan, ```pull_dockter_tar```, returns imagename on line 1002 from inside the try: finally: block. 

As far as testing, without this change, running any variation of ```atomic install --system docker:xxx``` or ```atomic install --system dockertar:xxx``` failed for me because the image name that gets assigned to the ostree ref does not match the image name that gets passed to ```_do_checkout```.

I added some tests to ```test_system_containers.sh```. Hopefully that's the right place.

rhatdan:
Sorry I missed this.

@rh-atomic-bot r+

petervo:
On 10/22/2016 03:22 AM, Daniel J Walsh wrote:
> @petervo <https://github.com/petervo> It looks like pull_dockter_tar
> does not return anything?

It returns imagename on line 1002 from inside the try: finally: block.

PR #724, opened 2016-10-24
[merged] .redhat-ci: test on multiple platforms

jlebon:
Make use of the new support for multiple testsuites in Red Hat CI. As a
start, we test on Fedora 24 Atomic Host, Fedora 24 Cloud, and CentOS
Continuous.


jlebon:
The CentOS test seems to fail because of formatting differences between the output of gpg2 on Fedora & CentOS:

```
$ gpg2 --with-colons key1.pub # on F24
pub:-:4096:1:199E2F91FD431D51:1256212795:::-:
uid:::::::::Red Hat, Inc. (release key 2) <security@redhat.com>:
$ gpg2 --with-colons key1.pub # on CentOS
pub:-:4096:1:199E2F91FD431D51:1256212795:::-:Red Hat, Inc. (release key 2) <security@redhat.com>:
```

Seems like an easy fix, though I'd rather let someone more well-versed in the trust code do the fix.

.redhat-ci.sh
```diff
@@ -0,0 +1,34 @@
...
+# somewhat mimic the spec conditional
+source /etc/os-release
+if [ "$ID" == fedora ]; then
+  PYTHON=python3
+else
+  PYTHON=python
+fi
+
+DOCKER_RUN="docker run \
+              --privileged \
```
cgwalters: You also want `--rm` here I assume so we're not leaking containers.

.redhat-ci.Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+MAINTAINER Jonathan Lebon <jlebon@redhat.com>
+
+# NB: we also install python2 reqs here (which the builddep
+# does not catch since on F24 we build for py3) so that we
+# can do pylint for both py2 and py3 as well as reuse the
+# same image for RHEL & CentOS tests rather than maintaining
+# two separate images.
+
+RUN dnf install -y \
+        make \
```
cgwalters: AKA `@buildsys-build`, as mock uses for Fedora in e.g. `config_opts['chroot_setup_cmd'] = 'install @buildsys-build'`

.redhat-ci.sh
jlebon: This runs on dynamically provisioned hosts, so not too worried here. Though for good form we should.

.redhat-ci.Dockerfile
jlebon: Except `@buildsys-build` pulls in **a lot** more than what we need here. I could make the change, though I don't see the big win when the only overlapping package we actually need is `make`.

jlebon:
:arrow_up:  Added `--rm` above!

> Two minor comments. Looks interesting, so the key thing seems to be `---` for adding multiple documents?

Yup exactly! :) I added some details to the sample YAML file to document it: https://github.com/jlebon/redhat-ci/blob/master/sample.redhat-ci.yml.

jlebon:
@aweiteka Could you investigate the CentOS trust issues above?

jlebon:
There's also a `test_system_containers` failure on CentOS:

```
+ atomic --debug verify busybox
/usr/bin/atomic --debug verify busybox: too few arguments
Try '/usr/bin/atomic --debug verify busybox --help' for more information.
  File "/usr/bin/atomic", line 186, in <module>
    sys.exit(_func())
  File "/usr/lib/python2.7/site-packages/Atomic/verify.py", line 69, in verify
    return self.verify_system_image()
  File "/usr/lib/python2.7/site-packages/Atomic/verify.py", line 191, in verify_system_image
    self.validate_system_image_manifests(layers)
  File "/usr/lib/python2.7/site-packages/Atomic/verify.py", line 229, in validate_system_image_manifests
    mismatches = self.syscontainers.validate_layer(layer)
  File "/usr/lib/python2.7/site-packages/Atomic/syscontainers.py", line 1194, in validate_layer
    traverse(it)
  File "/usr/lib/python2.7/site-packages/Atomic/syscontainers.py", line 1174, in traverse
    dir_checksum = it.get_dir().out_content_checksum
AttributeError: 'tuple' object has no attribute 'out_content_checksum'
```

Seems like it might be a Python 2/3 introspection issue -- @giuseppe could you take a look?

giuseppe:
@jlebon could you please cherry-pick this patch https://github.com/giuseppe/atomic/commit/7108584cb9ca1846ae910089c47bcbceb50b9522 ?

jlebon:
OK, this is ready to merge!

rhatdan:
@rh-atomic-bot r+

jlebon:
Let's keep the previous tester around for a week or two and then we can talk about decommissioning it.

PR #728, opened 2016-10-25
[merged] Allow granular UNIT tests

baude:
Making several changes to allow for more granular testing with make
check.  First, pylint tests are now run from tests.sh.  They do not
get run if a granular test is desired.

You can still run a granular INTEGRATION test with the TEST_INTEGRATION
keyword followed by the name (excluding test and .sh).  For example,

make check TEST_INTEGRATION=top

This will only run the top integration test.  We then do NOT run any of
the UNIT tests as it is assumed that the developer only wants to test
the specific integration test.

You can now test a singular UNIT test with the TEST_UNIT keyword.  Like
the integration test, simply name the test and exclude the prefixed
"test_" and file suffix of ".py"  For example,

make check TEST_UNIT=discovery

This will only run the discovery unit test; the pylint test and integration
tests will be skipped.


baude:
@aweiteka @jlebon @rhatdan PTAL

test.sh
```diff
@@ -127,8 +129,21 @@ make_docker_images () {
...
+
+if [ ! -n "${TEST_INTEGRATION+ }" ] && [ ! -n "${TEST_UNIT+ }" ]; then
+    $PYTHON -m pylint --disable=all --enable=E --enable=W --additional-builtins=_ *.py atomic Atomic tests/unit/*.py -d=no-absolute-import,print-statement,no-absolute-import,bad-builtin
+    $PYTHON3 -m pylint --disable=all --enable=E --enable=W --additional-builtins=_ *.py atomic Atomic tests/unit/*.py -d=no-absolute-import,print-statement,no-absolute-import,bad-builtin
+
+else
+    echo "      SKIPPING PYLINT TESTS..."
+fi
+
 make_docker_images
```
aweiteka: Can we move this below unittests? My env requires sudo to run this function and it takes 8 seconds to attempt the build.

rhatdan:
Hopefully the test_trust.py is fixed now.  Changed self.savesigstore to self.save.

rhatdan:
I think the pylint tests should continue to be done on make install.

baude:
@rhatdan reverted some of the behaviour for pylint for rhci.sh.  pylint is still run for make install

test.sh
```diff
@@ -127,6 +129,10 @@ make_docker_images () {
     done
 }
 
+# export args to environment so the conditional checks below
+# work as expected.
+export $@
+
```
jlebon: Is this really necessary? When called from `make`, this is already done for you. If calling `test.sh` directly, then you can set it yourself beforehand: `TEST_UNIT=mytest ./test.sh`.

test.sh
```diff
@@ -11,6 +11,8 @@ export DOCKER=${DOCKER:-"/usr/bin/docker"}
 export SECRET=`dd if=/dev/urandom bs=4096 count=1 2> /dev/null | sha256sum`
 export ATOMIC_LIBEXEC="$(pwd)"
 export ATOMIC_CLIENT="$(pwd)/atomic_client.py"
+export PYTHON=/usr/bin/python
+export PYTHON3=/usr/bin/python3
```
baude: Oversight on my part, leftover from a previous commit.  Removing

rhatdan:
I want to wait on this til after atomic-1.13 is released.

rhatdan:
@rh-atomic-bot r+

Issue #730, opened 2016-10-25
should continue to execl rpm-ostree for `host`


cgwalters:
In https://github.com/projectatomic/atomic/pull/723/commits/125e33044a3d887272024fff6e70c787e80219fb
we changed:

```
-        os.execl("/usr/bin/" + cmd, *(args + aargs))
+        return util.check_output([ "/usr/bin/" + cmd ] + aargs)
```

This means the user doesn't see _any_ output until the command completes, when executed interactively.  This is an annoying regression.

cgwalters:
Why are we exposing the Host command via DBus?  The chain here is rather silly, since of course rpm-ostree is just going to talk dbus to its daemon...

Basically unless there's a user of this, let's nuke it.

cgwalters:
There's a bigger picture architectural issue here in that at some point, I'd like `host status` to say something meaningful about containers/docker/etc.  The appropriate place to integrate those is probably in the atomic command, not in rpm-ostree.   So if we did something like that we'd probably want a DBus API for it etc.

But let's not introduce complexity unless we need it.

rhatdan:
The user would be cockpit, we are attempting to make all atomic commands available via cockpit, but as you point out interactive commands won't work.

rhatdan:
I can change the code to indicate whether or not you are are running it from the console or via dbus.

cgwalters:
Cockpit already talks to rpm-ostree directly, what would be the value in changing that?

rhatdan:
Not sure.  I was just going through all of the Verbs of atomic to add dbus support.  If there is no value add to atomic host, then we can remove the patch.

rhatdan:
Removed host from dbus interfaces.

PR #731, opened 2016-10-26
[merged] Just exec rpmostree if in tty mode

rhatdan:
If running from dbus continue to return output.


rhatdan:
@cgwalters Does this fix https://github.com/projectatomic/atomic/issues/730

cgwalters:
@rh-atomic-bot r+ 476609d

PR #733, opened 2016-10-26
[merged] Revert atomic host dbus support

rhatdan:
OSTree commands already have Cockpit/dbus support.  Users should
call them directly.


rhatdan:
@cgwalters PTAL

cgwalters:
Thanks, looks sane to me.

@rh-atomic-bot r+ a2d1b5f

PR #734, opened 2016-10-27
[merged] Add substitutions for Opt variables

rhatdan:
I think this will fix https://github.com/projectatomic/atomic/issues/540

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -304,13 +304,13 @@ class Atomic(object):
         newenv['IMAGE'] = self.image or ""
 
         if hasattr(self.args, 'opt1') and self.args.opt1:
-            newenv['OPT1'] = self.args.opt1
+            newenv['OPT1'] = os.path.expandvars(self.args.opt1)
 
         if hasattr(self.args, 'opt2') and self.args.opt2:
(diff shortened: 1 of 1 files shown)


rhatdan:
@adelton PTAL

rhatdan:
@adelton PTAL

Atomic/atomic.py
```diff
@@ -304,13 +304,13 @@ def cmd_env(self):
         newenv['IMAGE'] = self.image or ""
 
         if hasattr(self.args, 'opt1') and self.args.opt1:
-            newenv['OPT1'] = self.args.opt1
+            newenv['OPT1'] = os.path.expandvars(self.args.opt1)
```
rhatdan: The documentation defines these as UPPER CASE
```
man atomic-run
...
       OPT1, OPT2, OPT3
         Additional options which can be specified via the command.
```

baude: OK LGTM

rhatdan:
@rh-atomic-bot r+

adelton:
> @adelton PTAL

Sorry for late reply. It looks good to me.

rhatdan:
@rh-atomic-bot r+

PR #735, opened 2016-10-27
[merged] test.sh: also look for coverage3

jlebon:
I meant for the `fedora/24/cloud` suite to run its tests under `coverage`, but that hasn't been happening because it uses the Python 3 coverage package, which is installed at `/usr/bin/coverage3`, which we don't look for.


rhatdan:
@rh-atomic-bot r+

PR #736, opened 2016-10-27
make vagrant-check: Run tests with vagrant

baude:
Run tests in a vagrant session (w/libvirt for now).  This allows
the tests to occur similar to the way our jenkins tests work. Run
make vagrant-check and we created an F24 session and install
the proper deps.  We then call the .redhat-ci.sh script inside
the vagrant session and the tests are executed.  At this time, you
will need to vagrant halt to shutdown the session when comlete.


Vagrantfile
```diff
@@ -0,0 +1,27 @@
...
+  config.vm.synced_folder ".", "/vagrant", type: "rsync",
+    rsync__exclude: ".tmp*"
+
+  # The most common configuration options are documented and commented below.
+  # For a complete reference, please see the online documentation at
+  # https://docs.vagrantup.com.
+
+  # Every Vagrant development environment requires a box. You can search for
+  # boxes at https://atlas.hashicorp.com/search.
+  config.vm.box = "fedora/24-cloud-base"
```
jlebon: Why not `fedora/24-atomic-host` here? Or even better, define all three (the CentOS box is called `centos/atomic-host`). Someone can then choose to only bring up a single one, e.g. `vagrant up fedora/24/atomic`.

baude: Well, two quick reasons.  

1) If I understood you correctly, the CentOS and Fedora Atomic host required greater preparations than simply dnf install coverage and atomic.  And that is normally handled specially by your tools right which arent in play local use.

2. Aim small, miss small :)

In all seriousness, if you think I should pursue, I'm happy to do so.  I think the Makefile would then need to call a bash script instead of trying to handle this all in make.  Thoughts?

rhatdan:
SGTM

Vagrantfile
baude: Agree.  I'll fixer it up.

cgwalters:
See also https://github.com/projectatomic/rpm-ostree/pull/502

Vagrantfile
cgwalters: We actually have Vagrant boxes generated for CAHC (both continuous and alpha) btw, and I switched to using CAHC Alpha in https://github.com/projectatomic/rpm-ostree/pull/502

baude:
Rebasing

miabbott:
One thing of note, if you run `make vagrant-check` in succession, you are going to get an error like this:

```
$ make vagrant-check
vagrant up "fedora_atomic"
Bringing machine 'fedora_atomic' up with 'libvirt' provider...
vagrant ssh "fedora_atomic" -c "cd /home/vagrant/atomic && sudo sh ./.redhat-ci.sh"
+ mount --make-rshared /
+ systemctl start docker
+ '[' -f /run/ostree-booted ']'
+ ostree admin unlock
error: Deployment is already in unlocked state: development
Connection to 192.168.121.23 closed.
Makefile:97: recipe for target 'vagrant-check' failed
make: *** [vagrant-check] Error 1
```

I think you'll have to `vagrant destroy` before re-running it.

vagrant.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+
+is_running() {
+    status=$(vagrant status | grep fedora_atomic | awk '{print $2}')
```
miabbott: You'll have to change this to grep for `${BOX}` since it is selectable

vagrant.sh
```diff
@@ -0,0 +1,31 @@
...
+    if [ ${status} == "running" ]; then
+        echo ""
+        echo "'${BOX}' is already running.  Perform a 'vagrant halt|destroy ${BOX}'"
+        echo ""
+        exit 1
+    fi
+}
+
```
miabbott: Maybe add some logic before running the bulk of the script to validate that the value of `${BOX}` is supported?  Some usage feedback with supported values would go nicely here, too.

vagrant.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
```
miabbott: `set -euo pipefail` would be a nice addition

vagrant.sh
```diff
@@ -0,0 +1,31 @@
...
+        exit 1
+    fi
+}
+
+echo "Testing on ${BOX}"
+timestamp=$(date +%Y_%m_%d_%H_%M)
+tee_file="${BOX}_${timestamp}.log"
+is_running
+vagrant up ${BOX}
+vagrant ssh ${BOX} -c "cd /home/vagrant/atomic && sudo sh ./.redhat-ci.sh"
```
miabbott: I think this needs to be piped to `tee`?

vagrant.sh
baude: Got it, good catch @miabbott

vagrant.sh
baude: The piping to tee seems to break $? reflecting the actual test result.

baude:
@rhatdan this is good to go now, not sure what is up with the ci tests

vagrant.sh
```diff
@@ -0,0 +1,52 @@
...
+# out in _FINISH
+trap _FINISH EXIT
+
+BOXES="fedora_atomic centos_atomic"
+
+is_running() {
+    status=$(vagrant status | grep ${BOX} | awk '{print $2}')
+    if [ ${status} == "running" ]; then
+        echo ""
+        echo "'${BOX}' is already running.  Perform a 'vagrant halt|destroy ${BOX}'"
```
jlebon: Rather than having to destroy and bring back up each time, can't we just `vagrant rsync` here if we see it's already up?

jlebon: We can leave that to the user, no? I.e. expect them to know when the previous `make vagrant-check` is done before starting a new one. Or are you thinking of another use case here?

baude: @jlebon OK I have implemented your suggestion but I hit ...

```*** 'fedora_atomic' is already running.  Re-syncing and rerunning test ...

/usr/share/gems/gems/net-ssh-2.9.1/lib/net/ssh/transport/session.rb:67:in `initialize': Object#timeout is deprecated, use Timeout.timeout instead.
/usr/share/gems/gems/net-ssh-2.9.1/lib/net/ssh/transport/session.rb:84:in `initialize': Object#timeout is deprecated, use Timeout.timeout instead.
==> fedora_atomic: Rsyncing folder: /home/bbaude/docker/atomic/ => /home/vagrant/atomic
==> fedora_atomic:   - Exclude: [".vagrant/", ".tmp*"]
+ mount --make-rshared /
+ systemctl start docker
+ '[' -f /run/ostree-booted ']'
+ ostree admin unlock
error: Deployment is already in unlocked state: development
Connection to 192.168.121.217 closed.

** Test failed.  Leaving 'fedora_atomic' running for debug.
** Be sure to halt or destroy prior to re-running the check
** Logs are stored at fedora_atomic_2016_11_03_09_34.log

Makefile:97: recipe for target 'vagrant-check' failed
make: *** [vagrant-check] Error 1
```

So can your rh ci script handle this?

rhatdan:
@miabbott @cgwalters Are you guys happy with this?

cgwalters:
Down the line we'll definitely want to do more dedup across projects and such, but this seems sane to me to start.

cgwalters:
Why does the PR have a merge from master instead of being rebased though?

jlebon:
Yup, looks good on my end as well.

rhatdan:
@baude REBASE please and then we can merge.

baude:
@rhatdan rebased

rhatdan:
@rh-atomic-bot r+

baude:
@rh-atomic-bot retry

PR #737, opened 2016-10-27
[merged] Atomic/sign.py: Fix manifest digest for write location

baude:
The manifest has was incorrect in atomic, and therefore the
wrote location for signatures is incorrect.  The problem was
two-fold in that the manifest_has that should be written should
not be json_encoded (ever).  Also, in some cases, the manifest
file was null.  This has been fixed.

### DIFF
--- a/Atomic/sign.py
+++ b/Atomic/sign.py
@@ -3,7 +3,6 @@ from . import Atomic
 from . import discovery
 import os
 import tempfile
-import json
 
 try:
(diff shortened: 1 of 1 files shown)


rhatdan:
@aweiteka PTAL

Is this something we need now in atomic-1.13.4 for RHEL7.3.1?

rhatdan:
@rh-atomic-bot r+

PR #738, opened 2016-10-28
[merged] Atomic/sign.py: Remove tag from signing path

baude:
It looks like skopeo removed the tag from the signing path
after 7.3.1.  Removing tag in atomic as well.


rhatdan:
@ddarrah PTAL

ddarrah:
@rhatdan @baude tested, LGTM

rhatdan:
@rh-atomic-bot r+

PR #739, opened 2016-10-28
[merged] .redhat-ci.yml: switch to CAHC image stream

jlebon:
Now that the backend has support for frequent, up-to-date CAHC images,
we can make use of it here and avoid a rebase and a reboot.


cgwalters:
@rh-atomic-bot r+ 5b4511f

PR #741, opened 2016-10-31
[merged] Atomic/sign.py: Ensure image tag is included in signing.

baude:
The previous change where we removed the tag from the write
file location impacted the signing because the tag was also
being omitted. This ensures the tag is omitted from the file
path but included in the call skopeo standalone-sign.


ddarrah:
@baude LGTM

rhatdan:
@rh-atomic-bot r+

PR #743, opened 2016-11-01
[merged] Atomic/mount.py: Handle self.options == None

baude:
When using the dbus interface to perform a scan, self.options
was being set to None.  This tripped a condition with the
setting of args where it could not be iterated.  We now ensure
self.options is not None prior to the iteration.


rhatdan:
@rh-atomic-bot r+

PR #744, opened 2016-11-03
[merged] Atomic/verify.py: Ensure probes use fq name

baude:
When walking the layers of an image in atomic verify, we need to be sure
we use a fq image name when probing.  In cases where the tagging includes
a non-fq image name, we were not.  This resolves Bugzilla 1377952.


baude:
@rh-atomic-bot r+

PR #745, opened 2016-11-03
[merged] Atomic storage reset does not work on docker-latest

rhatdan:
This patch will allow user to specify the graphdriver on atomic reset

If /var/lib/docker or /var/lib/docker-latest is the only thing installed
it will reset the correct path.  If both exists or the user as chosen
a different location, the --graph option must be specified.

### DIFF
--- a/Atomic/storage.py
+++ b/Atomic/storage.py
@@ -121,14 +121,14 @@ class Storage(Atomic):
             self.graphdir = self.args.graph
         else:
             if os.path.exists("/var/lib/docker") and os.path.exists("/var/lib/docker-latest"):
-                raise ValueError("You must select the graph storage path to reset: /var/lib/docker or /var/lib/docker-latest")
+                raise ValueError("You must specify the --graph storage path to reset /var/lib/docker or /var/lib/docker-latest")
             if os.path.exists("/var/lib/docker"):
                 self.graphdir = "/var/lib/docker"
             else:
                 if os.path.exists("/var/lib/docker-latest"):
                     self.graphdir = "/var/lib/docker-latest"
                 else:
-                    raise ValueError("You must select the graph storage path to reset")
+                    raise ValueError("Could not find any default graph storage path. Specify one using --graph option")
 
     def reset(self):
         root=self.graphdir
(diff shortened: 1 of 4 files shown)


rhatdan:
@rhvgoyal PTAL

rhatdan:
@rhvgoyal PTAL

rhatdan:
@baude PTAL

Atomic/storage.py
```diff
@@ -111,8 +114,24 @@ class Storage(Atomic):
...
+        else:
+            if os.path.exists("/var/lib/docker") and os.path.exists("/var/lib/docker-latest"):
+                raise ValueError("You must select the graph storage path to reset: /var/lib/docker or /var/lib/docker-latest")
+            if os.path.exists("/var/lib/docker"):
+                self.graphdir = "/var/lib/docker"
+            else:
+                if os.path.exists("/var/lib/docker-latest"):
+                    self.graphdir = "/var/lib/docker-latest"
+                else:
+                    raise ValueError("You must select the graph storage path to reset")
```
rhvgoyal: We probably could be little more help and say something like, `Could not find any default graph storage path. Specify one using --graph option`

docs/atomic-storage.1.md
```diff
@@ -90,5 +90,15 @@ Drivers supported: devicemapper, overlay, overlay2
...
+# reset OPTIONS
+**-h** **--help**
+  Print usage statement
+
+**--graph**
+Root of the docker runtime. If you are running docker at the default
+location (/var/lib/docker), you don't need to pass this flag. However
+if you are running docker at a custom location. This flag must be set.
+
+
```
rhvgoyal: We probably need to specify default behavior if --graph is not specified. `By default /var/lib/docker or /var/lib/docker-latest/ will be reset. If both are present or none are present user needs to specify the path using --graph option.`

rhatdan:
@rhvgoyal Hows it look now.

baude:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #747, opened 2016-11-04
[merged] Atomic/discovery.py: Python3 byte vs str fixes

baude:
In code paths with Python 3, when dealing with remote registries,
we find that under certain conditions values are set as bytes which
need to be encoded/decoded to work properly with things like json
and ord().

### DIFF
--- a/Atomic/discovery.py
+++ b/Atomic/discovery.py
@@ -87,17 +87,17 @@ class RegistryConnection():
         if 'signatures' in self.manifest_json:
             return self._get_digest_from_signature()
         else:
-            return hashlib.sha256(self.orig_manifest).hexdigest()
+            return hashlib.sha256(self.orig_manifest.encode('utf-8')).hexdigest()
 
     def _get_digest_from_signature(self):
         for i in self.manifest_json['signatures']:
             _protected = i['protected']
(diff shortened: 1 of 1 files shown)


rhatdan:
Tests are failing with the following
```
ERROR: test_inspect (test_discovery.TestAtomicUtil)
  File "/root/checkout/tests/unit/test_discovery.py", line 31, in test_inspect
    inspect_info = ri.inspect()
  File "/root/checkout/Atomic/discovery.py", line 235, in inspect
    return _get_inspect_info()
  File "/root/checkout/Atomic/discovery.py", line 223, in _get_inspect_info
    return self.rc.inspect_schema2(self.assemble_fqdn(include_tag=False), self.tag)
  File "/root/checkout/Atomic/discovery.py", line 76, in inspect_schema2
    'Digest': "sha256:{}".format(self.get_digest()),
  File "/root/checkout/Atomic/discovery.py", line 90, in get_digest
    return hashlib.sha256(self.orig_manifest).hexdigest()
TypeError: Unicode-objects must be encoded before hashing
```

baude:
@rhatdan looks good now I believe

baude:
@rhatdan looks good now

rhatdan:
@rh-atomic-bot r+

PR #748, opened 2016-11-05
[merged] Return image if exact match

rhatdan:


rhatdan:
In current code I have three container images named docker
```
atomic mount rhatdan/docker:latest /mnt
"rhatdan/docker:latest" matched multiple items. Try one of the following:
	rhatdan/docker:latest
	dwalsh/docker:latest
	docker.io/docker:latest
```
But even though I use the exact name, it fails.

This patch will fix this issue.

rhatdan:
@baude PTAL

baude:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #751, opened 2016-11-08
[merged] syscontainers: fix installation

giuseppe:
bug introduced by:

syscontainers: use image instead of self.args.image


baude:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

cgwalters:
Okay, something got confused here on the homu side, I think because the PR is trying push to the `refactor` branch.

I think we should close this PR and make a new one that is intended for master.

PR #752, opened 2016-11-08
[merged] syscontainers: environment variable detection

yuqi-zhang:
Add functionality to detect environment variables in template files.
The variables that can be set by the user through --set is displayed
on 'atomic images info'. The variables with overridable default
values will either have their default value displayed, or shown as
{SET_BY_OS} if atomic sets them. Variables without default value
must be set by the user and is listed separately.

Also add check to install of system containers to error out if a
value is required but not set.


rhatdan:
@giuseppe PTAL

Atomic/syscontainers.py
```diff
@@ -306,6 +309,23 @@ def _do_checkout(self, repo, name, img, upgrade, values, destination, unitfileou
         if rev is None:
             raise ValueError("Image %s not found" % img)
 
+        if not upgrade and not extract_only:
```
giuseppe: we can use `--set` on `update --container` as well.  Can we drop the `not upgrade` condition here?

yuqi-zhang: As pointed out the check is redundant, and I have removed it.

giuseppe:
@rh-atomic-bot r+ 981d644

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot r+

giuseppe:
@rh-atomic-bot r+

rhatdan:
@jlebon Is something wrong with the BOT?

jlebon:
@rh-atomic-bot retry

jlebon:
It's been feeling a little under the weather recently.

Issue #753, opened 2016-11-09
atomic help only looks for "HELP" label


twaugh:
There is no "HELP" label listed here:
https://github.com/projectatomic/ContainerApplicationGenericLabels

However there is a "help" label. I suggest that `help.py` should check for `help` first, then fall back to the compatibility name `HELP`, when looking for a command to display image help.

twaugh:
Changing the issue name.

This does not only affect HELP, but also INSTALL, UNINSTALL, STOP, and probably every label. None of these are defined as upper-case, and all of them are defined as lower-case. The atomic CLI *must* use the names specified in the label documentation.

rhatdan:
https://github.com/projectatomic/atomic/pull/754 Should fix this issue.

INSTALL, UNINSTALL, and STOP already try all cases.

twaugh:
Oh yes, I hadn't seen `_get_args()` in the other cases.

baude:
Is this ok to close?

PR #754, opened 2016-11-09
[merged] The HELP label by default should be "help"

rhatdan:
But we want to check for all cases.  Currently if a image uses the
standard help label, atomic will fail to find it.


rhatdan:
@baude PTAL

rhatdan:
Fixes https://github.com/projectatomic/atomic/issues/753

rhatdan:
@baude PTAL

baude:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #757, opened 2016-11-09
[merged] Refactor

baude:
Initial work to move from using conditionals for detecting the proper backend to using an abstract class instead.  This allows us to add new backends with less difficulty.

We also now implement the use of "objects" for images, layers, and containers which eliminates the need to deal with multiple data structures in Atomic and other methods.


Atomic/backends/__init__.py
```diff
@@ -0,0 +1 @@
+# Backends
```
rhatdan: Add RETURN.

Atomic/backends/_docker.py
```diff
@@ -0,0 +1,271 @@
...
+        if not deep:
+            con_obj.status = con_struct['Status']
+            con_obj.image = con_struct['ImageID']
+
+        if deep:
+            # Add in the deep inspection stuff
+            con_obj.status = con_struct['State']['Status']
+            con_obj.running = con_struct['State']['Running']
+            con_obj.image = con_struct['Image']
+
```
rhatdan: Make this block 
```
if deep:
   ...
else: 
  ...
```

Atomic/backends/_docker.py
```diff
@@ -0,0 +1,271 @@
...
+    def get_images(self, get_all=False):
+        images = self.get_docker_images(get_all=get_all)
+        image_objects = []
+        for image in images:
+            image_objects.append(self._make_image(image['Id'], image))
+        return image_objects
+
+    def make_remote_image(self, image):
+        img_obj = self._make_remote_image(image)
+        img_obj.remote_inspect()
```
rhatdan: Does this return something or just update an img_obj?

Atomic/backends/_docker.py
```diff
@@ -0,0 +1,271 @@
...
+        return con_objects
+
+    def get_docker_images(self, get_all=False, quiet=False, filters=None):
+        if filters:
+            assert isinstance(filters, dict)
+        else:
+            filters = {}
+        return self.d.images(all=get_all, quiet=quiet, filters=filters)
+
+    def get_docker_containers(self):
```
rhatdan: Why have these functions inside of a "Docker" backend?

Atomic/backends/_docker.py
```diff
@@ -0,0 +1,271 @@
...
+    def has_image(self, img):
+        err_append = "Refine your search to narrow results."
+        self.input = img
+        image_info = self.get_docker_images(get_all=True)
+
+        img_obj = self.inspect_image(image=img)
+        if img_obj:
+            return img_obj
+        name_search = util.image_by_name(img, images=image_info)
+        if len(name_search) > 0:
```
baude: Fixed

Atomic/backends/_docker.py
baude: Fixed

Atomic/backends/_docker.py
baude: It returns a new "remote" image object and would be used like:

remote_busybox = backend.make_remote_image('docker.io/library/busybox:latest')

Atomic/backends/_docker.py
baude: I have changed get_docker_images|get_docker_containers to _get_images|_get_containers respectively so they are now private.  The differ from the public methods in that the public methods return a list of objects.

Atomic/backends/_docker.py
rhatdan: My point is remote_inspect is a horrible name for this function.  I would figure this would  return a json file describing the image.  Why does it exist as a public member, Why is it not built into _make_remote_image?

Atomic/backends/__init__.py
baude: Fixed

Atomic/backends/_docker.py
baude: FWIW, remote image objects will be handy for verify, version, and info verbs.

rhatdan:
@baude could you merge some of these together. And then it LGTM

baude:
@rh-atomic-bot r+

PR #758, opened 2016-11-10
[merged] syscontainers: output better json errors

yuqi-zhang:


Atomic/syscontainers.py
```diff
@@ -184,7 +184,10 @@ def install(self, image, name):
 
     def _check_oci_configuration_file(self, conf_path, remote=None):
         with open(conf_path, 'r') as conf:
-            configuration = json.loads(conf.read())
+            try:
+                configuration = json.loads(conf.read())
+            except ValueError:
+                raise ValueError("Invalid json in configuration file.")
```
rhatdan: Output the conf_path also.

Atomic/syscontainers.py
```diff
@@ -400,7 +403,10 @@ def _do_checkout(self, repo, name, img, upgrade, values, destination, unitfileou
         manifest_file = os.path.join(exports, "manifest.json")
         if os.path.exists(manifest_file):
             with open(manifest_file, "r") as f:
-                manifest = json.loads(f.read())
+                try:
+                    manifest = json.loads(f.read())
+                except ValueError:
+                    raise ValueError("Invalid manifest.json file in image.")
```
rhatdan: Output the manifest_file

Atomic/syscontainers.py
```diff
@@ -446,7 +452,10 @@ def _write_template(inputfilename, data, values, destination):
 
         if remote_path:
             with open(destination_path, 'r') as config_file:
-                config = json.loads(config_file.read())
+                try:
+                    config = json.loads(config_file.read())
+                except ValueError:
+                    raise ValueError("Invalid config.json file in given remote path.")
```
rhatdan: output the destination path.

Atomic/syscontainers.py
```diff
@@ -671,7 +680,10 @@ def get_template_variables(self, image):
         if manifest_template:
             fd = manifest_template.get_fd()
             with os.fdopen(fd) as f:
-                data = json.loads(f.read())
+                try:
+                    data = json.loads(f.read())
+                except ValueError:
+                    raise ValueError("Invalid manifest.json file in image.")
```
rhatdan: If you can get back to the fd path, output it.

Atomic/syscontainers.py
yuqi-zhang: Same as above, this would be in the rootfs/exports (also not created since the install fails)

Atomic/syscontainers.py
yuqi-zhang: Same as above, except this is saved in an ostree layer, which doesn't really help the user much.

Atomic/syscontainers.py
rhatdan: I understand your point, but If I was using the atomic tool, and got this error message, I would have no idea what it means.  Which configuration file would be my question, and I could potentially waste time searching and looking at any configuration file related to atomic command.  /etc/atomic.conf, /etc/atomic.d/* /etc/containers ...

Atomic/syscontainers.py
rhatdan: Maybe 
Invalid manifest.json file in IMAGENAME image.

Atomic/syscontainers.py
rhatdan: Yes I think so.

Atomic/syscontainers.py
rhatdan: Well name the image.

rhatdan:
@rh-atomic-bot r+

PR #762, opened 2016-11-17
[merged] syscontainers: add rollback

yuqi-zhang:
Add --rollback flag to atomic update, which switches a system
container to the other deployment if one exists.


yuqi-zhang:
As a side note perhaps it makes sense to refractor the commands to `atomic containers update` and `atomic containers rollback` instead of being `atomic update --container` and `atomic update --rollback --container`.

rhatdan:
Yes that makes more sense to me.  I love the idea of getting rid of base verbs also.

giuseppe:
can we do the refactoring separately?  It doesn't depend from this PR, which I think can be already merged

yuqi-zhang:
Sure, I am working on refactoring already based on this branch. Feel free to merge this, or I can add commits to this PR if it's better.

giuseppe:
fine to merge this for now?  We can work on the refactoring separately, as anyway it influences the entire "update" command

Atomic/syscontainers.py
```diff
@@ -607,14 +601,12 @@ def rollback(self, name):
         previous_deployment = 0
         if os.path.realpath(path).endswith(".0"):
             previous_deployment = 1
-        else:
-            previous_deployment = 0
 
-        if not os.path.exists("%s.%d" % (path, previous_deployment)):
-            raise ValueError("Error: Cannot find a previous deployment to rollback.")
+        destination = "%s.%d" % (path, previous_deployment)
```
rhatdan: How about remove prvious_deployment altogether.
 destination = "%s.%d" % (path, (1 if os.path.realpath(path).endswith(".0") else 0)

rhatdan:
LGTM except for last suggestion.

yuqi-zhang:
Fixed and rebased

rhatdan:
@rh-atomic-bot r+

PR #767, opened 2016-11-23
Generate RPM and support arbitrary files copied to the host

giuseppe:
this is a prototype to add "pre installed" system containers.  The idea is to have another deployment tree under `/usr/lib/containers/atomic` that is not handled by "atomic install/uninstall/update", where '/var/lib/containers/atomic' is fully handled by atomic.  The systemd files are also copied under '/usr/lib'.

The second patch adds the possibility to install a system container into a rpm.  The `---generate-rpm` (three dashes) option is hidden for now.   The spec file is auto generated, and stuff like `Requires`, `Provides`, `Conflicts` can be specified through labels.  So in the Dockerfile something like: `LABEL Provides="etcd"`, will be translated directly to the specfile.

Example:
```
# atomic --debug install --system ---generate-rpm gscrivano/etcd  
# rpm -i x86_64/atomic-container-etcd-1.0-1.0.x86_64.rpm
```

### DIFF
--- a/Atomic/install.py
+++ b/Atomic/install.py
@@ -62,6 +62,9 @@ def cli(subparser):
         system_xor_user.add_argument("--system", dest="system",
                                      action='store_true', default=False,
                                      help=_('install a system container'))
+        installp.add_argument("---generate-rpm", dest="generate_rpm",
+                              action='store_true', default=False,
+                              help=_('generate an rpm instead of installing the container'))
         installp.add_argument("--rootfs", dest="remote",
                               help=_("choose an existing exploded container/image to use "
                                      "its rootfs as a remote, read-only rootfs for the "
--- a/Atomic/install.py
+++ b/Atomic/install.py
@@ -62,9 +62,8 @@ def cli(subparser):
         system_xor_user.add_argument("--system", dest="system",
                                      action='store_true', default=False,
                                      help=_('install a system container'))
-        installp.add_argument("---generate-rpm", dest="generate_rpm",
-                              action='store_true', default=False,
-                              help=_('generate an rpm instead of installing the container'))
+        installp.add_argument("--system-package", dest="system_package", default="auto",
+                              help=_('control how to install the package.  It accepts `auto`, `yes`, `no`, `build`'))
         installp.add_argument("--rootfs", dest="remote",
                               help=_("choose an existing exploded container/image to use "
                                      "its rootfs as a remote, read-only rootfs for the "
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -293,7 +293,7 @@ class SystemContainers(object):
             raise ValueError("The container's rootfs is set to remote, but the remote rootfs does not exist")
         return real_path
 
-    def _checkout(self, repo, name, img, deployment, upgrade, values=None, destination=None, extract_only=False, remote=None, prefix=None):
+    def _checkout(self, repo, name, img, deployment, upgrade, values=None, destination=None, extract_only=False, remote=None, prefix=None, installed_files=None):
         destination = destination or "%s/%s.%d" % (self._get_system_checkout_path(), name, deployment)
         unitfileout, tmpfilesout = self._get_systemd_destination_files(name, prefix)
 
@@ -303,7 +303,7 @@ class SystemContainers(object):
                     raise ValueError("The file %s already exists." % f)
 
         try:
-            return self._do_checkout(repo, name, img, upgrade, values, destination, unitfileout, tmpfilesout, extract_only, remote, prefix)
+            return self._do_checkout(repo, name, img, upgrade, values, destination, unitfileout, tmpfilesout, extract_only, remote, prefix, installed_files=installed_files)
         except (ValueError, OSError) as e:
             try:
                 if not extract_only and not upgrade:
@@ -365,7 +365,23 @@ class SystemContainers(object):
 
         return [get_image(i) for i in matches]
 
-    def _do_checkout(self, repo, name, img, upgrade, values, destination, unitfileout, tmpfilesout, extract_only, remote, prefix=None):
+    @staticmethod
+    def _write_template(inputfilename, data, values, destination):
+        try:
+            os.makedirs(os.path.dirname(destination))
+        except OSError:
+            pass
+        with open(destination, "w") as outfile:
+            template = Template(data)
+            try:
+                result = template.substitute(values)
+            except KeyError as e:
+                os.unlink(destination)
+                raise ValueError("The template file '%s' still contains an unreplaced value for: '%s'" % \
+                                 (inputfilename, str(e)))
+            outfile.write(result)
+
+    def _do_checkout(self, repo, name, img, upgrade, values, destination, unitfileout, tmpfilesout, extract_only, remote, prefix=None, installed_files=None):
         if not values:
             values = {}
 
@@ -488,28 +504,13 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         values["HOST_UID"] = os.getuid()
         values["HOST_GID"] = os.getgid()
 
-        def _write_template(inputfilename, data, values, destination):
-            try:
-                os.makedirs(os.path.dirname(destination))
-            except OSError:
-                pass
-            with open(destination, "w") as outfile:
-                template = Template(data)
-                try:
-                    result = template.substitute(values)
-                except KeyError as e:
-                    os.unlink(destination)
-                    raise ValueError("The template file '%s' still contains an unreplaced value for: '%s'" % \
-                                     (inputfilename, str(e)))
-                outfile.write(result)
-
         src = os.path.join(exports, "config.json")
         destination_path = os.path.join(destination, "config.json")
         if os.path.exists(src):
             shutil.copyfile(src, destination_path)
         elif os.path.exists(src + ".template"):
             with open(src + ".template", 'r') as infile:
-                _write_template(src + ".template", infile.read(), values, destination_path)
+                SystemContainers._write_template(src + ".template", infile.read(), values, destination_path)
         else:
             self._generate_default_oci_configuration(destination)
 
@@ -537,6 +538,8 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
                 except subprocess.CalledProcessError:
                     pass
 
+        new_installed_files = self._rm_add_files_to_host(installed_files, exports, prefix)
+
         missing_bind_paths = self._check_oci_configuration_file(destination_path, remote_path, True)
 
         image_manifest = self._image_manifest(repo, rev)
@@ -551,8 +554,10 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
                     "ostree-commit": rev,
                     'created' : calendar.timegm(time.gmtime()),
                     "values" : values,
+                    "installed-files": new_installed_files,
                     "remote" : remote}
             info_file.write(json.dumps(info, indent=4))
+            info_file.write("\n")
 
         if os.path.exists(unitfile):
             with open(unitfile, 'r') as infile:
@@ -566,10 +571,10 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         else:
             tmpfiles_template = SystemContainers._generate_tmpfiles_data(missing_bind_paths, values["STATE_DIRECTORY"])
 
-        _write_template(unitfile, systemd_template, values, unitfileout)
+        self._write_template(unitfile, systemd_template, values, unitfileout)
         shutil.copyfile(unitfileout, os.path.join(prefix, destination, "%s.service" % name))
         if (tmpfiles_template):
-            _write_template(unitfile, tmpfiles_template, values, tmpfilesout)
+            self._write_template(unitfile, tmpfiles_template, values, tmpfilesout)
             shutil.copyfile(tmpfilesout, os.path.join(prefix, destination, "tmpfiles-%s.conf" % name))
 
         if prefix:
@@ -634,6 +639,35 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             return [image_inspect]
         return None
 
+    @staticmethod
+    def _rm_add_files_to_host(old_installed_files, exports, prefix="/"):
+        # if any file was installed on the host delete it
+        if old_installed_files:
+            for i in old_installed_files:
+                try:
+                    os.remove(i)
+                except OSError:
+                    pass
+
+        if not exports:
+            return []
+
+        # if there is a directory hostfs/ under exports, copy these files to the host file system.
+        hostfs = os.path.join(exports, "hostfs")
+        new_installed_files = []
+        if os.path.exists(hostfs):
+            for root, _, files in os.walk(hostfs):
+                rel_root_path = os.path.relpath(root, hostfs)
+                if not os.path.exists(os.path.join(prefix, rel_root_path)):
+                    os.makedirs(os.path.join(prefix, rel_root_path))
+                for f in files:
+                    path = os.path.join(prefix, rel_root_path, f)
+                    shutil.copy2(os.path.join(root, f), path)
+                    new_installed_files.append(os.path.join("/", rel_root_path, f))
+            new_installed_files.sort()  # just for an aesthetic reason in the info file output
+
+        return new_installed_files
+
     def update_container(self, name, setvalues=None, rebase=None):
         if self._is_preinstalled_container(name):
             raise ValueError("Cannot update a preinstalled container")
@@ -659,6 +693,7 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         image = rebase or info["image"]
         values = info["values"]
         revision = info["revision"] if "revision" in info else None
+        installed_files = info["installed-files"] if "installed-files" in info else None
 
         # Check if the image id or the configuration for the container has
         # changed before upgrading it.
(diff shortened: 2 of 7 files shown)


Atomic/syscontainers.py
```diff
@@ -1355,3 +1361,77 @@ def get_out_checksum(obj): return obj.out_checksum if hasattr(obj, 'out_checksum
...
+
+        spec = spec + "\n%description\n"
+        if description:
+            spec = spec + "%s\n" % description
+
+        spec = spec + "\n%files\n"
+        for root, _, files in os.walk(os.path.join(destdir, "etc")):
+            rel_path = os.path.relpath(root, destdir)
+            for f in files:
+                spec += "%config \"%s\"\n" % os.path.join("/", rel_path, f)
```
jfilak: I'm confused, this line looks like you want to install container's /etc to host's /etc.

giuseppe: here I am mapping from the temporary directory (where the container was checked out) to the absolute path `/etc' in the host (where the .rpm will install these files)

jfilak: Oh sorry, I thought that the destdir variable points to the container's root.

Atomic/install.py
```diff
@@ -59,6 +59,9 @@ def cli(subparser):
                 system_xor_user.add_argument("--system", dest="system",
                                              action='store_true', default=False,
                                              help=_('install a system container'))
+                installp.add_argument("---generate-rpm", dest="generate_rpm",
```
rhatdan: I would prefer that the user not need to specify this.  This should happen automatically if the image supports generation of rpms for install.

rhatdan: Users will forget to do this, and we want to eventually make this the default for all of our system containers.

Atomic/syscontainers.py
```diff
@@ -419,7 +432,11 @@ def _do_checkout(self, repo, name, img, upgrade, values, destination, unitfileou
                 key, val = i[:split], i[split+1:]
                 values[key] = val
 
-        values["DESTDIR"] = destination
+        if prefix:
+            values["DESTDIR"] = os.path.join("/", os.path.relpath(destination, prefix))
```
rhatdan: values["DESTDIR"] = (os.path.join("/", os.path.relpath(destination), prefix)) if prefix else destination)

Atomic/syscontainers.py
```diff
@@ -501,24 +518,28 @@ def _write_template(inputfilename, data, values, destination):
...
         if (tmpfiles_template):
             _write_template(unitfile, tmpfiles_template, values, tmpfilesout)
-            shutil.copyfile(tmpfilesout, os.path.join(destination, "tmpfiles-%s.conf" % name))
+            shutil.copyfile(tmpfilesout, os.path.join(prefix, destination, "tmpfiles-%s.conf" % name))
 
-        sym = "%s/%s" % (self._get_system_checkout_path(), name)
-        if os.path.exists(sym):
-            os.unlink(sym)
-        os.symlink(destination, sym)
+        if not prefix:
```
rhatdan: Should we just
```
if prefix:
     return
```

Atomic/syscontainers.py
```diff
@@ -1317,3 +1361,77 @@ def get_out_checksum(obj): return obj.out_checksum if hasattr(obj, 'out_checksum
...
+            cwd = os.getcwd()
+            cmd = ["rpmbuild", "--noclean", "-bb", spec_file,
+                   "--define", "_sourcedir %s" % temp_dir,
+                   "--define", "_specdir %s" % temp_dir,
+                   "--define", "_builddir %s" % temp_dir,
+                   "--define", "_srcrpmdir %s" % cwd,
+                   "--define", "_rpmdir %s" % cwd,
+                   "--build-in-place",
+                   "--buildroot=%s" % rpm_content]
+            util.write_out(" ".join(cmd))
```
rhatdan: if self.debug:

rhatdan:
I think there should be a rpm specification template in the container image that triggers the rpm build for install.  This should be simpler then the rpm.spec,   Perhaps just the files section as well as description, license, summary and potentially some other fields.

Should we default post install of rpm package to 

atomic uninstall CONTAINER

Atomic/install.py
giuseppe: the --generate-rpm doesn't probably belong here, I just  didn't know where to add it.

I think we will need a way to just generate the .rpm, so that it can be used as any regular .rpm, you can even install it without using atomic at all.

Installing the .rpm could be the default here, but we will need a way to specify also to install the container in the usual way using `/var/lib` instead of '/usr/lib' if needed.

giuseppe:
We can add a spec template, as we do with the other configuration files (and offer a substitution for the list of the installed files, it would be very tedious to list them otherwise).  I thought of using labels so we will be able to query resolves/provides/conflicts remotely without fetching the image first.

I have not added a preun script yet.  I think it should just be to stop the systemd service.  I think "atomic uninstall" should not interfere with rpm to handle/delete the files.

rhatdan:
Right I guess we could get a circular dependency here.  
I guess atomic uninstall should execute

rpm -e PACKAGE.

giuseppe:
@rhatdan I have addressed your comments for the code.

I have pushed a bunch of other patches to cleanly support copied files to the host.

I have included the idea of @jfilak to copy files from `/exports/hostfs` to the host.  These files are tracked by the rpm when generating one, otherwise they are managed by atomic and removed on "atomic uninstall".

The files copied to the host can be also template files, the same substitution rules that are used for the runc configuration file and the systemd service file apply here.

It is possible to specify that a system container doesn't install any service, so that can be used just as a way to copy files to the host, or to generate an rpm without the overhead of the service+rootfs files.

rhatdan:
@TomasTomecek PTAL

rhatdan:
@baude PTAL

Atomic/install.py
rhatdan: Right I would say that is the fallback.  
I believe the user puts the data in as templates rather then in labels.  I am thinking a rpm.yaml file which specifies information for the spec file, but allows atomic CLI to do substitutions like the name of the container so the spec file ends up with foobar_etcd, and we could prefix all executables or paths with the container names.

jfilak:
I like this pull request and I told @giuseppe that he has done a great job in private chat. However, there is one thing that concerns me a bit. I think it's a pity that the whole system container thing is bundled in 'atomic' utility. I develop my containers on Fedora and I would appreciate a tool capable of installing a test container on my development machine.

rhatdan:
atomic command is available on Fedora as well as atomic host.  It is a separate package.

cgwalters:
Atomic Host is part of Fedora.  It *is* Fedora as much as a system managed via yum.

rhatdan:
Right  atomic CLI is available on Fedora Server, Fedora Workstation, Fedora Cloud/Atomic

jfilak:
So it is possible to develop a DNF plugin installing system containers through 'atomic' utility. It would be cool to run:
```
dnf container-install gscrivano/etcd
```

Atomic/syscontainers.py
```diff
@@ -1317,3 +1493,86 @@ def get_out_checksum(obj): return obj.out_checksum if hasattr(obj, 'out_checksum
...
+        for root, _, files in os.walk(os.path.join(destdir, "usr/lib/systemd/system")):
+            for f in files:
+                spec = spec + "/usr/lib/systemd/system/%s\n" % f
+        for root, _, files in os.walk(os.path.join(destdir, "usr/lib/tmpfiles.d")):
+            for f in files:
+                spec = spec + "/usr/lib/tmpfiles.d/%s\n" % f
+        if installed_files:
+            for i in installed_files:
+                spec = spec + "%s\n" % i
+
```
TomasTomecek: Would it make sense to add changelog? So users know that the rpm is generated and installed by `atomic`.

TomasTomecek:
Got this traceback with image you provided to me:

```
$ sudo ./atomic --debug install --system ---generate-rpm --display img
Extracting to /tmp/tmprTqPaB/rpmroot/usr/lib/containers/atomic/img
./atomic --debug install --system ---generate-rpm --display img: too few arguments
Try './atomic --debug install --system ---generate-rpm --display img --help' for more information.
  File "./atomic", line 187, in <module>
    sys.exit(_func())
  File "/home/tt/g/atomic/Atomic/install.py", line 90, in install
    return self.syscontainers.install(self.image, self.name)
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 169, in install
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 1507, in generate_rpm
    self._checkout(repo, name, image, 0, False, destination=rootfs, prefix=rpm_content)
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 276, in _checkout
    return self._do_checkout(repo, name, img, upgrade, values, destination, unitfileout, tmpfilesout, extract_only, remote, prefix, installed_files
=installed_files)
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 363, in _do_checkout
    was_service_active = self._is_service_active(name)
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 965, in _is_service_active
    return self._systemctl_command("is-active", name, quiet=True).replace("\n", "") == "active"
AttributeError: 'NoneType' object has no attribute 'replace'
```

giuseppe:
@TomasTomecek  `--display` is not really supported when generating an rpm as we need to read information from the files that are installed/checked out from the OSTree repository.  I have added a fixup patch though to not cause the traceback when `--display` is used.

rhatdan:
I would like to setup a BlueJeans tomorrow to help design this tool to make sure everyone is on the same page.  We need this to work for use cases outside of --system containers as well.

TomasTomecek:
> We need this to work for use cases outside of --system containers as well.

This would be awesome. Right now I'm in a process of writing down use cases which could benefit greatly from this feature. Please, invite me to the discussion.

giuseppe:
yes, this PR makes possible to install arbitrary files on the host and track them with a rpm without requiring a running or even installed container/systemd service.

Every file under /exports/hostfs will be copied to the host file system.  If `noContainerService` is used in the `manifest.json` file then atomic won't try to install the systemd service and the runc container.

rhatdan:
@giuseppe I scheduled the meeting tomorrow at 8:30 AM EST.  I would like you to take the lead to describe what you are building and display different parts.  Could you throw together a Google Doc in the mean time to help facilitate the discussion.

rhatdan:
BTW Add anyone you would like to the meeting.

rhatdan:
@TomasTomecek If you could write up your Use Cases, that would be helpful also.

giuseppe:
I've added a new patch that allows the user to directly specify a spec file, with these rules:

- If `/exports/rpm.spec` is present, then it is used as it is, otherwise,
- If `/exports/rpm.spec.template` is present, it is first preprocessed as any template file and then it is used, otherwise
- We generate a spec file

rhatdan:
@giuseppe @baude we need to start moving forward on this.

giuseppe:
@rhatdan we can probably skip the "automatically install the rpm if .spec is defined" part for now.  We can add this later once we are able to install local rpm on Atomic Host.  We will still need to be able to build rpm offline anyway.  Not sure where the option should be placed though. "atomic install --system --only-generate-rpm" ?

rhatdan:
I am on PTO this week. I guess you and @baude need to figure this out.  We can discuss the options next Monday, but that should not prevent you guys from hacking on it.

giuseppe:
@rhatdan let's just decide how to present the offline "generate rpm" functionality, which probably we want to have anyway.  We can postpone the automatic installation of an .rpm later, once we can do it on Atomic Host (and also have clearer ideas on when/how to use it).

rhatdan:
I would prefer to have this be automatic and not involve the user at all.  The atomic command and image should have enough data that atomic CLI can figure out if it needs to build an RPM.

rhatdan:
@giuseppe should we have a meeting to discuss this feature?  Or do you have enough to start?

giuseppe:
@rhatdan we can setup a meeting so we can discuss the last details.

@rhatdan I think we will still need to be able to build an rpm offline though, so that it can be included in the Atomic Host image if needed.  @cgwalters do you think we should allow this?

TomSweeneyRedHat:
Hi Giuseppe,

If possible, could you include me on the meeting invite too please?  I'd
like to sit in for informational purposes.

Thanks!

t

TomasTomecek:
I would also appreciate invite. @giuseppe will you be in Westford next week?

giuseppe:
@TomasTomecek no, I won't be in Westford.  Sure I'll send an invitation to you for the meeting

rhatdan:
I added @baude @cgwalters @sctweedie and Stef Walters to invite also

giuseppe:
While there is probably more to do (e.g. show some information in `atomic show`) and more tests, I think it is fine to give it a try.  I dropped the [RFC][WIP] tags to make it clearer.

rhatdan:
@baude PTAL

TomasTomecek:
I played with this today and wasn't able to make it work, unfortunately.

This is the exception I got:

```
$ sudo ./atomic --debug install --system --system-package=build docker:modularitycontainers/nginx
Namespace(_class=<class 'Atomic.install.Install'>, args=[], assumeyes=False, debug=True, display=False, func='install', image='docker:modularitycontainers/nginx', name=None,
opt1=None, opt2=None, opt3=None, remote=None, setvalues=None, storage=None, system=True, system_package='build', user=False)
g-io-error-quark: Not a regular file or symlink: loop1 (0)
  File "./atomic", line 189, in <module>
    sys.exit(_func())
  File "/home/tt/g/atomic/Atomic/install.py", line 106, in install
    return self.syscontainers.install(self.image, self.name)
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 215, in install
    image = self._pull_image_to_ostree(repo, image, False)
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 160, in _pull_image_to_ostree
    image = self._pull_docker_image(repo, image.replace("docker:", "", 1))
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 1401, in _pull_docker_image
    return self._pull_docker_tar(repo, temptar.name, image)
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 1422, in _pull_docker_tar
    self._pull_dockertar_layers(repo, imagebranch, temp_dir, input_layers, labels=labels)
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 1595, in _pull_dockertar_layers
    self._import_layers_into_ostree(repo, imagebranch, manifest, layers_to_import)
  File "/home/tt/g/atomic/Atomic/syscontainers.py", line 1367, in _import_layers_into_ostree
    repo.write_directory_to_mtree(Gio.File.new_for_path(temp_dir), mtree, modifier)
Error: g-io-error-quark: Not a regular file or symlink: loop1 (0)
```

[This is](https://github.com/container-images/nginx/blob/64c2f087c19f6fb20ab33e8a1ed96e5a5a4e17ab/Dockerfile) my Dockerfile.

What's really interesting is that when I checked files in the temp directory (`/var/lib/containers/atomic/.Bze55J`), the `/dev` was fully populated and that's what I'm assuming caused the error above:

```
/var/lib/containers/atomic/.Bze55J $ ll dev
total 0
crw------- 1 root mock  5,   1 Mar  8 19:10 console
lrwxrwxrwx 1 root mock      13 Mar 20 17:28 fd -> /proc/self/fd
crw-rw-rw- 1 root mock  1,   7 Mar  8 19:10 full
brw-rw-rw- 1 root mock  7,   0 Mar  8 19:10 loop0
brw-rw-rw- 1 root mock  7,   1 Mar  8 19:10 loop1
brw-rw-rw- 1 root mock  7,   2 Mar  8 19:10 loop2
brw-rw-rw- 1 root mock  7,   3 Mar  8 19:10 loop3
brw-rw-rw- 1 root mock  7,   4 Mar  8 19:10 loop4
crw-rw-rw- 1 root mock 10, 237 Mar  8 19:10 loop-control
crw-rw-rw- 1 root mock  1,   3 Mar  8 19:10 null
lrwxrwxrwx 1 root mock       8 Mar 20 17:28 ptmx -> pts/ptmx
drwxrwsr-x 2 root mock       6 Mar  8 19:10 pts
crw-rw-rw- 1 root mock  1,   8 Mar  8 19:10 random
drwxrwsr-x 2 root mock       6 Mar  8 19:10 shm
lrwxrwxrwx 1 root mock      15 Mar 20 17:28 stderr -> /proc/self/fd/2
lrwxrwxrwx 1 root mock      15 Mar 20 17:28 stdin -> /proc/self/fd/0
lrwxrwxrwx 1 root mock      15 Mar 20 17:28 stdout -> /proc/self/fd/1
crw-rw-rw- 1 root tty   5,   0 Mar  8 19:10 tty
cr--r--r-- 1 root mock  1,   9 Mar  8 19:10 urandom
crw-rw-rw- 1 root mock  1,   5 Mar  8 19:10 zero
```

Also, we would really appreciate if we could use this functionality, copying files to host & creating an RPM out of them, with docker containers.

giuseppe:
sorry I probably gave you the wrong info.  To pull an image to the OSTree storage you need to do:

`./atomic pull --storage ostree docker:modularitycontainers/nginx`

and once it is stored in OSTree then you can install it:

`./atomic install --system modularitycontainers/nginx`

We could probably make atomic install smarter and do the previous steps at once.

I am not sure how this could be done with Docker containers, as differently than system containers, it is not enough to copy some files to get the container seen by Docker.  If you refer to copy some files to the host part, we could probably do that, just you'll have to bind the .rpm to the docker container lifecycle somehow (i.e. you can uninstall the docker container but the .rpm is still installed).  Do you have any example of what kind of files do you need to copy from a Docker container?

rhatdan:
Yes 
atomic install --system docker:modularitycontainers/nginx  
Should pull the image and store it in OSTree.

giuseppe:
@rhatdan I have done few more changes.  A lot of changes are accumulated here and if there are no objections, I think it is fine to merge, some parts need to be cleaned up and also tests need to be added but we can do it on master.
I know that @ashcrow is working on the same portion on the code, getting this merged will probably save some merge conflicts.

rhatdan:
Ok I would prefer to wait to merge until tomorrow, so we can get RHEL release out, without this patch.

giuseppe:
@rhatdan yes sure, let's give this some time before it gets into a release

rhatdan:
@rh-atomic-bot r+ 4321643

PR #769, opened 2016-11-23
[merged] Atomic/mount.py: Re-Add _clean_temp_container_by_path (BZ 1397839)

baude:
Commit 9a0114c2845beba372116b2fdcb605a3f59b445b removed several
methods not used by Atomic CLI.  Unfortunately, one of the methods
was being called by openscap.  This is reported in bz# 1397839.

Reverted the removal of the method.


baude:
merging as it doesnt impact atomic cli code, only oscap
@rh-atomic-bot r+

PR #770, opened 2016-11-24
[merged] atomic_dbus: keep the name until the process exits

### LINKED ISSUE #763: Each D-Bus call on org.atomic adds an atomic_dbus.py process
None of them owns the name `org.atomic`, which is why every call activates the program again.

larskarlitski:
The call dbus.service.BusName() claims the name while the returned
object is alive. It wasn't stored into a variable, so the name was
released immediately, but the service keeps running (potentially
indefinitely) with no use.

Fix this by calling request_name() directly, so that we can give useful
feedback (and end the process) when the name is already owned by another
process.

Fixes #763

### DIFF
--- a/atomic_dbus.py
+++ b/atomic_dbus.py
@@ -29,6 +29,8 @@ from Atomic.trust import Trust
 from Atomic.uninstall import Uninstall
 from Atomic.verify import Verify
 
+DBUS_NAME_FLAG_DO_NOT_QUEUE = 4
+DBUS_REQUEST_NAME_REPLY_PRIMARY_OWNER = 1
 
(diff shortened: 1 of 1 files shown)


rhatdan:
bot, add author to whitelist

rhatdan:
@cgwalters PTAL

cgwalters:
I'd be surprised if there wasn't a nicer binding for this somewhere in slip or python-dbus, but eh:

@rh-atomic-bot r+ b2a125f

cgwalters:
Looks like the fedora/24/atomic failure is the test suite not being safe for concurrent runs on the same system.

PR #771, opened 2016-11-28
[merged] Refactor images

baude:
Covers all but verify and generate.  This is a refactoring of the
images subverbs (i.e. info, version, delete, ...)

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -625,7 +625,7 @@ class Atomic(object):
         as a JSON string so it can then be parsed appropriately.
         """
         try:
-            return open(os.path.join(self.results, "scan_summary.json"), "r").read()
+            return json.loads(open(os.path.join(self.results, "scan_summary.json"), "r").read())
         except IOError:
             return "{}"
 
--- a/Atomic/delete.py
+++ b/Atomic/delete.py
@@ -1,8 +1,7 @@
 from . import Atomic
 from . import util
-from docker.errors import NotFound
-from docker.errors import APIError
 import sys
+from Atomic.backendutils import BackendUtils
 
 class Delete(Atomic):
     def __init__(self):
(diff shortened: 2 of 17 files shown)


baude:
@giuseppe can you review this as well? Also, there seems to be a system_containers test failure on my system; if it shows here, can you help me understand what is going wrong?

giuseppe:
I think the test is failing because we are now calling the backend "ostree", while we used to refer to  the type of image as "system" (or "user" for user mode).

Could you try to `s|type=system|type=ostree|` in the bash/atomic and tests/integration/test_system_containers.sh files?

Atomic/delete.py
```diff
@@ -67,44 +94,44 @@ def _delete_remote(self, targets):
...
-                except APIError as e:
-                    util.write_err("Failed operation for delete Image {}: {}".format(target, e))
-                    results = 2
-
-            else:
-                util.write_err("{} is not a valid storage".format(self.args.storage))
-                results = 2
-
-        return results
+#    def _delete_local(self, targets, force=False):
```
giuseppe: if this code is not needed anymore, let's just drop it

Atomic/delete.py
```diff
@@ -13,34 +12,62 @@ def delete_image(self):
...
             if not confirm in ['y', 'yes']:
                 util.write_err("User aborted delete operation for {}".format(self.args.delete_targets))
                 sys.exit(2)
 
-        if self.args.remote:
-            results = self._delete_remote(self.args.delete_targets)
-        else:
-            results = self._delete_local(self.args.delete_targets, self.args.force)
-        return results
+        # Peform the delete
```
giuseppe: s|Peform|Perform|

Atomic/info.py
```diff
@@ -104,123 +98,32 @@ def info(self):
...
+
+        with closing(StringIO()) as buf:
+            try:
+                info_name = img_obj.fq_name
+            except RegistryInspectError:
+                info_name = img_obj.input_name
+            buf.write("Image Name: {}\n".format(info_name))
+            buf.writelines(sorted(["{}: {}\n".format(k, v) for k,v in list(img_obj.labels.items())]))
+            if img_obj.template_variables_set:
+                buf.write("\n\nEnvironment variables with default value, but overridable with --set:\n")
```
giuseppe: let's use "Template variables", as they are not passed as env variables for system containers.

Atomic/info.py
```diff
@@ -104,123 +98,32 @@ def info(self):
...
+                buf.writelines(["{}: {}\n".format(k, v) for k,v in list(img_obj.template_variables_set.items())])
+            if img_obj.template_variables_unset:
+                buf.write("\n\nEnvironment variables that has no default value, and must be set with --set:\n")
```
giuseppe: same here

Atomic/objects/image.py
```diff
@@ -149,4 +164,108 @@ def long_version(self):
...
+
+def convert_size(size):
+    if size > 0:
+        size_name = ("B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB")
+        i = int(math.floor(math.log(size, 1000)))
+        p = math.pow(1000, i)
+        s = round(size/p, 2) # pylint: disable=round-builtin,old-division
+        if s > 0:
+            return '%s %s' % (s, size_name[i])
+    return '0B'
```
giuseppe: missing '\n' here

Atomic/delete.py
rhatdan: I agree drop it

Atomic/images.py
```diff
@@ -109,6 +110,7 @@ def cli(subparser):
 class Images(Atomic):
     def __init__(self):
         super(Images, self).__init__()
+        self.beu = backendutils.BackendUtils()
```
rhatdan: Not crazy about self.beu.  Means nothing to me.

Atomic/delete.py
baude: Fixed

Atomic/delete.py
baude: Done, was holding it to be certain it passed tests first.

Atomic/info.py
baude: Done

Atomic/info.py
baude: Done

Atomic/objects/image.py
baude: Added newline, thanks

Atomic/images.py
baude: Done

baude:
@rhatdan addressed feedback, etc

Atomic/atomic.py
```diff
@@ -178,8 +178,10 @@ def set_args(self, args):
                 self.name = self.name + "-spc"
             if self.system or self.user:
                 self.name = self.syscontainers.get_default_system_name(self.image)
-
-        self.syscontainers.set_args(self.args)
+        try:
+            self.syscontainers.set_args(self.args)
+        except (NameError, AttributeError):
```
rhatdan: Is this dangerous, since some args might not have been set?

rhatdan:
Other then question LGTM.  If it is good with @giuseppe Merge it.

baude:
I'll revert that bit and see if it passes tests.  @giuseppe  reviewed it this morning.  If it passes, Ill merge.

baude:
@rh-atomic-bot r+

PR #772, opened 2016-11-29
[merged] Add fedora25_cloud target for vagrant

baude:
This allows us to test on Fedora 25 cloud images via make vagrant-check.

### DIFF
--- a/.redhat-ci.sh
+++ b/.redhat-ci.sh
@@ -4,13 +4,16 @@ set -xeuo pipefail
 # https://bugzilla.redhat.com/show_bug.cgi?id=1318547#c7
 mount --make-rshared /
 
-systemctl start docker
 
 if [ -f /run/ostree-booted ]; then
     if [ ! -e /var/tmp/ostree-unlock-ovl.* ]; then
         ostree admin unlock
     fi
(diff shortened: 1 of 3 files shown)


.redhat-ci.sh
```diff
@@ -4,14 +4,17 @@ set -xeuo pipefail
...
 mount --make-rshared /
 
-systemctl start docker
 
 if [ -f /run/ostree-booted ]; then
     if [ ! -e /var/tmp/ostree-unlock-ovl.* ]; then
         ostree admin unlock
     fi
+else
+    dnf install -y atomic python3-coverage
```
cgwalters: What's simpler to debug?

baude: It is simpler to debug on a cloud-image for certain test failure (least for me).  It really doesnt matter.  This is only for local testing for developers so we can close this if there is objection.

cgwalters: But what commands are different?  What's simpler about it?

cgwalters:
This wasn't really answered, but we need to support both paths I guess, so OK.

@rh-atomic-bot r+ bf1b2a1

PR #774, opened 2016-11-29
[merged] Atomic Info Unittests

baude:
Add unittests for ostree, docker backends for the version
sub-verb to images.

### DIFF
--- a/Atomic/info.py
+++ b/Atomic/info.py
@@ -67,19 +67,24 @@ class Info(Atomic):
         self.beu = BackendUtils()
 
     def version(self):
+        self._version(util.write_out)
+
+    def _version(self, write_func):
         layer_objects = self.get_layer_objects()
         max_version_len = max([len(x.long_version) for x in layer_objects])
         max_version_len = max_version_len if max_version_len > 9 else 9
         max_img_len = len(max([y for x in layer_objects for y in x.repotags], key=len)) + 9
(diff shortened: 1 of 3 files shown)


cgwalters:
Looks like pylint is being annoying about having a `TODO` string.

Atomic/info.py
```diff
@@ -66,20 +66,29 @@ def __init__(self):
         super(Info, self).__init__()
         self.beu = BackendUtils()
 
+    def version_tty(self):
+        if self.args.debug:
+            util.write_out(str(self.args))
+        util.write_out(self.version())
```
cgwalters: Oh, I see, so we can capture it more easily in the unit test.   I think I probably would have just tested the JSON in the unit tests, and done screen-scraping (i.e. exec `atomic` as a subprocess) for the tty output.

But this is OK too.

cgwalters: A cleaner way to do this would be to have an internal `_versions()` function that takes the stream as an argument.  Then then unit tests could pass `StringIO()`, the  tty version would use `sys.stdout`.

cgwalters: Oh yeah, using stringio is better than `+=`.  My comment above builds on that - we'd just use `sys.stdout` for ttys.

baude:
Comments fixed @rhatdan

rhatdan:
@rh-atomic-bot r+

PR #776, opened 2016-11-30
[merged] Various D-Bus related fixes and enhancements

jlebon:
These patches make it easier to use atomic through the D-Bus interface.

### DIFF
--- a/Atomic/pull.py
+++ b/Atomic/pull.py
@@ -3,7 +3,7 @@ try:
 except ImportError:
     from atomic import Atomic  # pylint: disable=relative-import
 from .trust import Trust
-from .util import skopeo_copy, get_atomic_config, Decompose, write_out, strip_port, is_insecure_registry
+from .util import skopeo_copy, get_atomic_config, Decompose, write_out, strip_port, is_insecure_registry, check_call
 
 ATOMIC_CONFIG = get_atomic_config()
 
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
@baude PTAL

Atomic/pull.py
```diff
@@ -38,19 +38,23 @@ def pull_docker_image(self):
...
-        trust = Trust()
-        trust.set_args(self.args)
-        trust.discover_sigstore(fq_name)
-        write_out("Pulling {} ...".format(fq_name))
-        skopeo_copy("docker://{}".format(fq_name), image,
-                    debug=self.args.debug, insecure=insecure,
-                    policy_filename=self.policy_filename)
+        if self.args.image.startswith("dockertar:"):
+            path = self.args.image.replace("dockertar:", "", 1)
+            check_call(["docker", "load", "-i", path])
```
baude: would you consider using the load_image API from docker-py? i.e self.d.load_image()

baude:
A couple of simple comments.  Else LGTM

cgwalters:
Call me Captain Obvious :ship: but a some point we need to cut :scissors: over and have the command be a client for the API, otherwise the DBus API is going to bitrot.

rhatdan:
I agree but that also brings in complications, in that dbus will bring its own complications.  
I guess 
stage 1 is @baude refactoring. 
Stage 2 is start making atomic CLI use DBUS API for common tasks.  Some will probably not be implementable where TTY needs to be shared like atomic run.

Atomic/pull.py
jlebon: Thanks for the suggestion, updated below ⬇️. I didn't change to using `return 0`, since it doesn't add much here. I see this more as the docker backend equivalent of the if-else cascade in `syscontainers.py`. E.g. we might support more prefixes in the future.

jlebon:
Ping, I think this is ready to go.

rhatdan:
@rh-atomic-bot r+

Issue #761, opened 2016-11-11
atomic diff: image filesystem comparison is shallow


twaugh:
When using 'atomic diff' to inspect differences between images, `filecmp.dircmp` is used. This performs a shallow comparison, meaning that files with the same time stamp and size are assumed to be identical.

As a result, these things may be reported as identical:
* symbolic links with different targets
* regular files with different content
* files with different permissions or ownership
* flies with different timestamps but identical content

With the exception of files with different timestamps but identical content, the other cases are very useful to report as they are likely to make the image behave differently.

baude:
After talking wqith @twaugh , we agreed if we are going to do this, we should focus on: symlinks with different targets, regular files with different content,  and files with different perms or ownerships.

rhatdan:
Sounds good, is this something we should look at mtree for or is this something we should put into atomic CLI?

baude:
@rhatdan Yes, I plan to use go-mtree and have been doing some preliminary testing and patching of go-mtree  so that it works correctly.  Been also keeping vbatts up to date as well.

baude:
Closing, re-open if you need additional function

PR #777, opened 2016-12-01
[merged] Atomic/diff.py: Use go-mtree for file comparisons

baude:
The previous algorithm for comparing files used python's
dircmp and is considered to be a shallow comparision.  This
allowed distinctly small possibilities that two files being
compared could be different but not caught.

We now use go-mtree to do the comparison.  This can emulate the
shallow comparison we had before but we can also adding a
sha256digest as part of the comparison using the new --keywords
option.

Also, made slight tweaks to gomtree functions in Atomic.util
so we debug and influence the return of JSON data.

This solves https://github.com/projectatomic/atomic/issues/761

### DIFF
--- a/Atomic/diff.py
+++ b/Atomic/diff.py
@@ -2,13 +2,15 @@ import os
 import sys
 import rpm
 import tempfile
-from filecmp import dircmp
 from . import util
 from . import mount
 from . import Atomic
 from Atomic.client import get_docker_client
 from docker.errors import NotFound
+import json
 
+
+CHOICES=['all', 'link', 'nlink', 'mode', 'type', 'time', 'uid', 'gid', 'size', 'sha256digest']
 
 def cli(subparser):
     # atomic diff
@@ -22,6 +24,10 @@ def cli(subparser):
                        help=_("Compare images' metadata"))
     diffp.add_argument("--json", default=False, action='store_true',
                        help=_("output json"))
+    diffp.add_argument("-k", "--keywords", nargs='?',
+                       action='append',
+                       choices=CHOICES,
+                       help=_("Exclusive keywords to be used for file level comparision"))
     diffp.add_argument("-n", "--no-files", default=False, action='store_true',
                        help=_("Do not perform a file diff between the docker objects"))
     diffp.add_argument("--names-only", default=False,
@@ -49,13 +55,14 @@ class Diff(Atomic):
         '''
         if self.args.debug:
             util.write_out(str(self.args))
+
         helpers = DiffHelpers(self.args)
         images = self.args.compares
         # Check to make sure each input is valid
         for image in images:
             self.get_input_id(image)
 
-        image_list = helpers.create_image_list(images)
+        image_list = helpers.create_image_list(images, self.args)
         try:
             # Set up RPM classes and make sure each docker object
             # is RPM-based
@@ -109,7 +116,7 @@ class DiffHelpers(object):
             image.remove()
 
     @staticmethod
-    def create_image_list(images):
+    def create_image_list(images, args):
         """
         Instantiate each image into a class and then into
         image_list
@@ -119,7 +126,7 @@ class DiffHelpers(object):
         image_list = []
         for image in images:
             try:
-                image_list.append(DiffObj(image))
+                image_list.append(DiffObj(image, args))
             except mount.SelectionMatchError as e:
                 if len(image_list) > 0:
                     DiffHelpers.cleanup(image_list)
@@ -128,20 +135,23 @@ class DiffHelpers(object):
         return image_list
 
     def output_files(self, images, image_list):
-        """
-        Prints out the file differences when applicable
-        :param images:
-        :param image_list:
-        :return: None
-        """
-        file_diff = DiffFS(image_list[0].chroot, image_list[1].chroot)
+        img_left, img_right = (x for x in image_list)
+        foo = img_left.manifest_file_name
+        img_right.manifest_file_name = foo
+        results = img_right.validation_result
+
+        file_diff = DiffFS(image_list)
         for image in image_list:
             self.json_out[image.name] = {'unique_files': file_diff.get_only(image.chroot)}
+
         self.json_out['files_differ'] = file_diff.common_diff
 
         if not self.args.json:
             file_diff.print_results(images[0], images[1])
             util.write_out("\n")
+        # It might make sense to save the mtree results
+        # for the user with a switch
+        return results
 
     def output_rpms(self, rpm_image_list):
         """
@@ -187,7 +197,8 @@ class DiffHelpers(object):
 
 
 class DiffObj(object):
-    def __init__(self, docker_name):
+    def __init__(self, docker_name, args):
+        self.args = args
         self.dm = mount.DockerMount(tempfile.mkdtemp(), mnt_mkdir=True)
         self.name = docker_name
         self.root_path = self.dm.mount(self.name)
@@ -197,6 +208,8 @@ class DiffObj(object):
         else:
             self.chroot = self.root_path
         self.metadata_results = None
+        self._manifest_file_name = None
+        self._validation_results= None
 
     def remove(self):
         """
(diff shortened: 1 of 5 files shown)


Atomic/diff.py
```diff
@@ -22,6 +22,10 @@ def cli(subparser):
                        help=_("Compare images' metadata"))
     diffp.add_argument("--json", default=False, action='store_true',
                        help=_("output json"))
+    diffp.add_argument("-k", "--keywords", nargs='+',
+                       choices=['link', 'nlink', 'mode', 'type', 'time', 'uid', 'gid', 'size', 'sha256digest'],
+                       default=['link', 'nlink', 'mode', 'type', 'time', 'uid', 'gid', 'size'],
+                       help=_("iExclusive keywords to be used for file level comparision"))
```
twaugh: typos?

twaugh:
This works well for me.

Atomic/diff.py
baude: yeah, vim command :)

baude: fixed

Atomic/diff.py
```diff
@@ -22,6 +22,10 @@ def cli(subparser):
                        help=_("Compare images' metadata"))
     diffp.add_argument("--json", default=False, action='store_true',
                        help=_("output json"))
+    diffp.add_argument("-k", "--keywords", nargs='+',
+                       choices=['link', 'nlink', 'mode', 'type', 'time', 'uid', 'gid', 'size', 'sha256digest'],
+                       default=['link', 'nlink', 'mode', 'type', 'time', 'uid', 'gid', 'size'],
+                       help=_("Exclusive keywords to be used for file level comparision"))
```
rhatdan: comparison

rhatdan: Should you add "all"

Atomic/diff.py
```diff
@@ -363,14 +420,29 @@ class DiffFS(object):
...
+    def __init__(self, image_list):
+        #self.compare = dircmp(chroot_left, chroot_right)
+        self.img_left, self.img_right = (x for x in image_list)
         self.left = []
         self.right = []
         self.common_diff = []
-        self.chroot_left = chroot_left
-        self.chroot_right = chroot_right
-        self.delta(self.compare)
+        self.chroot_left = [] #chroot_left
```
rhatdan: Comments are useless, either describe or remove the comment.

Atomic/diff.py
```diff
@@ -363,14 +420,29 @@ class DiffFS(object):
...
+        self.chroot_right =[] # chroot_right
+        self.parse_mtree_json()
+
+    def parse_mtree_json(self):
+        def extra(_result): #pylint: disable=unused-variable
```
rhatdan: These variables are used, aren't they, remove pylint?

Atomic/diff.py
```diff
@@ -363,14 +420,29 @@ class DiffFS(object):
...
+            self.right.append(_result['path'])
+
+        def missing(_result): #pylint: disable=unused-variable
```
rhatdan: Same as above

Atomic/diff.py
```diff
@@ -363,14 +420,29 @@ class DiffFS(object):
...
+            self.left.append(_result['path'])
+
+        def modified(_result): #pylint: disable=unused-variable
```
rhatdan: ditto

Atomic/diff.py
```diff
@@ -455,7 +530,7 @@ def _print_diff(file_list):
             util.write_out("\nFiles only in {}:".format(right_docker_obj))
             _print_diff(self.right)
         if len(self.common_diff):
-            util.write_out("\nCommon files that are different:")
+            util.write_out("\nCommon files that are different: (reason)")
```
rhatdan: Not sure what you are doing here?

bash/atomic
```diff
@@ -554,6 +554,7 @@ _atomic_diff() {
 		--help
 		--json
 	--display
+	--keywords, -k
```
rhatdan: It would be nice to automatically expand the keywords.

docs/atomic-diff.1.md
```diff
@@ -25,6 +25,11 @@ but there are switches for comparing RPMs and metadata as well.
 **--json**
   Output in the form of JSON.
 
+**-k** **--keywords**
```
rhatdan: Give a list of keywords available.

docs/atomic-diff.1.md
```diff
@@ -63,6 +69,10 @@ Compare the files and RPMs (without versions) in images 'foo1' and 'foo2' and ou
 Compare only the metadata between images 'foo1' and 'foo2'
 
     atomic diff -m foo1 foo2
+    
+Compare files by 'sha256digests' and 'time' between images 'foo1' and 'foo2'
+
+    atomic diff foo1 foo2 --keywords sha256digest time
```
rhatdan: I think the handling of multiple keywords does not work well.  What happens if I do


atomic diff --keywords sha256digest time foo1 foo2

Does it work
Or if I had a container named time.

atomic diff --keywords sha256digest time foo1

tests/integration/test_diff.sh
```diff
@@ -34,13 +35,14 @@ teardown () {
 # logic can be added to a "cleanup stack", by cascading function calls
 # within traps. See tests/integration/test_mount.sh for an example.
 trap teardown EXIT
+if [ -e ${GOMTREE} ]; then
```
jlebon: Probably better to indicate a skip here by returning 77.

Also, if you want these tests to run on the Atomic Host platforms until gomtree is pulled in automatically, you can simply add to the YAML:

```yaml
packages:
  - gomtree
```

Atomic/diff.py
baude: @rhatdan I can do 'all' but if we ever add two algo's for digesting, they both would then get used.  Adding for now as it makes sense.

Atomic/diff.py
baude: Fixed.

Atomic/diff.py
baude: Because of the use of eval below, pylint doesnt recognize they are used.

Leaving until better solution.

Atomic/diff.py
baude: Because of the use of eval below, pylint doesnt recognize they are used.

Leaving until better solution.

Atomic/diff.py
baude: Because of the use of eval below, pylint doesnt recognize they are used.

Leaving until better solution.

bash/atomic
baude: I couldn't figure that out ... do you know how to?

tests/integration/test_diff.sh
baude: @jlebon is that done with an exit 77 or elsewise?

jlebon: Yes exactly. See e.g. `test_verify.sh`.

docs/atomic-diff.1.md
baude: @rhatdan 

It only works if if the keywords are the last arguments.  Verified this in argparse as well.  Example:

```
[bbaude@bbaude atomic (diff_mtree)]$ sudo ./atomic diff -k sha256digest centos:latest 19f8d8b7147b
atomic diff: argument -k/--keywords: invalid choice: 'centos:latest' (choose from 'all', 'link', 'nlink', 'mode', 'type', 'time', 'uid', 'gid', 'size', 'sha256digest')
Try 'atomic diff --help' for more information.
```
^^ Fails.  Keywords have to be last.
```
[bbaude@bbaude atomic (diff_mtree)]$ sudo ./atomic diff centos:latest 19f8d8b7147b -k sha256digest
```
Ideas and/or suggestions welcome.

Atomic/diff.py
rhatdan: Ok, I thought they might have been left over.

Atomic/diff.py
rhatdan: How does that happen.  This looks like it will just output 
"(reason)"

baude: That LOC is just to print the header(title).  The rest comes in the _print_diff method.

docs/atomic-diff.1.md
rhatdan: Man page should include list.

docs/atomic-diff.1.md
rhatdan: Lets make it a comma separated list?  Or do it the way we did for TOP?

rhatdan:
```
  -o [{time,stime,ppid,uid,gid,user,group}], --optional [{time,stime,ppid,uid,gid,user,group}]
```

Atomic/diff.py
rhatdan: Do you handle this if there are multiple reasons?

docs/atomic-diff.1.md
twaugh: FWIW using "--" between the keyword list and the image names also works, but that's not very easy to use/discover.

Atomic/diff.py
baude: Yup!

bash/atomic
baude: Great

docs/atomic-diff.1.md
baude: Done

docs/atomic-diff.1.md
baude: Added the ability to use -k one or more times.  One keyword per -k

rhatdan: Why not do the "," separated list?  

-k time -k stime -k ppid

Versus

-k time,ktime,ppid

But I guess it is not that big a deal

Most people will just do -k all, or is this the default?

rhatdan: Ok, I though atomic top did both, but if it use -o foo -o bar, then LGTM

rhatdan:
@rh-atomic-bot r+

PR #778, opened 2016-12-01
[merged] generate: specify storage for mounts

yuqi-zhang:
If duplicate images exist in ostree and docker, mounting images
will fail during generate if storage is not specified.


rhatdan:
I think we should add a --storage flag for this.

Something like this
```
    generate_parser.add_argument("--storage", default=storage, dest="storage",
                               help=_("Specify the storage from which to generate the image manifest from. "
                                      "If not specified and there are images with the same name in "
                                      "different storages, you will be prompted to specify."))

```

THen m.storage=self.args.storage.

yuqi-zhang:
I agree, this will be better for the long run, but since currently generate only works on docker images, should I just have the default be docker and keep the --storage flag hidden?

baude:
I have a couple of comments here.  I'll group them here.

1. We really need to correlate this PR with https://github.com/projectatomic/atomic/pull/777 because there is similar work going on there.

2. The "choices" or keywords between the two PRs should be singular.

3. The generate function (at least) could/should be refactored.  I would also like to see a manifest be created from the image object.  The results of the manifest and where it is stored should be attributes to the image object.  For example:
   db = DockerBackend()
   for img_obj in db.get_images():
      img_obj.generate_manifest()
      print("Manifest results are as follows: \n{}".format(img.obj.manifest_results)
      print("Manifest is stored at: {}".format(img_obj.manifest_results)

Also as a side note, the the __init__ for image, the manifest results should probably be a @property so that it can be loaded if it exists.

4. It would be nice to have our generate methods from the two PRs go hand in hand.  i.e.  generate_manifest and generate_manifests where if you want to do all images, then generate_manifestS calls generate_manifest.

yuqi-zhang:
I initially just intended this to be a temporary fix for the test_verify integration test, as with the new storage flags, the old method fails due to storage not being specified for mount. This was glossed over for awhile since gomtree wasn't included, and thus that test was being skipped.

rhatdan:
Ok I think we should merge this to fix the issue, and then we can fix it the way @baude wants in a different PR.

Need Man page update and bash command completions.

yuqi-zhang:
Do we plan on supporting generate for ostree images? At the moment generate only works on docker images, so if I add --storage to man page and auto-complete, it doesn't make much sense since docker is the only option (hence why I suppress the flag to not even display to the user).

rhatdan:
Well if we move to container/storage we would.  But for now, ostree already takes care of its integrity.

@rh-atomic-bot r+

PR #780, opened 2016-12-02
[merged] Atomic/top.py: Fix options handling in top

baude:
Fix the behaviour for atomic top where if you want to include
additional fields to monitor, you now specificy one field per
-o switch.  For example:

atomic top -o ppid -o time


baude:
@rh-atomic-bot r+

PR #785, opened 2016-12-05
[merged] Unify and refactor atomic verify

baude:
Using our refactoring model, verify is now streamlined. We no longer
compare base images as that is not currently possible for both
V1 and V2 schemas.

Verify will now always look at the release and version labels for
comparison.  Should those labels not exist, it will use the
manifest digest for ostree; and it will use the image IDs for
docker.

### DIFF
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -76,9 +76,10 @@ class DockerBackend(Backend):
 
     def _make_image(self, image, img_struct, deep=False, remote=False):
         img_obj = Image(image, remote=remote)
+        img_obj.backend = self
+
         if not remote:
             img_obj.id = img_struct['Id']
-            img_obj.backend = self
             img_obj.repotags = img_struct['RepoTags']
             img_obj.created = img_struct['Created']
             img_obj.size = img_struct['Size']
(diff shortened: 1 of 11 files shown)


Atomic/backends/_docker.py
```diff
@@ -267,9 +268,11 @@ def get_layer(self, image):
     def get_layers(self, image):
         layers = []
         layer = self.get_layer(image)
+        layer.remote = image.remote
```
rhatdan: Shouldn't this happen in the self.get_layer()

Atomic/backends/_ostree.py
```diff
@@ -140,3 +144,11 @@ def get_layers(self, image):
...
     def get_dangling_images(self):
         return []
+
+    def make_remote_image(self, image):
+        img_obj = self._make_remote_image(image)
+        img_obj.populate_remote_inspect_info()
+        return img_obj
+
+    def _make_remote_image(self, image):
+        return self._make_image(image, None, remote=True)
```
rhatdan: Add carriage return

Atomic/objects/image.py
```diff
@@ -138,6 +138,7 @@ def get_label(self, label):
 
     def populate_remote_inspect_info(self):
         remote_inspect_info = self.remote_inspect()
+        self.backend = self.backend
```
rhatdan: Looks like a bug.

Atomic/backends/_ostree.py
baude: Done

rhatdan:
Tests are failing.

baude:
@rhatdan good to go now

rhatdan:
@rh-atomic-bot r+

PR #788, opened 2016-12-06
[merged] redhat-ci: make testsuites required

jlebon:
In preparation for switching Homu over to using the Red Hat CI statuses,
we define all the current testsuites as required. This means that they
will all have to pass before Homu accepts to merge the PR.


cgwalters:
@rh-atomic-bot r+ 5c2e9fd

PR #790, opened 2016-12-07
[merged] fix get auth from docker.io

runcom:
`docker.io` auth credentials aren't keyed with `docker.io` in `~/.docker/config.json`. Instead, credentials for the Docker Hub are keyed with `https://index.docker.io/v1/` as a special case.

This patch makes sure login isn't prompted at all when pushing and there are auth credentials for `docker.io`.

You can reproduce this bug with:
```sh
$ docker login # this defaults to docker.io if official binary or no --add-registry
[ fill in the login info which will be stored under $HOME/.docker/config.json ]

$ atomic push docker.io/runcom/busybox:latest
[ ERR: you are prompted for username and password when it should not cause you're already logged in ]
```

/cc @rhatdan 

Signed-off-by: Antonio Murdaca <runcom@redhat.com>


rhatdan:
@rh-atomic-bot r+

PR #792, opened 2016-12-08
[merged] Refactor containers verb

baude:
With the exception of fstrim, the containers verb has now been
refactored.  It primarily now uses the containers object in its
implementation.

### DIFF
--- a/Atomic/backends/_ostree.py
+++ b/Atomic/backends/_ostree.py
@@ -26,18 +26,20 @@ class OSTreeBackend(Backend):
 
     def _make_container(self, info):
         container_id = info['Id']
-
         runtime = self.syscontainers.get_container_runtime_info(container_id)
-
         container = Container(container_id, backend=self)
         container.name = container_id
+        container.command = info['Command']
         container.id = container_id
+        container.image_name = info['Image']
+        container.image_id = info['ImageID']
         container.created = info['Created']
-        container.status = runtime['status']
+        container.status = container.state = runtime['status']
         container.input_name = container_id
         container.original_structure = info
         container.deep = True
         container.image = info['Image']
+        container.running = False if container.status == 'inactive' else True
 
         return container
 
@@ -142,7 +144,8 @@ class OSTreeBackend(Backend):
             layers.append(layer)
         return layers
 
-    def get_dangling_images(self):
+    @staticmethod
+    def get_dangling_images():
         return []
 
     def make_remote_image(self, image):
@@ -152,3 +155,7 @@ class OSTreeBackend(Backend):
 
     def _make_remote_image(self, image):
         return self._make_image(image, None, remote=True)
+
+    def delete_container(self, container, force=False):
+        return self.syscontainers.uninstall(container)
+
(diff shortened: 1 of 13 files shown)


Atomic/backends/_ostree.py
```diff
@@ -26,18 +26,20 @@ def backend(self):
...
-
         container = Container(container_id, backend=self)
         container.name = container_id
+        container.command = info['Command']
         container.id = container_id
+        container.image_name = info['Image']
+        container.image_id = info['ImageID']
         container.created = info['Created']
-        container.status = runtime['status']
+        container.status = container.state = runtime['status']
```
rhatdan: Why do we have container.status and container.state?

Atomic/backendutils.py
```diff
@@ -62,7 +62,7 @@ def get_backend_and_image(self, img, str_preferred_backend=None):
         raise ValueError("Found {} in multiple storage backends: {}".
                          format(img, ', '.join([x.backend for x in img_in_backends])))
 
-    def get_backend_for_container(self, container, str_preferred_backend=None):
+    def get_backend_and_container(self, container, str_preferred_backend=None):
```
rhatdan: Confusing name of method.  You are passing in "container" and returning con_obj.  Either
container should be renamed to container_id, or function should be get_backend_and_container_object

Atomic/backendutils.py
```diff
@@ -75,16 +75,18 @@ def get_backend_for_container(self, container, str_preferred_backend=None):
...
         container_in_backends = []
         for backend in backends:
             be = backend()
-            if be.has_container(container):
-                container_in_backends.append(be)
+            con_obj = be.has_container(container)
+            if con_obj:
+                container_in_backends.append((be, con_obj))
         if len(container_in_backends) == 1:
             return container_in_backends[0]
```
rhatdan: Is this multiple objects?

Atomic/containers.py
```diff
@@ -66,6 +68,9 @@ def cli(subparser):
 
 class Containers(Atomic):
 
+    FILTERALES= {"container": "id", "image": "image_name", "command": "command",
```
rhatdan: FILTERALES?  Bad name.

Atomic/backends/_ostree.py
baude: Depending on how the struct information is gather (i.e. inspect vs containers), they have State and/or Status and they differ.  In ostree, we set both to be the same and handle it.

Atomic/backendutils.py
baude: Done

Atomic/containers.py
baude: Fixed, although that was the previous name as well.  Now called FILTER_KEYWORDS

Atomic/backendutils.py
rhatdan: Function returns two objects.  This seems to be returning 1?

baude: No, it returns an instance of the backend and a container object seen there on line 89.

rhatdan:
If tests pass, I will merge.

baude:
@rhatdan looks like i got it this time.

rhatdan:
@rh-atomic-bot r+

PR #793, opened 2016-12-08
[merged] Atomic/diff.py: Fix options bug

baude:
When I added the 'all' option to the keywords and messed around with
the keywords, I created a small bug where I checked for a value in
a python list like: 'all' in list which failed in the case that the
keywords was None.

Trivial fix and passes tests


rhatdan:
@rh-atomic-bot r+

PR #794, opened 2016-12-09
[merged] generate an UUID value for system containers

giuseppe:


Atomic/syscontainers.py
```diff
@@ -419,6 +420,8 @@ def _do_checkout(self, repo, name, img, upgrade, values, destination, unitfileou
                 key, val = i[:split], i[split+1:]
                 values[key] = val
 
+        if "UUID" not in values:
+            values["UUID"] = str(uuid.uuid4())
```
rhatdan: Does this generate the UUID IN a form that systemd likes?  Or in the Docker format?

giuseppe: it generates an UUID in the NNNNNNNN-NNNN-NNNN-NNNN-NNNNNNNNNNNN format that is the format expected by systemd for $container_uuid

Atomic/syscontainers.py
```diff
@@ -591,7 +594,7 @@ def update(self, name):
         values = info["values"]
         revision = info["revision"] if "revision" in info else None
 
-        if revision and self.args.setvalues is None:
+        if revision and self.args.setvalues is None and not self.args.force:
```
giuseppe: I added a fixup patch, that move the not force condition.  And took the occasion to add a comment, for what we are doing here

Atomic/syscontainers.py
rhatdan: great

rhatdan:
@rh-atomic-bot r+

PR #795, opened 2016-12-09
[merged] Add --all to images delete

baude:
To make containers delete and images delete consistent, we have
added an --all option to images delete.


rhatdan:
@rh-atomic-bot r+

PR #798, opened 2016-12-12
[merged] Refactor atomic stop

baude:
Refactoring of atomic stop to use container objects and the new backend
methods.

### DIFF
--- a/Atomic/stop.py
+++ b/Atomic/stop.py
@@ -1,7 +1,7 @@
 import argparse
 
 from . import util
-from .util import DockerObjectNotFound
+from Atomic.backendutils import BackendUtils
 
 try:
     from . import Atomic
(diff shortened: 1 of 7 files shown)


Atomic/stop.py
```diff
@@ -16,38 +16,33 @@ def cli(subparser):
...
-        if args:
-            cmd = self.gen_cmd(args + self.quote(self.args.args))
+
+        if self.args.debug:
+            util.write_out(str(self.args))
+
+        beu = BackendUtils()
+        be, con_obj = beu.get_backend_and_container_obj(self.args.image, storage)
+        label_stop_args = con_obj.get_label('stop')
+        if label_stop_args and be.backend == 'docker':
```
rhatdan: Why is this specific to docker?

rhatdan:
Everything failed?

Atomic/stop.py
baude: Long term it should be ... but this would require splicing and dicing the Atomic class because we use two methods in Atomic to format out args.  I'm not ready to dice that up yet.  Think we can forego it for now?

rhatdan:
@rh-atomic-bot r+

PR #799, opened 2016-12-13
Atomic.run: allow atomic run to pull image without TLS authentication

chuanchang:
At present, the atomic run uses update() to update image, but w/o passing
'--tls-verify=false' argument to update(), which directly call
util.skopeo_copy() and also w/o '--tls-verify=false' argument setting.

This patch is a simple fix, and only replace update() w/ pull(), it looks
like docker run command, if the image doesn't exist then pull the image
automatically.

BZ: 1401842

### DIFF
--- a/Atomic/run.py
+++ b/Atomic/run.py
@@ -86,7 +86,7 @@ class Run(Atomic):
             if self.args.display:
                 return self.display("Need to pull %s" % self.image)
 
-            self.update()
+            self.pull()
             self.inspect = self._inspect_image()
 
         if self.spc:


chuanchang:
We should install python-ipaddress package in centos/7/atomic .

ERROR: test_discovery (unittest.loader.ModuleImportFailure)
ImportError: Failed to import test module: test_discovery
  File "/usr/lib64/python2.7/unittest/loader.py", line 252, in _find_tests
    module = self._get_module_from_name(name)
  File "/usr/lib64/python2.7/unittest/loader.py", line 230, in _get_module_from_name
    __import__(name)
  File "/var/tmp/checkout/tests/unit/test_discovery.py", line 2, in <module>
    from Atomic import util
  File "Atomic/__init__.py", line 1, in <module>
    from .pulp import PulpServer, PulpConfig
  File "Atomic/pulp.py", line 10, in <module>
    from . import util
  File "Atomic/util.py", line 21, in <module>
    import ipaddress
ImportError: No module named ipaddress

rhatdan:
So we need to add that package as a requires for atomic package.

Looks like we have a conflict for this package.

yuqi-zhang:
I updated that to use "pull" at https://github.com/projectatomic/atomic/commit/5cea34fc4a7586543b51fafd7114ee7d34a40f31

rhatdan:
@baude @chuanchang Is this still true?  We lost this in the noise.  Is this still something that needs to be fixed?

baude:
The latest code uses pull.

https://github.com/projectatomic/atomic/blob/master/Atomic/run.py#L108

I'd be curious if the issue is still observed?

rhatdan:
@chuanchang can we close this?

PR #802, opened 2016-12-20
Docker backend fixes

yuqi-zhang:
Allow "docker daemon not running" error when fetching
containers/images. Also fix parsing of docker state.


rhatdan:
@baude PTAL

Atomic/backendutils.py
```diff
@@ -41,9 +41,12 @@ def get_backend_and_image_obj(self, img, str_preferred_backend=None):
         # Check preferred backend first
         if str_preferred_backend:
             be = self.get_backend_from_string(str_preferred_backend)
-            img_obj = be.has_image(img)
-            if img_obj:
-                return be, img_obj
+            try:
```
rhatdan: How about we pass an option to the be saying that it should ignore nodaemon errors?  I don't want to accidently eat errors when we want them reported to the user?

rhatdan: be = backend(IgnoreNoDaemon=True) 
...

yuqi-zhang: I originally intended this patch to replicate behavior in the old (separated) backend, where if the docker daemon wasn't running, the commands would just display blank for docker images/containers and not error.

With the exception of generate (which requires gomtree) every other image/container command supports ostree images/containers. I could pass the option to ignore NoDaemon errors, but for all these commands, there is no "error" for docker daemon not running. Alternatively I could display a warning?

yuqi-zhang:
I added a flag no_docker_daemon, which I currently defaulted to true since all usages of the functions (i.e. info/delete/list) should work without the docker daemon running. This way future docker-specific functions for instances of BackendUtils can make no_docker_daemon=True calls. Or is that not a use case and it would be better to have it be an instance variable for BackendUtils?

yuqi-zhang:
Any updates @rhatdan @baude ? I'd like to get this merged since many system container functions cannot be used without docker running right now, this causes annoying issues such as: cannot install docker as a system container unless docker is running, which makes no sense.

rhatdan:
@rh-atomic-bot r+

yuqi-zhang:
Closed in favor of https://github.com/projectatomic/atomic/pull/847

PR #809, opened 2017-01-03
[merged] Dbus verify

baude:
Atomic/verify.py: Fix dbus implementation of image verify
    
The dbus verify is now similar to the tty version in how it
processes the request.  We now return a list of layers and
their comparisons.

### DIFF
--- a/atomic_dbus.py
+++ b/atomic_dbus.py
@@ -545,17 +545,16 @@ class atomic_dbus(slip.dbus.service.Object):
     # The Verify method takes in an image name and returns whether or not the
     # image should be updated
     @slip.dbus.polkit.require_auth("org.atomic.read")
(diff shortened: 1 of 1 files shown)


baude:
I'll squash the wip and last commit if the PR looks good.

Atomic/verify.py
```diff
@@ -74,16 +74,18 @@ def verify(self):
             util.write_out("\n")
 
     def verify_dbus(self):
+        print(self.args)
```
rhatdan: Remove or add if debug.

baude:
atomic image verify as it is currently coded is only one image at a time.

rhatdan:
@rh-atomic-bot r+

PR #810, opened 2017-01-05
[merged] docs: add missing document into atomic images

chuanchang:
atomic images generate command is available now, so add document into man page of atomic images, and rearrange existing commands in man page according to alphabet order.


rhatdan:
@rh-atomic-bot r+

PR #818, opened 2017-01-09
(DONTMERGE) Allow developers to specify atomic.type label in container image

rhatdan:
If the atomic command sees this flag set then it will automatically install
based on the system type.

### DIFF
--- a/Atomic/atomic.py
+++ b/Atomic/atomic.py
@@ -665,10 +665,13 @@ class Atomic(object):
                 img += "/{}".format(repo)
             img += "/{}:{}".format(image, tag)
             return img
-        if not registry:
-            ri = RegistryInspect(registry, repo, image, tag, debug=self.args.debug, orig_input=self.image)
-            return ri.find_image_on_registry()
+        ri = RegistryInspect(registry, repo, image, tag, debug=self.args.debug, orig_input=self.image)
+        return ri.find_image_on_registry()
 
+    def remote_inspect(self, image=None):
+        if image == None:
+            image=self.image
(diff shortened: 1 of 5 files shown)


rhatdan:
I have not been able to test, but @giuseppe and @baude PTAL.

I think this should work if user specifies 
```
atomic install --system=False etcd
```
Then even if the etcd container has a label atomic.type="system", it will not install a system container.

giuseppe:
what if we use this new label at `pull` time so to choose the correct storage?

we can then change `install`to check for a system image if there is no docker image with that name, --system will force it to skip the check and directly install the system container.

If an install needs to pull the image first, as there is no docker or system image stored, then we can check for what storage was used by pull before installing it.

rhatdan:
So you want atomic.storage label, where storage could be `docker, ostree,  or container`?

rhatdan:
I have partially done what you want.   But just because the container is stored in ostree, does that indicate that it is a `--system` container?

giuseppe:
@rhatdan, right that could be a problem in future.  Right now, `--storage ostree` means that the container is a system container (or user when using it as not root user).

Should we still keep the label `atomic.type` and infer the storage from that instead of the other way around?  `atomic.type=docker` -> `storage=docker`, `atomic.type=system` -> `storage=ostree`

rhatdan:
@giuseppe we are having a meeting at 1:00 EST to discuss storage for CRI-O I just invited you to it.

rhatdan:
This is handled by other pull requests.

Issue #811, opened 2017-01-05
use hostname/URI validation when adding trust policies


miabbott:
Before I RTFM, I tried to modify the default trust policy with the following:

`# atomic trust add default -t reject`

Which was accepted happily and I saw this:

```
# atomic trust show                                                                                                                                                                          
* (default)                         accept                               
default                             reject
```

Going further down the rabbit hole, I found I could basically give `atomic trust` any string I wanted for a registry name and it would eat it up.

```
# atomic trust add "my[awful]registry(string)" -t reject
# atomic trust show
* (default)                         reject                               
my[awful]registry(string)           reject
# atomic trust add this.is.not.a.real.url.docker.registry.z -t insecureAcceptAnything
# atomic trust show
* (default)                         reject                               
my[awful]registry(string)           reject                                                    
this.is.not.a.real.url.docker.registry.z accept
```

A little bit of input validation would be helpful for the user.

rhatdan:
I guess we need to determine a valid string.

@aweiteka PTAL

aweiteka:
While this is incredibly ugly and we need to validate, all of these rules will be quietly ignored since they will never match.

This one--this.is.not.a.real.url.docker.registry.z--**is** technically a valid registry URI. We should constrain on this character list: `[a-zA-Z1-9.:-_]`. Need to confirm.

rhatdan:
Should we validate the URI by appending http:// on the front and running it through a function like:
```
try:
    import urllib2
    urlparse = urllib2.urlparse.urlparse
except:
   import urllib.parse
   urlparse=urllib.parse.urlparse
   
min_attributes = ('scheme', 'netloc')

def is_valid(url, qualifying=None):
    qualifying = min_attributes if qualifying is None else qualifying
    token = urlparse(url)
    return all([getattr(token, qualifying_attr)
                for qualifying_attr in qualifying])
```

aweiteka:
@miabbott validation landed in #816. The basic list of valid/invalid trust policy keys is here: https://github.com/projectatomic/atomic/blob/master/tests/unit/test_util.py#L84

Please confirm/close.

miabbott:
I found some other cases that don't seem to be valid by my read, but I don't want to drag this out, either.

Like this:

```
# ./atomic trust add -t insecureAcceptAnything private.registry.url:aaa
# ./atomic trust add -t insecureAcceptAnything 9000:9000:9000:9000:9000:9000:9000:9000:9000:9000
```

[...]

PR #819, opened 2017-01-10
[merged] Extend image URI validation

aweiteka:
More better. Resolves https://github.com/projectatomic/atomic/issues/811#issuecomment-271406453


rhatdan:
@rh-atomic-bot r+

PR #825, opened 2017-01-16
[merged] Refactor Pull, Update, Install, Run

baude:
Refactor several of the atomic verbs and subverbs to take advantage
of object refactoring.

Also, do not pull images with skopeo if the local image is already
at the latest.

$ sudo python ./atomic --debug pull busybox
Namespace(_class=<class 'Atomic.pull.Pull'>, assumeyes=False, debug=True, func='pull_image', image='busybox', reg_type=None, storage='docker')
Latest version of busybox already present.


baude:
@giuseppe would you mind peeking at the system_container test failure? Maybe you can hint me on whats wrong there and I can fix it.

yuqi-zhang:
The tests are failing due to the "secret" value that was set for "$RECEIVER" during install was not set, so $RECEIVER took the default value of "world". From a glance "self.args.setvalues" (which contained the secret value) was overridden at some point before it reached _do_checkout in syscontainers.py

Atomic/backends/_docker.py
```diff
@@ -191,26 +210,31 @@ def start_container(self, name):
...
+            if self.already_has_image(local_image, remote_image):
+                raise ValueError("Latest version of {} already present.".format(image))
+        registry, _, _, tag, _ = util.Decompose(fq_name).all
+        image = "docker-daemon:{}".format(image)
+        if not image.endswith(tag):
+            image += ":{}".format(tag)
+        insecure = True if util.is_insecure_registry(self.d.info()['RegistryConfig'], util.strip_port(registry)) else False
         trust = Trust()
-        trust.set_args(pull_args)
+        #trust.set_args(pull_args)
```
rhatdan: Why is this commented out?

Atomic/backends/_docker.py
```diff
@@ -247,14 +265,17 @@ def _ping(self):
     def delete_image(self, image, force=False):
         return self.d.remove_image(image, force=force)
 
-    def update(self, name, force=False):
+    def update(self, name, force=False, **kwargs):
+        debug = kwargs.get('debug', False)
+        # Only delete containers if a new image is actually pulled.
```
rhatdan: Not sure what this comment means?

rhatdan:
Tests are failing

rhatdan:
@giuseppe PTAL

Atomic/objects/image.py
```diff
@@ -56,7 +60,11 @@ def __gt__(self, other):
         pass
 
     def __eq__(self, other):
-        if self.long_version == other.long_version:
+        if self.long_version == other.long_version \
+                and not any([True for x in [self.long_version, other.long_version] if x is None]) \
```
giuseppe: is this `None not in [self.long_version, other.long_version]` ?

Atomic/objects/image.py
```diff
@@ -56,7 +60,11 @@ def __gt__(self, other):
...
+                and not any([True for x in [self.long_version, other.long_version] if x is '']):
```
giuseppe: and this `'' not in [self.long_version, other.long_version]` ?

giuseppe:
I agree with @yuqi-zhang on this, could you please verify why the args are not passed correctly?

Also, it is a huge patch, I think it will make sense to split it in one for each refactored command.  It will make easier to spot the problem.

Atomic/backends/_docker.py
baude: The comment was misplaced a little. I have moved it down.  If you use update with --force, existing containers based on the old image are deleted.  So in the case that a new image is not pulled, the containers should not be deleted.

Atomic/backends/_docker.py
baude: Aaron and I speculated that these are no longer needed.  I commented it out to test it.  Just need to delete them.  Will do now.

Atomic/objects/image.py
baude: Yes much better

Atomic/objects/image.py
baude: Yes much better

baude:
Im looking into the tests.  Because all of these commands are so linked, it does not work well to split up frankly.  I would prefer to keep them together.  if that is going to keep it from being merged, Ill see what I can do.

miabbott:
If `docker` is not running, `atomic pull` throws a cryptic error; could be improved:

```
# systemctl status docker
● docker.service - Docker Application Container Engine
   Loaded: loaded (/usr/lib/systemd/system/docker.service; disabled; vendor preset: disabled)
   Active: inactive (dead) since Tue 2017-01-17 19:45:10 UTC; 3s ago
     Docs: http://docs.docker.com
  Process: 19000 ExecStart=/usr/bin/dockerd-current --add-runtime oci=/usr/libexec/docker/docker-runc-current --default-runtime=oci --containerd /run/containerd.sock --exec-opt native.cgroupdriver=systemd --user
 Main PID: 19000 (code=exited, status=0/SUCCESS)
# pwd
/root/atomic
# git rev-parse HEAD
f9a69999e3e7f2ba27fa033389577ed3dfed8aeb
# ./atomic pull docker.io/nginx
('Connection aborted.', error(2, 'No such file or directory'))
```

miabbott:
The user gets different errors when trying to pull an image that does not exist and choosing the different storage backend

```
# ./atomic pull --storage ostree foobar
ReturnTuple(return_code=1, stdout='', stderr='time="2017-01-17T19:59:34Z" level=fatal msg="unauthorized: authentication required" \n')
# ./atomic pull foobar
Unable to resolve foobar
```

miabbott:
Not sure how this should work, but it looks off...

When using `atomic run` to run a container that is already running, the command appears to start a new container with a different ID.  However, the output from `atomic containers list` shows that the previously running container is intact and its ID has not changed.

I would suggest changing the output to indicate that the container is already running or otherwise notifying the user that the new container ID is not going to be used.

```
# ./atomic containers list
   CONTAINER ID IMAGE                COMMAND              CREATED          STATE     RUNTIME   
   d7843c1a7f8c docker.io/cockpit/ws /container/atomic-ru 2017-01-17 19:59 running   docker    
# ./atomic run docker.io/cockpit/ws
/usr/bin/docker run -d --privileged --pid=host -v /:/host docker.io/cockpit/ws /container/atomic-run --local-ssh

INFO: --pid=host 
      Processes in this container can see and interact with all processes on the host and disables SELinux within the container.

INFO: --privileged 

50736975bed2f07fa9eeea91228d57340b3feeb13118e74c9f8aba40a37a7a70
# ./atomic containers list
   CONTAINER ID IMAGE                COMMAND              CREATED          STATE     RUNTIME   
   d7843c1a7f8c docker.io/cockpit/ws /container/atomic-ru 2017-01-17 19:59 running   docker    
# ./atomic run docker.io/cockpit/ws
/usr/bin/docker run -d --privileged --pid=host -v /:/host docker.io/cockpit/ws /container/atomic-run --local-ssh

INFO: --pid=host 
      Processes in this container can see and interact with all processes on the host and disables SELinux within the container.

INFO: --privileged 

5ad399a88ebdfb3c8cd039d622a88e224844b2f927f704546bc9785f0e05616b
# ./atomic containers list
   CONTAINER ID IMAGE                COMMAND              CREATED          STATE     RUNTIME   
   d7843c1a7f8c docker.io/cockpit/ws /container/atomic-ru 2017-01-17 19:59 running   docker
```

rhatdan:
Yes the second atomic run on the same container will just exec into the container if it is already running.  So it should indicate that it is joining the existing container.

baude:
Per agreement with @rhatdan and @miabbott , I will merge this and then xfer each of the identified issues and resolve them separately.

baude:
@rh-atomic-bot r+

rhatdan:
make PYTHON=python3 install

is blowing up.
************* Module Atomic.util
E:466,76: Module 're' has no 'MULTILINE' member (no-member)
E:483,67: Module 're' has no 'MULTILINE' member (no-member)
E:579,17: Module 're' has no 'UNICODE' member (no-member)

baude:
@rhatdan Im not able to replicate that.  Is it rawhide maybe?

rhatdan:
What version of python3-pylint do you have?

Issue #829, opened 2017-01-17
pulling with --storage ostree doesn't enforce signatures


aweiteka:
When running atomic pull with ostree storage backend signature policy is not enforced because the (skopeo layers command)[https://github.com/projectatomic/atomic/blob/master/Atomic/util.py#L295] doesn't enforce signatures (see https://github.com/projectatomic/skopeo/issues/287). Is there a reason we're not using `skopeo copy` here? https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L1172

/cc @giuseppe

rhatdan:
I don't think skopeo supports an ostree backend?  @runcom

runcom:
@rhatdan it doesn't, it relies on a-now-deprecated command called "skopeo layers" to import individual layer to an ostree repo on a system.

rhatdan:
Right so to get skopeo copy to work we would need to have a destination like ostree:/var/lib/containers
and put the code to install the layers into skopeo correct?

runcom:
Yeah, that would be nice, @giuseppe what do you think? I remember some conversion with @giuseppe and @mtrmac somewhere but I couldn't find the link anywhere right now.

mtrmac:
A low-effort fix might be to get https://github.com/containers/image/pull/159 + a corresponding `skopeo` enhancement, and use `skopeo copy --layer-subset` instead of `skopeo layers`. That would allow us to preserve the existing Python code, at the cost of having to be really careful so that we always do verification, and only pull exactly one image (I suppose, first pull the manifest + no layers, and then use digest references, using a digest computed from that manifest, for all subsequent operations).

Alternatively we could build an `ostree` `ImageTransport` directly into containers/image in Go (and drop the Python code from `atomic`?).

rhatdan:
I would prefer solution #2 
we could build an ostree ImageTransport directly into containers/image in Go (and drop the Python code from atomic?).

runcom:
The key functionality we need to port from python to go in containers/image is the ability to check for already pulled layers to do incremental pulls on image update (@giuseppe can tell us more about it I guess). But with the changes in c/image such as HasBlob we should be able to maintain that compatibility.

cgwalters:
I'd at some point like to be able to pull OCI from ostree, or at least rpm-ostree.  See discussion in https://github.com/ostreedev/ostree/pull/619

rhatdan:
That would be very cool

mtrmac:
[...]

PR #830, opened 2017-01-17
WIP: use skopeo copy instead of layer

aweiteka:
Address #829 

This is the path of least resistance to deprecate skopeo layer call. I'm open to a better solution.


aweiteka:
Due to the expense of downloading the whole image when only some layers may be needed, closing this workaround in favor of solutions discussed in #829 

/cc @giuseppe

PR #836, opened 2017-01-18
[merged] containers: use BACKEND instead of RUNTIME

giuseppe:
We may need to differentiate between runtime and storage, but for now,
just rename RUNTIME to BACKEND.  Otherwise it will show system
containers with RUNTIME=ostree.

Add an alias, so it doesn't break existing scripts.

### DIFF
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -84,7 +84,7 @@ def cli(subparser):
 class Containers(Atomic):
 
     FILTER_KEYWORDS= {"container": "id", "image": "image_name", "command": "command",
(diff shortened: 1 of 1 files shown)


rhatdan:
@rh-atomic-bot r+

rhatdan:
@jlebon Is rh-atomic-bot broken?

jlebon:
@rh-atomic-bot r+ 750df5d

cgwalters:
The queue is processing now it looks like.  http://escher.verbum.org:54856/queue/atomic

PR #837, opened 2017-01-18
[merged] Atomic/delete.py: bz 1413887 Delete image fails

baude:
Under certain conditions when trying to delete an image, a exception
occured because obj.repotags is not populated as a list.  This is OK
but wasnt being handled correctly.  New conditionals to use obj.id when
this happens.


rhatdan:
@rh-atomic-bot r+

Issue #838, opened 2017-01-18
atomic run regression


jlebon:
Looks like https://github.com/projectatomic/atomic/pull/825 broke the way run args are handled. From our test logs:

```
18:22:25 TASK [atomic_run_verify : atomic run centos_httpd] *****************************
18:22:26 fatal: [10.8.169.74]: FAILED! => {"changed": true, "cmd": ["atomic", "run", "centos_httpd"], "delta": "0:00:00.704818", "end": "2017-01-17 23:22:25.840092", "failed": true, "rc": 1, "start": "2017-01-17 23:22:25.135274", "stderr": "Cannot find file: `-D`", "stdout": "-D FOREGROUND", "stdout_lines": ["-D FOREGROUND"], "warnings": []}
```

Where `centos_httpd` is: https://github.com/projectatomic/atomic-host-tests/blob/master/roles/docker_build_httpd/files/centos_httpd_Dockerfile

It seems like it's using the first arg of Cmd, even when there's an entrypoint defined.

rhatdan:
@baude PTAL

baude:
@jlebon when you get a chance to see if this is good to go, mind closing this?

jlebon:
I will take @miabbott's word for it. :)

PR #842, opened 2017-01-19
[merged] Atomic/objects/image.py : Fix run when RUN label is present

baude:
The RUN label was not being used correctly by Atomic CLI and therefore
resulted in incorrect runargs being used. We should be leaning on using
the cmd and entry points as defined and not mess with those in the
building of the command itself.

This was reported in
issue https://github.com/projectatomic/atomic/issues/838.


rhatdan:
@rh-atomic-bot r+

PR #843, opened 2017-01-19
[merged] More run regressions

baude:


Atomic/objects/image.py
```diff
@@ -157,7 +157,9 @@ def populate_remote_inspect_info(self):
         self.labels = remote_inspect_info.get("Labels", None)
         self.release = self.get_label('Release')
         self.version = self.get_label('Version')
-        self.id = no_shaw(remote_inspect_info['id'])
+        self.id = remote_inspect_info['id']
+        if remote_inspect_info['id']  is not None:
```
rhatdan: Why not 
if not self.id

baude: The syntax wouldn't be right.  I did change to:
```
if self.id is not None:
```

rhatdan:
@rh-atomic-bot r+

miabbott:
Well, this has already been merged, but I ran a couple of tests with `atomic pull` and `atomic run` without issue.  Appears to fix #838 and #844

rhatdan:
Please close them if you are happy.

PR #847, opened 2017-01-24
Backend missing

baude:
Disconnect backends
    
    Ideally, the atomic CLI should be able to operate independently
    of the backends it supports.  For example, if dockerd is inactive,
    the ostree backend and atomic cli should still work.
    
    This requires some tweaking to the backendutils code and the work
    flow.  We also need to specifically know if the user passes
    --storage so that we treat that as an explicit override.  The work
    flow is now roughly:
    
    * a default storage can be defined in atomic.conf (was always this way)
    * if not defined, defaults to docker.
    * if --storage is passed, treat explictly and fail if cannot execute
    * if no --storage is specified, use default.  if default is not available, move
     onto the next backend

### DIFF
--- a/Atomic/run.py
+++ b/Atomic/run.py
@@ -89,7 +89,7 @@ class Run(Atomic):
                 db.pull_image(self.image)
                 img_object = db.has_image(self.image)
             except RegistryInspectError:
-                util.write_err("Unable to find image {}".format(self.image))
+                raise ValueError("Unable to find image {}".format(self.image))
 
         db.run(img_object, atomic=self, args=self.args)
 
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -22,7 +22,14 @@ class DockerBackend(Backend):
     def __init__(self):
         self.input = None
         self._d = None
-        self._ping()
+
+    @property
+    def available(self):
+        try:
+            _ = self.d
+            return True
+        except util.NoDockerDaemon:
+            return False
 
     @property
     def d(self):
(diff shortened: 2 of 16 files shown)


baude:
@rhatdan @giuseppe @yuqi-zhang  PTAL

Atomic/run.py
```diff
@@ -90,7 +90,7 @@ def run(self):
                 img_object = db.has_image(self.image)
             except RegistryInspectError:
                 util.write_err("Unable to find image {}".format(self.image))
-
+                return 1
```
giuseppe: what does the 1 here mean?

baude: mistake, should be a ValueError

yuqi-zhang:
`atomic images prune` is still using the old method of going through all backends
`atomic images info` on an ostree image returns: This Atomic function requires an active docker daemon. Even if "--storage ostree" is specified.

baude:
@yuqi-zhang Yes, that is going to be the case for a little while because we have to use dockerd to obtain the list of registries to search through when a fully-qualified image name is NOT provided.  I have some ideas about trying to incorporate this into skopeo.

yuqi-zhang:
Ok, other than `info` and `prune` everything else looks good. I will close https://github.com/projectatomic/atomic/pull/802 when we merge this.

baude:
@rh-atomic-bot r+

baude:
@rh-atomic-bot r+ 96cc5fb

baude:
@rh-atomic-bot retry

PR #848, opened 2017-01-25
Fixup trust tests

baude:
There is a python bug that in 2.7.5 that is causing
the unittest failures on centos_atomic.  The bug
is related to how regex is compiled and used.

Adding a simple decorator to skip tests that fail
because of that in the trust tests.


baude:
@rh-atomic-bot retest

baude:
@rh-atomic-bot r+

baude:
@rh-atomic-bot r+ e3dfb1d

PR #850, opened 2017-01-26
Test framework

baude:
A series of small commits that fix or enhance integration testing.

### DIFF
--- a/Atomic/storage.py
+++ b/Atomic/storage.py
@@ -7,7 +7,6 @@ from . import util
 from .Export import export_docker
 from .Import import import_docker
 from .util import NoDockerDaemon, default_docker_lib
-import subprocess
 
 try:
     from subprocess import DEVNULL  # pylint: disable=no-name-in-module
--- a/Atomic/storage.py
+++ b/Atomic/storage.py
@@ -7,6 +7,7 @@ from . import util
 from .Export import export_docker
 from .Import import import_docker
 from .util import NoDockerDaemon, default_docker_lib
+import subprocess
 
 try:
     from subprocess import DEVNULL  # pylint: disable=no-name-in-module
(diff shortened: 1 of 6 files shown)


test.sh
```diff
@@ -113,15 +113,16 @@ make_docker_images () {
         if [[ -d "${df}.d" ]]; then
             cp -r "${df}.d" "${WORK_DIR}/${BASE_NAME}.d"
         fi
-
+	SECONDS=0
         ${DOCKER} build -t ${iname} -f ${df_cp} ${WORK_DIR} >> ${LOG}
+	DURATION=$SECONDS
```
aweiteka: When is DURATION ever updated? Won't it always be 0?

Atomic/storage.py
```diff
@@ -63,6 +63,9 @@ def cli(subparser):
                          help=_("remove all unused block devices from storage pool"))
     modifyp.add_argument('--driver', dest="driver", default=None, help='The storage backend driver', choices=['devicemapper', 'overlay', 'overlay2'])
     modifyp.add_argument('--vgroup', dest="vgroup", default=None, help='The storage volume group')
+    modifyp.add_argument("--graph", dest="graph",
+                        default=default_docker_lib(),
+                        help=_("Root of the Docker runtime (Default: %s)" % default_docker_lib()))
```
aweiteka: Is this related to the test work? Looks like the opt was documented but never implemented?

test.sh
baude: No it is a special bash function.  Seems to work:

```
	Built   : atomic-test-1 in 16 seconds
	Built   : atomic-test-3 in 9 seconds
	Built   : atomic-test-4 in 7 seconds
	Built   : atomic-test-5 in 8 seconds
	Built   : atomic-test-6 in 17 seconds
	Built   : atomic-test-secret in 7 seconds
	Built   : atomic-test-system in 22 seconds
```

Atomic/storage.py
baude: Right, I agree.  Indeed it actually causes a test failure.

baude:
@rh-atomic-bot r+

baude:
@rh-atomic-bot r+ eed4861

PR #851, opened 2017-01-26
Use secondary disks for storage integration tests

baude:
Due to a couple of bugs related to loopback and sfdisk, we should
be using available secondary disks for testing atomic storage.  This
is part of the destructive testing we now do.

Also, configured vagrant to dispatch two disks so the tests can be done
there too.

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -42,6 +42,7 @@ host:
   distro: centos/7/atomic/continuous
   specs:
     secondary-disk: 10
+    secondary-disk: 10
 
 context: centos/7/atomic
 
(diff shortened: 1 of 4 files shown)


baude:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot r+

mvollmer:
Since the test is now allowed to destroy the system, I think we can remove all of `teardown`, actually.

baude:
@mvollmer , @rhatdan This is very close to passing now.  The centos passes tests.  Can either of you tell why fedora/24/atomic and fedora/25/atomic don't?  It looks like the disk configuration might be a little different?  If you look at the test_storage output there is info like mount, fdisk -l, and so on to help for debug.

baude:
@rh-atomic-bot r+  	938a8b6

baude:
@rh-atomic-bot retest

baude:
@rh-atomic-bot r+ 9469733

jlebon:
Good stuff Brent. I've switched the bot over to use the new tests, so they shouldn't break again.

PR #852, opened 2017-01-26
Atomic/mount.py: Use default_docker_lib()

baude:
Instead of using the hardcoded path to /var/lib/[docker, docker-latest],
we should use the helper method default_docker_lib() in util.  This
should allow overlay2 to work with both mount and scan.

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -403,15 +403,15 @@ class DockerMount(Mount):
 
     @staticmethod
     def _no_gd_api_dm(cid):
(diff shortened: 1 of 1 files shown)


rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot r+00d0e5c

rhatdan:
@rh-atomic-bot r+

PR #853, opened 2017-01-27
Inline pubkeys in policy.json

aweiteka:
Default is now to inline pubkeys as "[keyData](https://godoc.org/github.com/containers/image/signature#NewPRSignedByKeyData)" strings inside policy.json, with option to reference as installed files ("[keyPath](https://godoc.org/github.com/containers/image/signature#NewPRSignedByKeyPath)") on host system, the previous behavior.

We also now support passing in a URL to a public key.

### DIFF
--- a/Atomic/trust.py
+++ b/Atomic/trust.py
@@ -6,13 +6,12 @@ import json
 import yaml
 import requests
 import collections
+from base64 import b64encode
 
 TRANSPORT_TYPES = ["docker", "atomic", "dir"]
 
 def cli(subparser):
     # atomic trust
-    pubkeys_dir = util.get_atomic_config_item(['pubkeys_dir'])
-
     trustp = subparser.add_parser("trust",
                                   help=_("Manage system container trust policy"),
                                   epilog="Manages the trust policy of the host system. "
@@ -29,10 +28,16 @@ def cli(subparser):
                      atomic: openshift-based atomic registry"""
     commonp.add_argument("-k", "--pubkeys", nargs='?', default=[],
                          action="append", dest="pubkeys",
-                         help=_("Absolute path of installed public key(s) to trust for TARGET. "
+                         help=_("Local path or URL of public key(s) to trust for TARGET. "
+                                "Keys are parsed and encoded into policy.json. "
                                 "May used multiple times to define multiple public keys. "
-                                "File(s) must exist before using this command. "
-                                "Default directory is %s" % pubkeys_dir))
+                                "File(s) must exist before using this command."))
+    commonp.add_argument("-f", "--pubkeysfile", nargs='?', default=[],
+                         action="append", dest="pubkeysfile",
+                         help=_("Path of installed public key(s) to trust for TARGET. "
+                                "Absolute path to keys is added to policy.json. "
+                                "May used multiple times to define multiple public keys. "
+                                "File(s) must exist before using this command."))
     commonp.add_argument("--sigstoretype", dest="sigstoretype", default="web",
                          choices=['atomic', 'local', 'web'],
                          help=sigstore_help)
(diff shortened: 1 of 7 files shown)


docs/atomic-trust.1.md
```diff
@@ -52,6 +53,10 @@ testing.
   An absolute path to installed public key. May be used multiple times. Required
   for **signedBy** type.
 
+**-f**_**--file**
+  Reference pubkeys as files on host system instead of inlining pubkey data
```
rhatdan: Should this be public keys as files ...inlining public key data

Atomic/trust.py
```diff
@@ -71,6 +70,10 @@ def cli(subparser):
     showp.add_argument('--raw', action='store_true', help="Output raw policy file")
     showp.add_argument('-j', '--json', action='store_true', help="Output as json")
     showp.set_defaults(_class=Trust, func="show")
+    addp.add_argument('-f', '--file', action='store_true',
+                      help=_("Reference pubkeys as files on host system instead "
```
rhatdan: public keys?

Atomic/trust.py
```diff
@@ -83,7 +86,7 @@ def __init__(self, policy_filename="/etc/containers/policy.json"):
         self.policy_filename = os.environ.get('TRUST_POLICY', policy_filename)
         self.atomic_config = util.get_atomic_config()
 
-    def add(self, registry=None, pubkeys=None, sigstore=None, sigstoretype=None, keytype=None, trust_type=None):
+    def add(self, registry=None, pubkeys=None, sigstore=None, sigstoretype=None, keytype=None, trust_type=None, pubkeys_file=False):
```
rhatdan: Is pubkeys_file a boolean or a path?  Should we just check if path == None

Atomic/trust.py
```diff
@@ -92,6 +95,7 @@ def add(self, registry=None, pubkeys=None, sigstore=None, sigstoretype=None, key
         :param keytype: string, "GPGKeys"
         :param trust_type: string, one of "signedBy", "insecureAcceptAnything", "reject"
         :param sigstore: string, URL of signature server
+        :param pubkeys_file: boolean, reference pubkeys as filepath
```
rhatdan: If it is a boolean it should not be called a _file.

rhatdan:
You are still referencing pubkeys in the --help and man page for the user, I think we should spell it out for those cases.  Also the CLI now has --pubkeys and --file?  Is this confusing?

aweiteka:
> You are still referencing pubkeys in the --help and man page for the user, I think we should spell it out for those cases.

The option is pubkeys but I think I spell it out as "public keys" in the help/manpage. Do you want a longer option like `--public-keys`?

> Also the CLI now has --pubkeys and --file? Is this confusing?

Yes, possibly. How can we make this less confusing? I considered `--inline` but the user would need to use `--inline=False` to change default. Not sure.

rhatdan:
Lets discuss this together.

Atomic/trust.py
```diff
@@ -82,13 +87,20 @@ def __init__(self, policy_filename="/etc/containers/policy.json"):
         super(Trust, self).__init__()
         self.policy_filename = os.environ.get('TRUST_POLICY', policy_filename)
         self.atomic_config = util.get_atomic_config()
+        class Args():
+            def __init__(self):
+                self.debug = None
+                self.assumeyes = None
```
rhatdan: Why do I need this?

Atomic/trust.py
```diff
@@ -29,10 +28,16 @@ def cli(subparser):
                      atomic: openshift-based atomic registry"""
     commonp.add_argument("-k", "--pubkeys", nargs='?', default=[],
                          action="append", dest="pubkeys",
-                         help=_("Absolute path of installed public key(s) to trust for TARGET. "
+                         help=_("Local path or URL of public key(s) to trust for TARGET. "
+                                "Keys are parsed and encoded into policy.json. "
+                                "May used multiple times to define multiple public keys. "
+                                "File(s) must exist before using this command."))
+    commonp.add_argument("-f", "--pubkeysfile", nargs='?', default=[],
```
rhatdan: Shoudl --pubkeys and --pubkeysfile be mutually exlusive? Or would a user specify both on the same command line?

aweiteka: It's possible and valid to use both commands, although unexpected. No reason to restrict it, IMO.

Atomic/trust.py
aweiteka: This can be called from Atomic/pull.py (the auto-discover trust workflow), in which case it fails to pass in global opts. @baude we discussed this. Not sure if there is a more elegant workaround. It works but does seem silly.

rhatdan: Ok

Atomic/trust.py
rhatdan: ok

rhatdan:
Tests are failing on assumeyes?

aweiteka:
Fixed.

rhatdan:
@rh-atomic-bot r+

cgwalters:
@rh-atomic-bot r=rhatdan a296d6c

PR #854, opened 2017-01-30
Add ability to build/install on atomic-os

baude:
Making it easier for dev and test to be able to build and
install from git on an atomic os.

sudo make install-on-atomicos


baude:
@rh-atomic-bot r+ 15d334d

Issue #691, opened 2016-10-10
'atomic pull' should support pulling images by digest


miabbott:
In order to have more feature parity with `docker`, `atomic pull` should be able to retrieve images using the SHA256 ID.

See the following example using `docker`

```
# docker pull fedora@sha256:64a02df6aac27d1200c2572fe4b9949f1970d05f74d367ce4af994ba5dc3669e                                                                                      
Trying to pull repository registry.access.redhat.com/fedora ... 
unknown: Not Found
Trying to pull repository docker.io/library/fedora ... 
sha256:64a02df6aac27d1200c2572fe4b9949f1970d05f74d367ce4af994ba5dc3669e: Pulling from docker.io/library/fedora
2bf01635e2a0: Pull complete 
Digest: sha256:64a02df6aac27d1200c2572fe4b9949f1970d05f74d367ce4af994ba5dc3669e
Status: Downloaded newer image for docker.io/fedora@sha256:64a02df6aac27d1200c2572fe4b9949f1970d05f74d367ce4af994ba5dc3669e
```

baude:
@rhatdan do you know if the @ sign is unique to this kind of image description/pull?

I think we have the logic more or less, just need to figure out how to decompose this kind of input.

rhatdan:
@runcom PTAL

I have a feeling that this could also be used for other pulls.

runcom:
@baude yes, that @ sign is unique to the pull-by-digest in docker so you should be safe when decomposing a reference (either it has a tag with :tag or a digest with @sha256:hexvalue)

@rhatdan what do you mean by this could also be used for other pulls?

rhatdan:
@runcom I am not sure what I meant.  Ignore it.

baude:
Right now skopeo does not support this.  Will put this on hold until it does and then implement it into atomic.

rhatdan:
@runcom is there any movement on skopeo pulling by digest?

runcom:
Why is skopeo not supporting this already? what's missing here?
```
$ ./skopeo copy docker://busybox@sha256:29f5d56d12684887bdfa50dcd29fc31eea4aaf4ad3bec43daf19026a7ce69912 oci:test

Getting image source signatures
Copying blob sha256:56bec22e355981d8ba0878c6c2f23b21f422f30ab0aba188b54f1ffeff59c190
 603.20 KB / 652.49 KB [==================================================>----]
[...]

PR #856, opened 2017-01-31
Atomic/pull.py: Pull by Digest

baude:
Users wish to be able to pull images by digest.  This can be
done by docker and therefore parity is preferred.  The request
is described in https://github.com/projectatomic/atomic/issues/691.

```
$ atomic pull busybox@sha256:29f5d56d12684887bdfa50dcd29fc31eea4aaf4ad3bec43daf19026a7ce69912
Pulling docker.io/busybox@sha256:29f5d56d12684887bdfa50dcd29fc31eea4aaf4ad3bec43daf19026a7ce69912 ...
Copying blob sha256:56bec22e355981d8ba0878c6c2f23b21f422f30ab0aba188b54f1ffeff59c190
 495.23 KB / 652.49 KB [=========================================>-------------]
Copying config sha256:e02e811dd08fd49e7f6032625495118e63f597eb150403d02e3238af1df240ba
 0 B / 1.43 KB [---------------------------------------------------------------]
Writing manifest to image destination
Storing signatures
 1.43 KB / 1.43 KB [===========================================================]
$ docker images
REPOSITORY                         TAG                                                                IMAGE ID            CREATED             SIZE
docker.io/busybox                  29f5d56d12684887bdfa50dcd29fc31eea4aaf4ad3bec43daf19026a7ce69912   e02e811dd08f        3 months ago        1.093 MB
```
Note how the tag is populated with the digest.  Docker's default behaviour is
to populate the tag with nothing.  We cannot currently replicate that exactly.

### DIFF
--- a/Atomic/discovery.py
+++ b/Atomic/discovery.py
@@ -11,13 +11,14 @@ class RegistryAuthError(Exception):
 
 class RegistryInspect():
 
-    def __init__(self, registry=None, repo=None, image=None, tag=None, orig_input=None, debug=False):
+    def __init__(self, registry=None, repo=None, image=None, tag=None, digest=None, orig_input=None, debug=False):
         self.debug = debug
         self.registries = util.get_registries()
         self.registry = registry
         self.repo = repo
         self.image = image
         self.tag = tag
+        self.digest = digest
         self.orig_input = orig_input
         self._remote_inspect = None
         self.fqdn = None
@@ -52,7 +53,10 @@ class RegistryInspect():
         fqdn = fqdn if not self.repo else "{}/{}".format(fqdn, self.repo)
         fqdn += "/{}".format(self.image)
         if include_tag:
-            fqdn += ":{}".format(self.tag)
+            if self.tag:
+                fqdn += ":{}".format(self.tag)
+            elif self.digest:
+                fqdn += "@{}".format(self.digest)
         return fqdn
 
     def find_image_on_registry(self, quiet=False):
--- a/Atomic/objects/image.py
+++ b/Atomic/objects/image.py
@@ -27,6 +27,7 @@ class Image(object):
         self._virtual_size = None
         self.cmd = None
         self._command = None
+        self.repotags = None
 
 
         # Deeper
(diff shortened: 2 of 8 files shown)


baude:
@miabbott please try out at your convenience.

rhatdan:
Any chance you could add a test for this?

Atomic/discovery.py
```diff
@@ -52,7 +53,10 @@ def assemble_fqdn(self, include_tag=True, registry=None):
         fqdn = fqdn if not self.repo else "{}/{}".format(fqdn, self.repo)
         fqdn += "/{}".format(self.image)
         if include_tag:
-            fqdn += ":{}".format(self.tag)
+            if self.tag:
```
rhatdan: Should we raise an error if both tag and digest are defined at the same time?

Atomic/objects/image.py
```diff
@@ -109,21 +110,24 @@ def propagate(_img):
...
+            elif self.digest:
+                img += "/{}@{}".format(self.image, self.digest)
             self._fq_name = img
             propagate(self._fq_name)
             return img
 
         if not self.registry:
             ri = RegistryInspect(registry=self.registry, repo=self.repo, image=self.image,
-                                 tag=self.tag, orig_input=self.input_name)
+                                 tag=self.tag, digest=self.digest, orig_input=self.input_name)
```
rhatdan: Should tag and digest be combined like you did below? IE You pass either the tag or the digest to tagdigest field?

Atomic/discovery.py
baude: I think the right place for this would be in util.Decompose, I will add something there.

Atomic/objects/image.py
baude: At this point, if there is no objection, I'd like to keep them separate.

rhatdan: I have no problem, just a suggestion.

rhatdan:
@rh-atomic-bot r+

baude:
@rh-atomic-bot 2ce73e8

baude:
@rh-atomic-bot r+ 2ce73e8

Issue #797, opened 2016-12-12
atomic scan is not respecting fully-qualified image names


miabbott:
Seen on the following CAHC version:

```
# atomic host status
State: idle
Deployments:
● centos-atomic-continuous:centos-atomic-host/7/x86_64/devel/continuous
       Version: 7.2016.1192 (2016-12-12 15:37:20)
        Commit: 19cf5ddbfecf80c16602ea53c46e3cd49eae569292b2d38b76c1621d22854232
        OSName: centos-atomic-host
```

After pulling and installing the `registry.access.redhat.com/rhel7/openscap` image, the attempt to scan the image causes the `atomic` stack to go hunting for a version of the image on `docker.io`

```
# rpm -q atomic docker skopeo
atomic-1.13.8.50-8bfeae47e765f6ed775ddbdf8858d4819cd4179c.008b0012cb69d4dcb5b57cab83487e255a3feb67.el7.centos.x86_64
docker-1.10.3-46.el7.14.x86_64
skopeo-1.14-dd6441b5468f8f6d1b94cfe1d54cd628dbea914c.a8622550b28e564c396d659b08a84becf4a81e27.el7.centos.x86_64
# docker images -a
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
# atomic pull registry.access.redhat.com/rhel7/openscap
Image registry.access.redhat.com/rhel7/openscap is being pulled to docker ...
Pulling registry.access.redhat.com/rhel7/openscap:latest ...
Copying blob sha256:f9ab7164e76a8a165fce660638966566d3a4cb9dc8b63892f784acd4d1c7b9f3
 68.72 MB / ? [---------=-----------------------------------------------------] 
Copying blob sha256:d77ab047cab88f266864ed80af55249d37a06379b4bd1afef555996259bc819f
 0 B / ? [--------------------------------------------------------------------=]
Copying blob sha256:2835dc81b6c25e8e35cc217df7c04e5eb2586158c7b9b6a48e02311c14c29f9a
 30.54 MB / ? [---------------=-----------------------------------------------] 
Copying config sha256:e5b47a50e62783a7e7e02708239b1ba895d61db817bef1af3c3c95e88c199903
 0 B / 4.51 KB [---------------------------------------------------------------]
Writing manifest to image destination
Storing signatures
 4.51 KB / 4.51 KB [===========================================================]
# atomic --debug scan --scanner openscap registry.access.redhat.com/rhel7/openscap
[...]

PR #857, opened 2017-01-31
atomic.d/openscap: Change image name to be fully qualified

baude:
If the openscap file that defines the scanner image name uses the
short-name, i.e. rhel7/openscap, the on first use of atomic scan,
atomic will pull the scanning image.  However, atomic will name
the image with its fully qualified name, because it had to look it
up.  Therefore, in the local dockerd, the scanner will be named
with its full name.  The next time the scanner is run, it will
again attempt to pull down the short-named version.

We should just switch to the fq name to avoid this mess. This issue
was reported in https://github.com/projectatomic/atomic/issues/797.

### DIFF
--- a/atomic.d/openscap
+++ b/atomic.d/openscap
@@ -1,6 +1,6 @@
 type: scanner
 scanner_name: openscap
-image_name: rhel7/openscap
+image_name: registry.access.redhat.com/rhel7/openscap
 default_scan: cve
 custom_args: ['-v', '/etc/oscapd:/etc/oscapd:ro']
 scans: [ 


baude:
@miabbott please check this out too.

miabbott:
This worked OK for me in the case where I have no scanner pulled to the host already.

Interestingly, I found that if I `atomic install registry.access.redhat.com/rhel7/openscap` before running the scan, it drops a version of `/etc/atomic.d/openscap` that doesn't use the FQDN.  I'll file a bug downstream for that.

miabbott:
@baude Instead of specifying the RHEL version of openscap, maybe you should use this instead `docker.io/fedora/atomic_scan_openscap` ?

rhatdan:
Yes we should replace this file and have the distributions modify it in their rpms.  By default lets go with a fedora scanner, although I would prefer to pull it directly from the fedora repo.

baude:
No, right now we only ship this file in the RHEL package.  This gives them a starting point.  For Fedora and other Os's, the just do atomic install distro/openscap and it should lay that file in.

rhatdan:
@rh-atomic-bot r+

baude:
@rh-atomic-bot r+ fc069ad

miabbott:
This change worked out better than expected since the `docker.io/fedora/atomic_scan_openscap` image no longer exists.

Issue #832, opened 2017-01-17
Throw better error for unknown image


baude:
The user gets different errors when trying to pull an image that does not exist and choosing the different storage backend

```
# ./atomic pull --storage ostree foobar
ReturnTuple(return_code=1, stdout='', stderr='time="2017-01-17T19:59:34Z" level=fatal msg="unauthorized: authentication required" \n')
# ./atomic pull foobar
Unable to resolve foobar
```

baude:
https://github.com/projectatomic/atomic/pull/858

baude:
fixed

Issue #636, opened 2016-09-20
Need to handle pull 404 image


aweiteka:
Pulling an image that doesn't exist returns...

```
$ sudo atomic --debug pull docker.io/aweiteka/hello-openshift:latest
Image docker.io/aweiteka/hello-openshift:latest is being pulled to docker ...
ReturnTuple(return_code=1, stdout='', stderr='time="2016-09-20T08:38:22-04:00" level=fatal msg="error fetching manifest: status code: 404, body: {\\"errors\\":[{\\"code\\":\\"MANIFEST_UNKNOWN\\",\\"message\\":\\"manifest unknown\\",\\"detail\\":{\\"Tag\\":\\"latest\\"}}]}\\n" \n')
Traceback (most recent call last):
  File "/bin/atomic", line 186, in <module>
    sys.exit(_func())
  File "/usr/lib/python2.7/site-packages/Atomic/pull.py", line 54, in pull_image
    handler()
  File "/usr/lib/python2.7/site-packages/Atomic/pull.py", line 37, in pull_docker_image
    fq_name = skopeo_inspect("{}{}".format(pull_uri, self.args.image))['Name']
  File "/usr/lib/python2.7/site-packages/Atomic/util.py", line 274, in skopeo_inspect
    raise ValueError(results)
ValueError: ReturnTuple(return_code=1, stdout='', stderr='time="2016-09-20T08:38:22-04:00" level=fatal msg="error fetching manifest: status code: 404, body: {\\"errors\\":[{\\"code\\":\\"MANIFEST_UNKNOWN\\",\\"message\\":\\"manifest unknown\\",\\"detail\\":{\\"Tag\\":\\"latest\\"}}]}\\n" \n')
```

Version: 1.12.3

rhatdan:
I began playing with this, but I am failing to get even to the point of this error.

baude:
IMHO, we need to work in return codes to skopeo so we can catch this kind of stuff.

rhatdan:
I think skopeo is returning lots of data in the form of a json data. and we need to parse this to get some kind of better error message for the user.

@runcom is this the intention?

runcom:
> @runcom is this the intention?

yes, I'm having the same _issue_ when integrating simple signing into projectatomic/docker but I'm definitely just going to parse skopeo's output for now - we could work on error codes though.

> I began playing with this, but I am failing to get even to the point of this error.

This should trigger that error I guess:
```
[...]

PR #858, opened 2017-01-31
Unify 404/unable to find image messages

baude:
When looking for an image on a docker registry, atomic pull should
display similar error messages when unable to find an image
regardless of the storage backend.  We now currently do this:

$ sudo atomic pull foobar123
Unable to find foobar123
$ sudo atomic pull --storage ostree foobar123
Unable to find foobar123

This should satify https://github.com/projectatomic/atomic/issues/832 and
https://github.com/projectatomic/atomic/issues/636.

### DIFF
--- a/Atomic/discovery.py
+++ b/Atomic/discovery.py
@@ -78,5 +78,5 @@ class RegistryInspect():
                 if not quiet:
                     util.write_err("Failed: {}".format(e))
                 continue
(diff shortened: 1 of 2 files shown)


giuseppe:
@rh-atomic-bot r+ 7fe6fe9

baude:
@rh-atomic-bot r+ 7fe6fe9

PR #861, opened 2017-02-02
Detect and install systemcontainer images

baude:
If an image from a registry has a label like:

    atomic.type=system

the Atomic CLI will now detect that prior to pulling
the image.  It will then switch the backend storage
to ostree.  In the case of install, it will both
pull to ostree and install it.

If the backend is switched, a message informing the user
as such will be shown.

The detection and switch of the backend can be overriden
 with the --storage option.

The following is an example of install:
```
$ sudo atomic install atomic-registry.usersys.redhat.com:5000/brentbaude/helloworld:latest

Note: Switching from the 'docker' backend to the 'ostree' backend based on the 'atomic.type' label in the image.  You can use --storage to override this behaviour.

Extracting to /var/lib/containers/atomic/helloworld.0
systemctl daemon-reload
systemctl enable helloworld

$ sudo ./atomic images list
   REPOSITORY                                                      TAG      IMAGE ID       CREATED            VIRTUAL SIZE   TYPE
   atomic-registry.usersys.redhat.com:5000/brentbaude/helloworld   latest   0056366e4ebd   2017-02-01 18:04                  ostree

$ sudo ./atomic containers list -a
   CONTAINER ID IMAGE                COMMAND              CREATED          STATE     BACKEND    RUNTIME
   helloworld   atomic-registry.user /usr/bin/run.sh      2017-02-01 18:04 failed    ostree     runc
```

### DIFF
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -236,18 +236,20 @@ class DockerBackend(Backend):
     def stop_container(self, con_obj):
         return self.d.stop(con_obj.id)
 
-    def pull_image(self, image, **kwargs):
+
+    def pull_image(self, image, remote_image_obj, **kwargs):
+        assert(isinstance(remote_image_obj, Image))
         debug = kwargs.get('debug', False)
         if image.startswith("dockertar:"):
             path = image.replace("dockertar:", "", 1)
             with open(path, 'rb') as f:
                 self.d.load_image(data=f)
             return 0
-        remote_image = self.make_remote_image(image)
-        fq_name = remote_image.fq_name
+#        remote_image_obj = self.make_remote_image(image)
+        fq_name = remote_image_obj.fq_name
         local_image = self.has_image(image)
         if local_image is not None:
-            if self.already_has_image(local_image, remote_image):
+            if self.already_has_image(local_image, remote_image_obj):
                 raise ValueError("Latest version of {} already present.".format(image))
         registry, _, _, tag, _ = util.Decompose(fq_name).all
         image = "docker-daemon:{}".format(image)
@@ -295,9 +297,10 @@ class DockerBackend(Backend):
 
     def update(self, name, force=False, **kwargs):
         debug = kwargs.get('debug', False)
+        remote_image_obj = self.make_remote_image(name)
         try:
             # pull_image will raise a ValueError if the "latest" image is already present
-            self.pull_image(name, debug=debug)
+            self.pull_image(name, remote_image_obj, debug=debug)
         except ValueError:
             return
         # Only delete containers if a new image is actually pulled.
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -245,7 +245,6 @@ class DockerBackend(Backend):
             with open(path, 'rb') as f:
                 self.d.load_image(data=f)
             return 0
-#        remote_image_obj = self.make_remote_image(image)
         fq_name = remote_image_obj.fq_name
         local_image = self.has_image(image)
         if local_image is not None:
--- a/Atomic/backendutils.py
+++ b/Atomic/backendutils.py
@@ -159,3 +159,4 @@ class BackendUtils(object):
     def message_backend_change(previous, new):
         write_out("\nNote: Switching from the '{}' backend to the '{}' backend based on the 'atomic.type' label in the "
                   "image.  You can use --storage to override this behaviour.\n".format(previous, new))
+
--- a/Atomic/backendutils.py
+++ b/Atomic/backendutils.py
@@ -155,4 +155,7 @@ class BackendUtils(object):
     def get_container_obj_by_image_name(self, image_name, str_preferred_backend):
         pass
 
-
+    @staticmethod
+    def message_backend_change(previous, new):
+        write_out("\nNote: Switching from the '{}' backend to the '{}' backend based on the 'atomic.type' label in the "
+                  "image.  You can use --storage to override this behaviour.\n".format(previous, new))
--- a/Atomic/install.py
+++ b/Atomic/install.py
@@ -78,7 +78,6 @@ class Install(Atomic):
         super(Install, self).__init__()
 
     def install(self):
-        debug = self.args.debug
         if self.args.debug:
             util.write_out(str(self.args))
         be_utils = BackendUtils()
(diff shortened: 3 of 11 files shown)


baude:
@rhatdan and @giuseppe  PTAL

One issue is that run and install do not have a --storage option like pull.  So as of right now, this can not be overridden.  I'm thinking we should add --storage but in the case of install, would that conflict with --system?

Feel free to provide feedback or we can discuss tomorrow as well.

giuseppe:
what if install FOO, first looks if there is an image FOO in the Docker storage and if it is not found, tries to find it in other storages (like ostree)?
If we are going to add more storages, we will probably need something like `--storage=` for install and run (we can probably leave `--system` to be an alias of `--storage=ostree` , so that the command would fail if the image is not of the specified type.  If the type is not specified, we can use the first image we find in a storage.

What do you think?
What do you think?

rhatdan:
I think we should add --storage to pull and install so this can be overridden.

--system means more then just --storage=ostree, it also means to grap the templates out of /export, and put them on the host.

I think --system with a --storage!=ostree should be denied at this point.  Perhaps with a warning.  So that you can install a system container to be used for running orchestrated containers.

baude:
@rhatdan good to go

Atomic/backends/_docker.py
```diff
@@ -236,18 +236,20 @@ def start_container(self, name):
...
+        assert(isinstance(remote_image_obj, Image))
         debug = kwargs.get('debug', False)
         if image.startswith("dockertar:"):
             path = image.replace("dockertar:", "", 1)
             with open(path, 'rb') as f:
                 self.d.load_image(data=f)
             return 0
-        remote_image = self.make_remote_image(image)
-        fq_name = remote_image.fq_name
+#        remote_image_obj = self.make_remote_image(image)
```
rhatdan: Remove

Atomic/backendutils.py
```diff
@@ -155,4 +155,7 @@ def get_containers(self):
     def get_container_obj_by_image_name(self, image_name, str_preferred_backend):
         pass
 
-
+    @staticmethod
+    def message_backend_change(previous, new):
+        write_out("\nNote: Switching from the '{}' backend to the '{}' backend based on the 'atomic.type' label in the "
+                  "image.  You can use --storage to override this behaviour.\n".format(previous, new))
```
rhatdan: Add Return.

Atomic/install.py
```diff
@@ -40,6 +43,10 @@ def cli(subparser):
         "%s'" % Install.print_install())
     installp.set_defaults(_class=Install, func='install')
     add_opt(installp)
+    installp.add_argument("--storage", dest="storage", default=None,
+                   help=_("Specify the storage. Default is currently '%s'.  You can"
+                          " change the default by editing /etc/atomic.conf and changing"
+                          " the 'default_storage' field." % _storage))
```
rhatdan: Should we list out the valid storage.
docker, ostree

Atomic/install.py
```diff
@@ -78,9 +85,16 @@ def __init__(self):
         super(Install, self).__init__()
 
     def install(self):
-        debug = self.args.debug
         if self.args.debug:
             util.write_out(str(self.args))
+        storage_set = False if self.args.storage is None else True
+        storage = _storage if not storage_set else self.args.storage
```
rhatdan: Where is _storage defined

Atomic/install.py
```diff
@@ -78,9 +85,16 @@ def __init__(self):
...
+        if storage == 'ostree':
+            self.args.system = True
```
rhatdan: I don't think just because you are using ostree, we should set self.args.system.  You could be able to install an ostree container without running through the system stuff.

rhatdan:
Need man page updates and need bash completions.

rhatdan:
@giuseppe ptal

Atomic/install.py
baude: Right here -> https://github.com/baude/atomic/blob/cf0683c91b6be8ed0a76ec0f8739509670c15fc7/Atomic/install.py#L33

Atomic/install.py
baude: @giuseppe we discussed this yesterday and I think we agreed to what I coded up.  What do you think about dan's comment?

Atomic/install.py
baude: I'd prefer not to right now.  We don't have a great way to do it.  I can add later if you like.

Atomic/backendutils.py
baude: This isn't fatal.  Why does it need a return ?

Atomic/backends/_docker.py
baude: Done

Atomic/backendutils.py
baude: maybe you mean just a newline ?

baude: if so ... done

baude:
Added bash completions and man updates

Atomic/install.py
giuseppe: at the moment we have no other way of using ostree if not using all the system stuff, but we probably avoid this for now (storage == ostree implies --system) until we have a clearer idea.

Also, storage==ostree could imply --user if running as non root. if you are fine with it, let's skip this for now and expect an explicit --system for system containers

Atomic/install.py
rhatdan: Ok.  Just looking to enhance the User Experience.

Atomic/install.py
rhatdan: Yes I saw it was global after I asked.

Atomic/install.py
rhatdan: Yes keep them separate.  We might in the future not imply --system means --storage=ostree also. Think about network file system images.

docs/atomic-install.1.md
```diff
@@ -73,6 +74,12 @@ Set a value that is going to be used by a system container for its
 configuration and can be specified multiple times.  It is used only
 by --system.  OSTree is required for this feature to be available.
 
+**--storage**
```
rhatdan: What are the valid options?

docs/atomic-run.1.md
```diff
@@ -70,6 +71,11 @@ NAME will default to the IMAGENAME if it is not specified.
 
 `/usr/bin/docker run -t -i --rm --privileged -v /:/host -v /run:/run --net=host --ipc=host --pid=host -e HOST=/host -e NAME=${NAME} -e IMAGE=${IMAGE} --name ${NAME} ${IMAGE}`
 
+**-storage**
```
rhatdan: What are the valid options?

bash/atomic
```diff
@@ -688,6 +689,7 @@ _atomic_install() {
 		--help
 	       --display
 	       --rootfs
+	       --storage
```
rhatdan: It would be nice to have command completion on the valid options.

baude: how do I do that?

Atomic/install.py
baude: Fixed

docs/atomic-install.1.md
baude: Done

docs/atomic-run.1.md
baude: Added

bash/atomic
baude: done

baude:
ready

cgwalters:
Nice work on this!

baude:
@cgwalters thanks, really appreciate that.

rhatdan:
@rh-atomic-bot r+

baude:
@rh-atomic-bot r+ 001636b

PR #863, opened 2017-02-03
Atomic/backends/_docker.py: Modify check_args to check for '--security-opt label=disable'

dharmit:
`docker-run` man page on Fedora 25 and CentOS 7 states `label=disable`
as one of the security options. However, we check for `label:disable`
only. This commit should help us check for both

Example:

~~~python
In [40]: r.get_label("RUN")
Out[40]: u'docker run -it --privileged --security-opt label=disable registry.centos.org/centos/centos bash'

In [41]: r.check_args(r.get_label("RUN"))

INFO: --privileged

In [42]: r.get_label("RUN")
Out[42]: u'docker run -it --privileged --security-opt label:disable registry.centos.org/centos/centos bash'

In [43]: r.check_args(r.get_label("RUN"))

INFO: --security-opt label:disable
      Disabling label separation turns off tools like SELinux and could allow processes from the container to break out onto your host system.

INFO: --privileged

~~~


dharmit:
Okay, so creating a tuple `('--security-opt label:disable', '--security-opt label=disable')` obviously didn't go well at [line #427](https://github.com/dharmit/atomic/blob/4b7e9fc8ea1e767a0aaa2ba694e9dd95a88f6537/Atomic/backends/_docker.py#L427). I think the only way to cover both `label:disable` and `label=disable` here would be by having separate keys for them while keeping the same value for both keys.

I don't see another way.

rhatdan:
Since the label:disable is going away or gone, I think we should just change to the label=disable.

rhatdan:
@rh-atomic-bot r+

dharmit:
> Since the label:disable is going away or gone, I think we should just change to the label=disable.

@rhatdan should I remove `label:disable` in that case?

rhatdan:
@rh-atomic-bot r+ 7208284

PR #864, opened 2017-02-03
Fix Delete errors with images and containers

baude:
Fix a slew of issues related to deletion of images and containers.  Also
added --assume_yes to be consistent with other commands.  This will benefit
dbus interaction as well given that we do not want confirmation occuring
over dbus.

Issues that were fixed:
 * Proper catch and error for trying to delete all images|containers when no images|containers exist
 * Catch and error when no image|container is given by user and no --all
 * when --storage and --all are used, only images|containers in that backend are now deleted
 * when --all is given but not --storage, all images|containers from all backends are now deleted.
 * Fixed logical error when deleting containers
 * Deletion of images|containers now is consistent in its confirm and abort messages.

Added unittests.

### DIFF
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -1,6 +1,6 @@
 import os
 import copy
-
+import sys
 from . import util
 from . import Atomic
 from .client import AtomicDocker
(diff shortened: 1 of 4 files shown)


Atomic/containers.py
```diff
@@ -33,6 +33,8 @@ def cli(subparser):
     delete_parser.add_argument("-a", "--all", action='store_true',dest="all",
                                default=False,
                                help=_("Delete all containers"))
+    delete_parser.add_argument("--assume_yes", action='store_true',dest="assume_yes",
```
rhatdan: assumeyes is the way this is defined elsewhere.

rhatdan: assumeyes is defined at the atomic level, so you should just use that.

Atomic/images.py
```diff
@@ -55,6 +55,8 @@ def cli(subparser):
                                       "different storages, you will be prompted to specify."))
     delete_parser.add_argument("-a", "--all", action='store_true',dest="all",
                                default=False, help=_("Delete all images"))
+    delete_parser.add_argument("--assume_yes", action='store_true',dest="assume_yes",
```
rhatdan: remove

bash/atomic
```diff
@@ -787,6 +787,7 @@ _atomic_images_delete() {
     --storage
   "
   local all_options="$options_with_args
+    --assume_yes
```
rhatdan: remove

rhatdan:
@rh-atomic-bot r+ fa9433a

rhatdan:
@rh-atomic-bot r+

PR #867, opened 2017-02-03
Add git to atomic-tester

baude:
With the addition of git to the atomic-test image we can
script pulling and building the atomic cli on an
atomic host which lacks these tools.  It appears to add
about 12MB to the image.


baude:
@jlebon ptal and consider merging.

cgwalters:
@rh-atomic-bot r+ 3a543a4

Note it's a docker hub build, I don't know what the latency on that is.

PR #876, opened 2017-02-07
Trust show: parse inline public keys

aweiteka:
Follow-on to #853

### DIFF
--- a/Atomic/trust.py
+++ b/Atomic/trust.py
@@ -6,7 +6,8 @@ import json
 import yaml
 import requests
 import collections
-from base64 import b64encode
+from base64 import b64encode,b64decode
+import tempfile
 
 TRANSPORT_TYPES = ["docker", "atomic", "dir"]
 
@@ -486,17 +487,20 @@ class Trust(Atomic):
         """
         if not keys:
             return ""
-        keylist = None
+        (keylist, tmpkey) = None, None
         for key in keys:
             if not os.path.exists(key):
-                # extend when we support parsing inline 'keyData' base64 encoded key
-                return ""
-            else:
-                cmd = ["gpg2", "--with-colons", key]
+                with tempfile.NamedTemporaryFile(delete=False) as tmpkey:
+                    tmpkey.write(b64decode(key))
+                key = tmpkey.name
+            cmd = ["gpg2", "--with-colons", key]
             try:
                 results = util.check_output(cmd).decode('utf-8')
             except util.FileNotFound:
                 results = ""
+            if tmpkey:
+                if os.path.exists(tmpkey.name):
+                    os.remove(tmpkey.name)
             lines = results.split('\n')
             for line in lines:
                 if line.startswith("uid:") or line.startswith("pub:"):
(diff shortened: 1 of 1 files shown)


aweiteka:
@rh-atomic-bot retry

aweiteka:
bot, retest this please

Atomic/trust.py
```diff
@@ -478,10 +479,10 @@ def get_gpg_id(self, keys):
         keylist = None
         for key in keys:
             if not os.path.exists(key):
-                # extend when we support parsing inline 'keyData' base64 encoded key
-                return ""
-            else:
-                cmd = ["gpg2", "--with-colons", key]
+                with tempfile.NamedTemporaryFile(delete=False) as tmpkey:
```
rhatdan: I think you need to make sure that this file is created on /run, not somewhere a user might be able to attack

Atomic/trust.py
```diff
@@ -478,12 +479,13 @@ def get_gpg_id(self, keys):
...
-                return ""
-            else:
-                cmd = ["gpg2", "--with-colons", key]
+                with tempfile.TemporaryFile() as tmpkey:
+                    tmpkey.write(b64decode(key))
+                key = tmpkey.name
+            cmd = ["gpg2", "--with-colons", key]
             try:
                 results = util.check_output(cmd).decode('utf-8')
+                tmpkey.close()
```
aweiteka: @rhatdan I tightened the tempfile up a bit. I'm not sure what you mean by using /run. Prefix dir for tempfile?

rhatdan: yes

Atomic/trust.py
```diff
@@ -478,10 +479,10 @@ def get_gpg_id(self, keys):
...
         for key in keys:
             if not os.path.exists(key):
-                # extend when we support parsing inline 'keyData' base64 encoded key
-                return ""
-            else:
-                cmd = ["gpg2", "--with-colons", key]
+                with tempfile.TemporaryFile() as tmpkey:
+                    tmpkey.write(b64decode(key))
+                key = tmpkey.name
+            cmd = ["gpg2", "--with-colons", key]
```
rhatdan: Should we be removing the tmpfile when this completes?

Atomic/trust.py
```diff
@@ -486,17 +487,20 @@ def get_gpg_id(self, keys):
...
             return ""
-        keylist = None
+        (keylist, tmpkey) = None, None
         for key in keys:
             if not os.path.exists(key):
-                # extend when we support parsing inline 'keyData' base64 encoded key
-                return ""
-            else:
-                cmd = ["gpg2", "--with-colons", key]
+                with tempfile.NamedTemporaryFile(delete=False, dir="/run/") as tmpkey:
```
rhatdan: Add delete=True and then when the file is closed it should get deleted.  Then you should not need to do code below.  I think?

aweiteka: `delete=True` (the default) deletes the file when the with-open loop exists, so it's gone when we need it failing the gpg2 cmd.

rhatdan:
Can you merge the pull requests, and give me a test?

rhatdan:
@rh-atomic-bot r+ 0edc5f7

PR #877, opened 2017-02-07
Run args should include /bin/sh

rhatdan:


rhatdan:
@baude I think this fixes the issue.

atomic run fedora

Is supposed to work.

rhatdan:
https://bodhi.fedoraproject.org/updates/FEDORA-2017-b677c4a509

jlebon:
Is that what we want though? Shouldn't we just let docker pick up the default CMD?

rhatdan:
This change is wrong, problem is deeper in the code.

rhatdan:
jlebon, there is no default command, but maybe you are right.  Perhaps we should look for command or entrypoint inside of the image and then fall over to /bin/sh

rhatdan:
@jlebon ^^ @baude ^^

Bottom line is we need to get back to what we had before.

PR #880, opened 2017-02-07
Fix run behavior

baude:
There were bugs in the run behavior in several scenarios:

* running an image with a command
* running a container that is stopped with and without a command
* running a container that is running with and without a command

### DIFF
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -349,6 +349,12 @@ class DockerBackend(Backend):
         return layers
 
     def run(self, iobject, **kwargs):
+        def add_string_or_list_to_list(list_item, value):
+            if not isinstance(value, list):
+                value = value.split()
+            list_item += value
+            return list_item
+
         atomic = kwargs.get('atomic', None)
         args = kwargs.get('args')
         # atomic must be an instance of Atomic
@@ -361,7 +367,7 @@ class DockerBackend(Backend):
         # If container exists and is started, execute command inside it (docker exec)
         # If container doesn't exist, create one and start it
         if args.command:
-            iobject.command = args.command
+            iobject.user_command = args.command
         if isinstance(iobject, Container):
             if iobject.running:
                 return self._running(iobject, args, atomic)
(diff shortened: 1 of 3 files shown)


Atomic/objects/container.py
```diff
@@ -95,3 +96,6 @@ def name(self):
     def name(self, value):
         self._name = value[1:] if value[0] == '/' else value
 
+    @property
+    def user_command(self):
+        return self._user_command
```
rhatdan: Need new line

Atomic/backends/_docker.py
```diff
@@ -451,13 +464,20 @@ def check_args(cmd):
...
                 return atomic.display(cmd)
             else:
                 return util.check_call(cmd, stderr=DEVNULL)
         else:
-            cmd = [atomic.docker_binary(), "exec", "-t", "-i", con_obj.name] + con_obj.command
+            command = con_obj.command if not args.command else args.command
+            try:
+                cmd = [atomic.docker_binary(), "exec", "-t", "-i", con_obj.name] + command
+            except TypeError:
+                cmd = [atomic.docker_binary(), "exec", "-t", "-i", con_obj.name] + command
```
rhatdan: This is the same line as before?  If it throws the first TypeError it will throw the second

Atomic/backends/_docker.py
```diff
@@ -361,16 +367,19 @@ def run(self, iobject, **kwargs):
         # If container exists and is started, execute command inside it (docker exec)
         # If container doesn't exist, create one and start it
         if args.command:
-            iobject.command = args.command
+            iobject._user_command = args.command #pylint: disable=protected-access
```
rhatdan: Why not add an interface to set this?

rhatdan:
Display test is failing.

baude:
the test is failing because the test is bad.  It will generate different results based on tty or not which the bots dont have.  working on a proposal

Atomic/objects/container.py
baude: Fixed

rhatdan:
Lets change the test to ignore the -t flag.  

COMMAND | sed 's/ -t / /g'

Atomic/backends/_docker.py
baude: Yeah, good idea

baude:
I was trying a test to detect interactive or not ... prefer just removing the -t ?

baude:
@rhatdan ok, reverted, lets keep your new tests out of this. this is my fixes plus your review comments.

baude:
bot, retest this please

baude:
@rh-atomic-bot bot, retest this please

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot r=rhatdan 5007fb3

rhatdan:
@rh-atomic-bot r+ 5007fb3

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot retry

PR #883, opened 2017-02-09
syscontainers: support update containers with --rebase

giuseppe:
A couple of features I'd like to have for the OpenShift installer:

1) support `containers update --to`, let's not require to uninstall/install the container if updating to a different image/tag, it will enable to do things like:
```
atomic install --system --name origin origin:v1.3.4
atomic containers update origin --to origin:v1.3.5
```

2) make `containers update` smarter and not require an update if neither the image revision and the configuration was changed.  Before we were checking only for the image revision or if --set was used, not caring if it was used to set to the same configuration already in place.

### DIFF
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -73,6 +73,8 @@ def cli(subparser):
     updatep.set_defaults(_class=Containers, func='update')
     updatep.add_argument("container",
                          help=_("Specify one or more containers. Must be final arguments."))
+    updatep.add_argument("--rebase", dest="rebase", default=None,
+                         help=_("Update to a different image (useful for upgrading to a different tag)"))
     if OSTREE_PRESENT:
         updatep.add_argument("--set", dest="setvalues",
                              action='append',
@@ -263,9 +265,9 @@ class Containers(Atomic):
             if con.id in vulnerable_uuids:
                 con.vulnerable = True
 
-    def update(self, to=None):
+    def update(self):
         if self.syscontainers.get_checkout(self.args.container):
-            return self.syscontainers.update_container(self.args.container, self.args.setvalues)
+            return self.syscontainers.update_container(self.args.container, self.args.setvalues, self.args.rebase)
         raise ValueError("System container '%s' is not installed" % self.args.container)
 
     def rollback(self):
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -263,9 +263,9 @@ class Containers(Atomic):
             if con.id in vulnerable_uuids:
                 con.vulnerable = True
 
-    def update(self):
+    def update(self, to=None):
         if self.syscontainers.get_checkout(self.args.container):
-            return self.syscontainers.update_container(self.args.container)
+            return self.syscontainers.update_container(self.args.container, self.args.setvalues)
         raise ValueError("System container '%s' is not installed" % self.args.container)
 
     def rollback(self):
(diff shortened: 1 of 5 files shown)


Atomic/containers.py
```diff
@@ -73,6 +73,8 @@ def cli(subparser):
     updatep.set_defaults(_class=Containers, func='update')
     updatep.add_argument("container",
                          help=_("Specify one or more containers. Must be final arguments."))
+    updatep.add_argument("--to", dest="to", default=None,
+                         help=_("Update to a different image (useful to upgrade to a different tag)"))
```
yuqi-zhang: Minor nitpick, "useful for upgrading to a different tag" sounds better

bash/atomic
```diff
@@ -578,7 +578,7 @@ _atomic_containers_trim() {
 _atomic_containers_update() {
 	case "$cur" in
 		-*)
-			COMPREPLY=( $( compgen -W "--help --set" -- "$cur" ) )
+			COMPREPLY=( $( compgen -W "--help --set --to" -- "$cur" ) )
```
yuqi-zhang: I wonder if we should ever make these smarter, since we don't display the options for "set" and "to" when you try to continue auto complete (unlike, say, the --storage flag for pull). But that would be somewhat difficult.

yuqi-zhang:
Otherwise LGTM

giuseppe:
@yuqi-zhang pushed a fixup patch.

good idea, we could show the variables, that is doable.  It'll be a bit slow though, as we will need to inspect the image every time <tab> is pressed

Atomic/containers.py
giuseppe: I think --tag is not complete, as we can update to another image in a different registry, like:
```
atomic install --name etcd gscrivano/etcd
atomic containers update etcd --to rhel/etcd
```

Should we call it `--rebase`?

rhatdan:
I like --rebase is better the -to, at least we have a clue what that means.

giuseppe:
pushed a new version with `--rebase` instead of `--to`

Atomic/containers.py
```diff
@@ -73,6 +73,8 @@ def cli(subparser):
     updatep.set_defaults(_class=Containers, func='update')
     updatep.add_argument("container",
                          help=_("Specify one or more containers. Must be final arguments."))
+    updatep.add_argument("--rebase", dest="rebase", default=None,
+                         help=_("Update to a different image (useful for upgrading to a different tag)"))
```
rhatdan: Rebase to a different image ...

docs/atomic-containers.1.md
```diff
@@ -87,6 +87,9 @@ Can use --set to update environment variables.
   Only display container IDs
 
 # update OPTIONS:
+**--rebase=IMAGE**
+  Override the image to update to.  if not specified, the same image used to install the container will be used.
```
rhatdan: Rebase to a different image.

rhatdan:
Bash Completions?

giuseppe:
I pushed a fixup patch to address the two comments.  The bash completion was already added as part of the same commit "update: support update to a different image via --rebase"

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot r=rhatdan 08a5567

giuseppe:
@rh-atomic-bot retry

PR #884, opened 2017-02-10
Suppress STDOUT of lower level commands e.g. vgreduce, wipefs.

shishir-a412ed:

### DIFF
--- a/Atomic/storage.py
+++ b/Atomic/storage.py
@@ -209,30 +209,36 @@ class Storage(Atomic):
         n_pvs = len(pvs)
         devices = set(devices)
         parents = None
-        for pv in pvs:
-            parents = list_parents(pv)
-            if set(parents).isdisjoint(devices):
-                continue
-            devices -= set(parents)
-
-            if query_pvs(pv, "pv_used")[0][:-1] != '0':
-                if only_unused:
+        try:
+            for pv in pvs:
+                parents = list_parents(pv)
+                if set(parents).isdisjoint(devices):
                     continue
-                else:
-                    util.check_call([ "pvmove", pv ])
+                devices -= set(parents)
 
-            if n_pvs > 1:
-                util.check_call([ "vgreduce", vgroup, pv ])
-            elif len(lvs) == 0:
-                util.check_call([ "vgremove", vgroup ])
-            n_pvs -= 1
-            util.check_call([ "wipefs", "-a", pv ])
-            util.sh_modify_var_in_file(self.dss_conf, "DEVS",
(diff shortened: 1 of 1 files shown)


shishir-a412ed:
Right now when I do: `atomic storage modify --remove-device /dev/vdb`. I see this output.
 
````
Removed "/dev/vdb1" from volume group "fedora"
/dev/vdb: 2 bytes were erased at offset 0x000001fe (dos): 55 aa
/dev/vdb: calling ioctl to re-read partition table: Success
````

This patch suppresses the lower level commands output.

ping @rhatdan

sghosh151:
Log it with meta information about the action rather than suppressing it and missing occasionally important stuff

rhatdan:
I kind of like that output. Why do you want to remove it?

shishir-a412ed:
@rhatdan My thought was If I am running 
````
$$ atomic storage modify --remove-device /dev/vdb. 
````
I would just like to see if my device `/dev/vdb` was removed successfully or not. On a successful run. OUTPUT: `/dev/vdb removed successfully` should be enough for a user.

Ofcourse if there is an error, more information is helpful for debugging.

I just felt for a user (On a successful run of the command): 
a) If device signatures were wiped out properly or not.
b) If device partition table was updated and re-read by ioctl 
    would be too much information since the command was successful anways.

Let me know if you feel differently and I can close this PR.

Shishir

baude:
@shishir-a412ed How would you feel about having that information be put into a bytestring where on failure, it would be spit out ?

shishir-a412ed:
@baude PTAL.

Now on success, there will be no output message. {Same as `atomic storage modify --add-device /dev/vdb`}. When `--add-device` is successful, there is no output message, just a return code 0. This patch would exhibit same behavior for `--remove-device`.

In case of an error it would look something like this: (For testing purposes I changed the volume group name to `atomicos` {atomicos does not exist, my VG name is fedora}  to generate an error)

````
sh-4.4# atomic storage modify --remove-device /dev/vdb 
Return Code: 5
Failure: b'  Volume group "atomicos" not found\n  Cannot process volume group atomicos\n  Volume group "atomicos" not found\n  Cannot process volume group atomicos\n  Failed to find physical volume "/dev/vdb1".\n'
atomic storage modify --remove-device failed
````

shishir-a412ed:
@baude Any updates on this one ?

baude:
LGTM but @rhatdan  has the final call.

shishir-a412ed:
@baude :) thanks ! .... Will wait for his input.

rhatdan:
@rh-atomic-bot r+ 6be9adf

PR #885, opened 2017-02-11
atomic_dbus.py: Allow for async scans via dbus

baude:
Users like cockpit need to be able to submit scan jobs and not
tie up the dbus bus while waiting for results. They also want
to be able to check the status of active scans.

We now have an AsyncScan method that takes an image|container
id in string format as input.  It will then thread the scan
and return.  The scan can then be monitored through dbus
signals.

Results can be obtained using the GetScanResultsByID method.

There is also an ActiveScans method that allows you to check
the active scans.  It returns a list of container|image ids
that are being scanned.

### DIFF
--- a/atomic_dbus.py
+++ b/atomic_dbus.py
@@ -30,6 +30,7 @@ from Atomic.update import Update
 from Atomic.uninstall import Uninstall
 from Atomic.verify import Verify
 from Atomic import util
+from gi.repository import GLib
 
 DBUS_NAME_FLAG_DO_NOT_QUEUE = 4
 DBUS_REQUEST_NAME_REPLY_PRIMARY_OWNER = 1
@@ -95,6 +96,7 @@ class atomic_dbus(slip.dbus.service.Object):
             self.rootfs = []
             self.rpms = False
             self.save = False
+            self.scan_id = None
             self.scan_targets = []
             self.scan_type = None
             self.scanner = None
@@ -119,6 +121,7 @@ class atomic_dbus(slip.dbus.service.Object):
         self.tasks = []
         self.tasks_lock = threading.Lock()
         self.last_token = 0
+        self.scans = {}
         self.scheduler_thread = threading.Thread(target = self.Scheduler)
         self.scheduler_thread.daemon = True
         self.scheduler_thread.start()
@@ -364,6 +367,7 @@ class atomic_dbus(slip.dbus.service.Object):
         args.all = _all
         args.images = images
         args.containers = containers
+        args.scan_id = None
         scan.set_args(args)
         return scan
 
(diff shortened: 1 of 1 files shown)


baude:
@larskarlitski @cgwalters  please take a peek

baude:
[Make atomic scan async [dbus]](https://trello.com/c/wwTXcRiI/320-make-atomic-scan-async-dbus)

atomic_dbus.py
```diff
@@ -374,6 +378,40 @@ def _ScanSetup(self, scan_targets, scanner, scan_type, rootfs, _all, images, con
     def Scan(self, scan_targets, scanner, scan_type, rootfs, _all, images, containers):
         return self._ScanSetup(scan_targets, scanner, scan_type, rootfs, _all, images, containers).scan()
 
+    @slip.dbus.polkit.require_auth("org.atomic.read")
+    @dbus.service.method("org.atomic", in_signature='ssss', out_signature= 's')
+    # sudo busctl --system call org.atomic /org/atomic/object org.atomic ActiveScans
+    # Only support scanning one image at a time via the async method
```
larskarlitski: The code looks more like one scan at a time for each image. Is that correct?

atomic_dbus.py
```diff
@@ -374,6 +378,40 @@ def _ScanSetup(self, scan_targets, scanner, scan_type, rootfs, _all, images, con
...
+    # Same with rootfs
+    def ScanAsync(self, scan_id, scanner=None, scan_type=None, rootfs=None):
```
larskarlitski: I don't think default arguments make sense on D-Bus. Also, D-Bus doesn't have the concept of `None`. Please handle passing empty strings instead of `None`, so that cockpit doesn't have to look up available scanners and scan_types (each of those have a default, right?)

atomic_dbus.py
```diff
@@ -374,6 +378,40 @@ def _ScanSetup(self, scan_targets, scanner, scan_type, rootfs, _all, images, con
...
+        if rootfs is None:
+            rootfs = []
```
larskarlitski: The `rootfs` parameter is annotated as a string, but the default is an array?

atomic_dbus.py
baude: @larskarlitski yes, i would prefer to have it work that way for the async scan. In other words, it is a one for one scan now.  Functionally, that should be fine.  Now that it is async, you can just rapidly submit one after another.

atomic_dbus.py
baude: Yes, here again, like scan_id, it technically can be more than one.  But like the image scanning and async, I would like to only scan one rootfs per async call.

atomic_dbus.py
baude: The named arguments represent optional arguments.  So, in this case, if you accept the defaults, you don't have to pass anything.  This points out bug that you must provide either scan_id or rootfs. I'll fix that prior to merge.

As for None, you don't have to pass anything there so I think the lack of None is moot.  We can always tweak this after the fact if you prefer.

larskarlitski: I understand. The problem is that D-Bus itself doesn't have the concept of `None`. This seems to work in the python-dbus, but not from other languages. If you change `None` to the empty string as a default (which should be fine, as empty strings aren't valid ids or paths), it will work everywhere.

baude: Updated ..

baude:
@rh-atomic-bot r+  	49fc21a

PR #887, opened 2017-02-13
Enable dbus install

baude:
Enable for performing atomic install over dbus.

### DIFF
--- a/atomic_dbus.py
+++ b/atomic_dbus.py
@@ -287,17 +287,21 @@ class atomic_dbus(slip.dbus.service.Object):
     # atomic install section
     # The Install method will install the specified image
     @slip.dbus.polkit.require_auth("org.atomic.readwrite")
-    @dbus.service.method("org.atomic", in_signature='ssbbsasas', out_signature='')
-    def Install(self, image, name, user, system, remote, setvalues, extra_args):
+    @dbus.service.method("org.atomic", in_signature='ssbbsbas', out_signature='')
+    def Install(self, image, name='', system=False, remote=False, storage='', user=False, setvalues=''):
+        if setvalues is None:
+            setvalues = []
+        assert(isinstance(setvalues, list))
         i = Install()
         args = self.Args()
         args.image = image
         args.name = name
         args.user = user
         args.system = system
+        args.storage = storage
         args.remote = remote
         args.setvalues = setvalues
-        args.args = extra_args
+        args.args = []
         i.set_args(args)
         return i.install()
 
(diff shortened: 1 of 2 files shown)


baude:
[(3) [Dbus] Add support for atomic install](https://trello.com/c/Ca2xDnHa/268-3-dbus-add-support-for-atomic-install)

atomic_dbus_client.py
```diff
@@ -108,7 +108,7 @@ def ImageVersion(self, image, recurse=False):
         return self.dbus_object.ImageVersion(image, recurse, dbus_interface="org.atomic")
 
     @polkit.enable_proxy
-    def Install(self, image, name=None, user=False, system=False, remote="", setvalues=None, extra_args=None):
+    def Install(self, image, name=None, system=False, remote=False, storage=None, user=False, setvalues= None, extra_args=None):
```
rhatdan: Looks like you have a bogus space after setvalues.

atomic_dbus.py
```diff
@@ -283,14 +283,19 @@ def VulnerableInfo(self):
     # atomic install section
     # The Install method will install the specified image
     @slip.dbus.polkit.require_auth("org.atomic.readwrite")
-    @dbus.service.method("org.atomic", in_signature='ssbbsasas', out_signature='')
-    def Install(self, image, name, user, system, remote, setvalues, extra_args):
+    @dbus.service.method("org.atomic", in_signature='ssbbsbasas', out_signature='')
+    def Install(self, image, name=None, system=False, remote=False, storage=None, user=False, setvalues=None, extra_args=None):
+        if setvalues is None:
+            setvalues = []
+        if extra_args is None:
```
rhatdan: Should we do the extra check for whether setvalues and exta_args if passed in is a list?  Actually I think dbus would fail if they were not a list.

extra_args is weird, since it needs to precisely be setup by the user to match the internals of the Atomic python library.
I take it, this is just for debuging?

atomic_dbus_client.py
baude: Fixed

atomic_dbus.py
baude: Not for debug.  My understanding is that extra_args needs to be a list.  So if nothing is passed (i.e. None), we just pass an empty list.

baude: I added assertions that set_values and extra_args must be a list.

rhatdan: Do you ever see someone passing extra_args across dbus?  I think we should just remove it from the dbus inteface and hard code it to [ ]

baude: I have no idea. I was just following suit.  I'll remove it and hard set it like you said.

baude: Done

rhatdan: I meant hard code it on the server side and don't allow it to be passed from the client side.

baude: Rebased and fixed.

baude:
@rhatdan should be ready

baude:
@rh-atomic-bot r+ 7a02e29

PR #888, opened 2017-02-13
Refactor Uninstall and enable dbus

baude:


Atomic/uninstall.py
```diff
@@ -26,7 +30,12 @@ def cli(subparser):
                             action="store_true",
                             help=_("remove all containers based on this "
                                    "image"))
+    uninstallp.add_argument("--display", default=False, action="store_true", help=_("preview the command that %s would execute") % sys.argv[0])
     uninstallp.add_argument("image", help=_("container image"))
+    uninstallp.add_argument("--storage", dest="storage", default=None,
+               help=_("Specify the storage. Default is currently '%s'.  You can"
```
rhatdan: Fix indent

atomic_dbus.py
```diff
@@ -536,14 +537,16 @@ def TrustDefaultPolicy(self, default_policy):
...
+    def Uninstall(self, image, name=None, force=False, storage=None, extra_args=None):
         i = Uninstall()
         args = self.Args()
         args.image = image
         args.name = name
         args.force = force
-        args.args = extra_args
+        args.storage = storage
+        args.extra_args = [] if not extra_args else extra_args
+        assert(isinstance(args.extra_args, list))
```
rhatdan: I think dbus will do this for you.

bash/atomic
```diff
@@ -587,6 +587,18 @@ _atomic_containers_update() {
 }
 
 _atomic_uninstall() {
+	local all_options="
+	    --display
+	    --force
+		--help
```
rhatdan: indent

rhatdan:
@giuseppe PTAL

Atomic/uninstall.py
```diff
@@ -36,34 +45,23 @@ def __init__(self):
...
-
-        args = self._get_args("UNINSTALL")
-        if args:
-            cmd = self.gen_cmd(args + self.quote(self.args.args))
-            cmd = self.sub_env_strings(cmd)
-            self.display(cmd)
-            util.check_call(cmd, env=self.cmd_env())
+            be, img_obj = beu.get_backend_and_image_obj(self.args.image, str_preferred_backend=self.args.storage)
+        except ValueError as e:
+            if 'ostree' in [x().backend for x in beu.available_backends]:
```
giuseppe: why do we need this special logic for OSTree?

baude: Because the logic for uninstall differs for the docker and ostree backends.  In the case of docker, we look for an image and then remove its associated containers.  In the case of ostree, we remove the container.   Therefore, for ostree, I do two things conditionally.  First I have to make sure that ostree is an available backend and then I see if it has the container.

Atomic/uninstall.py
baude: Fixed

atomic_dbus.py
baude: Yeah it is enforced by the signature. However, given it can be None if not provided, I added the assert.  But I think the line prior enforces that.  So removing...

bash/atomic
baude: done

Atomic/uninstall.py
giuseppe: can't this be done in the ostree backend so we don't need to manage it here in uninstall?

Anyway, not a blocker, can be refactored later if possible

baude: The problem is we need to first figure out the image or container in question.  And if they exist, we call the backend methods.  In the case of docker, it's an image ... in the case of ostree, it's a container. 

One possibility would be that we could create a method in backendutils like get_object_and_backend_for_uninstall and that would obfuscate it.  Then just call be.uninstall(object).  I'm game for that.  Prefer that?

giuseppe: could we do something like?
for each backend be do:
    handled = be.uninstall(object)
    if handled:
        break

But we can refactor this later and go with with version for now, I am fine with it

baude: No bc the uninstall method requires an object.  The object cannot be found if it doesn't know what it is looking for.

giuseppe:
@rh-atomic-bot r+  	f460859

PR #889, opened 2017-02-14
Show an error if atomic images list -f FILTER=VALUE uses a not known filter

giuseppe:
plus two other minor fixes

### DIFF
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -71,7 +71,11 @@ class DockerBackend(Backend):
                              "matching '{0}'. They are:\n    {1} \n{2}"
                              .format(img, "\n    ".join(repo_tags),
                                      err_append))
-        return self._make_image(img, self._inspect_image(img), deep=True)
+
+        inspected = self._inspect_image(img)
+        if not inspected:
+            return None
+        return self._make_image(img, inspected, deep=True)
 
     def already_has_image(self, local_img, remote_img):
         """
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -187,9 +187,9 @@ class Containers(Atomic):
                 return True
             for f in self.args.filters:
                 _filter, _ = f.split('=', 1)
-                if _filter not in [x for x in self.FILTER_KEYWORDS]:
+                if _filter not in self.FILTER_KEYWORDS.keys():
                     raise ValueError("The filter {} is not valid.  "
-                                     "Please choose from {}".format(_filter, [x for x in self.FILTER_KEYWORDS]))
+                                     "Please choose from {}".format(_filter, self.FILTER_KEYWORDS.keys()))
         _check_filters()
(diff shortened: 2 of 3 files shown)


Atomic/backends/_docker.py
```diff
@@ -71,7 +71,10 @@ def has_image(self, img):
                              "matching '{0}'. They are:\n    {1} \n{2}"
                              .format(img, "\n    ".join(repo_tags),
                                      err_append))
-        return self._make_image(img, self._inspect_image(img), deep=True)
+
+        img = name_search[0]['RepoTags'][0]
+        inspected_image = self.inspect_image(image=img)
+        return inspected_image
```
baude: I'm OK with this approach but have you checked if we rely on an error being raised if nothing is found? My guts tell me we do.

baude: Perhaps this is the failing unittest?

giuseppe: I am not sure what should happen here, the issue is that self._inspect_image(img) is None and that will fail the make_image function.  What should happen with "atomic install busybox" ?

giuseppe:
@baude fixed the test failure

Atomic/containers.py
```diff
@@ -187,9 +187,9 @@ def _check_filters():
                 return True
             for f in self.args.filters:
                 _filter, _ = f.split('=', 1)
-                if _filter not in [x for x in self.FILTER_KEYWORDS]:
+                if _filter not in self.FILTER_KEYWORDS.keys():
                     raise ValueError("The filter {} is not valid.  "
-                                     "Please choose from {}".format(_filter, [x for x in self.FILTER_KEYWORDS]))
+                                     "Please choose from {}".format(_filter, self.FILTER_KEYWORDS.keys()))
```
yuqi-zhang: The thing about using keys() directly is that it will display: 
`Please choose from dict_keys(['state', 'container', 'command', 'backend', 'runtime', 'image', 'created'])`
instead of:
`Please choose from ['repo', 'tag', 'id', 'created', 'size', 'type', 'dangling']`
Which looks a little worse. Not a huge deal.

giuseppe: hm.. I get exactly the same output, also:

```
>>> h = {'a' : 1, 'b' : 2}
>>> print("message {}".format(h.keys()))
message ['a', 'b']
>>> print("message {}".format([x for x in h]))
message ['a', 'b']
```

What version of Python are you using?

```

yuqi-zhang: You're right, this only happens in Python3.5 and not 2.7, interesting

giuseppe:
@yuqi-zhang I pushed a fixup patch to address it

yuqi-zhang:
Thanks, LGTM!

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot r=rhatdan f58de44

PR #891, opened 2017-02-14
Fix re module bug

aweiteka:


rhatdan:
Does this fix the problem on RHEL?  Should we have @lsm5 grab this patch for the RHEL7 atomic command?

aweiteka:
> Does this fix the problem on RHEL? Should we have @lsm5 grab this patch for the RHEL7 atomic command?

Yes and yes

rhatdan:
@rh-atomic-bot r=rhatdan fc5536a

ddarrah:
fix is good on RHEL.

PR #894, opened 2017-02-15
Account for API Changes in docker-py-2

baude:
The new 2.x version of the docker python API has non-backward
compatible changes.  These changes are described here:

https://docker-py.readthedocs.io/en/stable/change-log.html#breaking-changes

We need to account for docker.Client and docker.APIClient as well
as changes in the way kwargs are handled.  Also, it appears the
AutoVersion method is deprecated.

### DIFF
--- a/Atomic/backends/_docker_errors.py
+++ b/Atomic/backends/_docker_errors.py
@@ -1,0 +1,7 @@
+class DockerObjectNotFound(ValueError):
+    def __init__(self, msg):
+        super(DockerObjectNotFound, self).__init__("Unable to associate '{}' with an image or container".format(msg))
+        
+class NoDockerDaemon(Exception):
+    def __init__(self):
+        super(NoDockerDaemon, self).__init__("The docker daemon does not appear to be running.")
--- a/Atomic/client.py
+++ b/Atomic/client.py
@@ -1,17 +1,22 @@
 import docker
-from docker.utils import kwargs_from_env
 import sys
 import requests
+from Atomic.backends._docker_errors import NoDockerDaemon
 
 def get_docker_client():
     """
     Universal method to use docker.client()
     """
+    kwargs = {"version": "auto"}
+    kwargs.update(docker.utils.kwargs_from_env(assert_hostname=False))
     try:
-        return docker.AutoVersionClient(**kwargs_from_env())
-
+        client = docker.APIClient #pylint: disable=no-member
+    except AttributeError:
+        client = docker.Client #pylint: disable=no-member
+    try:
+        return client(**kwargs)
     except docker.errors.DockerException:
-        return docker.Client(**kwargs_from_env())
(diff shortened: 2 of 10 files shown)


Atomic/backends/_docker_errors.py
```diff
@@ -0,0 +1,7 @@
+class DockerObjectNotFound(ValueError):
```
rhatdan: Why not just define these in 
backends/_docker?

Atomic/client.py
```diff
@@ -51,7 +56,10 @@ def __getattribute__(self, name):
         if name == "_dockerclient" or name == "_client":
             return object.__getattribute__(self, name)
         obj = self._dockerclient
-        attr = docker.AutoVersionClient.__getattribute__(obj, name)
+        try:
```
rhatdan: Could you reverse these.  Since going forward APIClinet will be default lets make that the primary and Client be the fall back.

Atomic/backends/_docker_errors.py
baude: Due to circular imports.

Atomic/client.py
baude: OK, fixed.

Atomic/backends/_docker_errors.py
rhatdan: Please explain?

baude: _docker.py imports client.py; therefore, client.py cannot import from _docker.py.  Furthermore, several other files need access to these objects which would likely cause more circular issues.  I wasn't able offhand to figure out how to work around this with certainty other than breaking them out to their own file.

rhatdan: OK, one other question, why are you using the _NAME.py syntax for these python objects?

baude: I dont understand the question. Do you mean camel case?

baude:
@TomasTomecek mind taking a peek?

TomasTomecek:
> AutoVersion method is deprecated.

It's not deprecated. There is no such class in version 2.

Atomic/backends/_docker_errors.py
rhatdan: Why is _docker_errors.py called that rather then docker_errors.py?

baude: dealer's choice.

TomasTomecek:
I just tested this PR briefly and can confirm it works with docker-py-1.10.6 and python-docker-2.1.0. All I did was to just run `atomic images list`. Well done @baude!

rhatdan:
Ok lets get this into Rawhide to see if we have any other issues.

@rh-atomic-bot r+ 621da8b

TomasTomecek:
Just one thing **↑**.

PR #896, opened 2017-02-15
.redhat-ci.sh: Correct avoid selinux issues

baude:
At present, when overlayfs is used on RHEL and Centos to unlock
ostree, selinux and permissions get set with the wrong permissions.

This is a work-around for now.

### DIFF
--- a/.redhat-ci.sh
+++ b/.redhat-ci.sh
@@ -5,15 +5,22 @@ NO_TEST=${NO_TEST:-}
 mount --make-rshared /
 
 
-if [ -f /run/ostree-booted ]; then
+if [ -f /run/ostree-booted ] && grep -q ID=fedora /etc/os-release; then
     if [ ! -e /var/tmp/ostree-unlock-ovl.* ]; then
         ostree admin unlock
+    fi
+elif [ -f /run/ostree-booted ]; then
+    # Until overlayfs and selinux get along, use remount
+    # instead of ostree admin unlock
+    if [ ! -w /usr ]; then
+	mount -o remount,rw /usr
     fi
 else
     dnf install -y atomic python3-coverage
 fi
 
-systemctl start docker
+# Restarting docker helps with permissions related to above.
+systemctl restart docker
 
 # somewhat mimic the spec conditional
 source /etc/os-release


baude:
@jlebon mind reviewing?

.redhat-ci.sh
```diff
@@ -7,13 +7,18 @@ mount --make-rshared /
 
 if [ -f /run/ostree-booted ]; then
     if [ ! -e /var/tmp/ostree-unlock-ovl.* ]; then
-        ostree admin unlock
+	# Until overlayfs and selinux get along, use remount
```
cgwalters: Can we keep doing the unlock for Fedora?  Should be basically `if grep -q ID=fedora /etc/os-release` ?

.redhat-ci.sh
```diff
@@ -7,13 +7,18 @@ mount --make-rshared /
 
 if [ -f /run/ostree-booted ]; then
     if [ ! -e /var/tmp/ostree-unlock-ovl.* ]; then
```
jlebon: This should become something like e.g. `findmnt /usr -o OPTIONS | grep -q rw`.

.redhat-ci.sh
```diff
@@ -7,13 +7,18 @@ mount --make-rshared /
...
+        #ostree admin unlock
+	mount -o remount,rw /usr
     fi
 else
     dnf install -y atomic python3-coverage
 fi
 
-systemctl start docker
+# Restarting docker helps with permissions related to above.
+systemctl restart docker
```
jlebon: You only need to do this if we did the remount though.

.redhat-ci.sh
cgwalters: Maybe just `test -w /usr`.  But I'm fine with either.

.redhat-ci.sh
baude: Well it is a catch all.  In the case of the cloud image, docker is installed and not started.  So the restart here catches that.

.redhat-ci.sh
baude: Got it.

.redhat-ci.sh
baude: Yeah, check out updated.

baude:
@cgwalters @jlebon you guys good with this now? if so, mind merging?

cgwalters:
Seems good to me...if @jlebon agrees I'll let him do the r+.

jlebon:
@rh-atomic-bot r+ 177f615

PR #899, opened 2017-02-16
Fix Stop Regression: BZ #1422448

baude:
BZ #1422448 actually points out one regression and one
bug related to atomic stop. The BZ itself points out
a TypeError exception in the code when trying to glue
together a python list and str.

Then the atomic stop command was failling in the case where a
STOP label was defined and it uses the variable $NAME resulting
in a subprocess exception.

The self.name variable was not being set by set_args after
refactoring occured.  Ideally, this should all be moved
into the image|container object handling but for now we
just handle it in set_args.

Also added a test in test_display to catch any future
regressions.

### DIFF
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -168,7 +168,12 @@ class DockerBackend(Backend):
         except KeyError:
             con_obj.command = con_struct['Config']['Cmd']
 
-        con_obj.state = con_struct.get('State', None) or con_struct.get['State'].get('Status', None)
+        try:
+            con_obj.state = con_struct.get('State', None) or con_struct.get['State'].get('Status', None)
+        except TypeError:
+            # Docker 1.10 on F24 has a different structure
+            con_obj.state = con_struct['Status']
+
         if isinstance(con_obj.state, dict):
             con_obj.state = con_obj.state['Status']
         con_obj.running = True if con_obj.state.lower() in ['true', 'running'] else False
@@ -237,7 +242,25 @@ class DockerBackend(Backend):
             raise ValueError("Unable to locate container '{}' in {} backend".format(name, self.backend))
         return self.d.start(name)
 
-    def stop_container(self, con_obj):
+    def stop_container(self, con_obj, **kwargs):
+        atomic = kwargs.get('atomic')
+        args = kwargs.get('args')
+        con_obj.stop_args = con_obj.get_label('stop')
+        if con_obj.stop_args:
+            try:
+                cmd = atomic.gen_cmd(con_obj.stop_args.split() + atomic.quote(args.args))
+            except TypeError:
+                cmd = atomic.gen_cmd(con_obj.stop_args + atomic.quote(args.args))
+            cmd = atomic.sub_env_strings(cmd)
+            atomic.display(cmd)
+            if args.display:
+                return 0
+            # There should be some error handling around this
+            # in case it fails.  And what should then be done?
+            return util.check_call(cmd, env=atomic.cmd_env())
+        elif args.display:
+            return 0
+
         return self.d.stop(con_obj.id)
 
 
--- a/Atomic/stop.py
+++ b/Atomic/stop.py
@@ -2,6 +2,7 @@ import argparse
 
 from . import util
 from Atomic.backendutils import BackendUtils
+import sys
 
 try:
     from . import Atomic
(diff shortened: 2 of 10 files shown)


Atomic/backends/_docker.py
```diff
@@ -237,7 +242,23 @@ def start_container(self, name):
...
+            except TypeError:
+                cmd = atomic.gen_cmd(con_obj.stop_args + atomic.quote(args.args))
+            cmd = atomic.sub_env_strings(cmd)
+            atomic.display(cmd)
+            if args.display:
+                return 0
+            # There should be some error handling around this
+            # in case it fails.  And what should then be done?
+            return util.check_call(cmd, env=atomic.cmd_env())
+
```
giuseppe: should args.display be honored also in the case there is not a label defined?

Should we print something like "docker stop " + con_obj.id?

baude: Previously we did not so I had planned to preserve that behaviour.

baude: More to the point, all we do is call the API to stop it.  So I don't believe we need to show anything.

giuseppe: I was commenting that that when `--display` is used, we should not stop the container, not the output itself.  From the code it seems that atomic doesn't check for `--display` when there is no stop label, so the container is always stopped, also when `--display` is used.

baude: @giuseppe better?

rhatdan:
@baude Rebase...

baude:
@rh-atomic-bot retry

baude:
@rh-atomic-bot retest

cgwalters:
@baude See https://github.com/jlebon/redhat-ci/pull/25 for usage instructions

Atomic/backends/_docker.py
```diff
@@ -237,7 +242,26 @@ def start_container(self, name):
...
+                cmd = atomic.gen_cmd(con_obj.stop_args + atomic.quote(args.args))
+            cmd = atomic.sub_env_strings(cmd)
+            atomic.display(cmd)
+            if args.display:
+                return 0
+            # There should be some error handling around this
+            # in case it fails.  And what should then be done?
+            return util.check_call(cmd, env=atomic.cmd_env())
+        else:
+            util.write_out("There is no STOP label defined.")
```
rhatdan: I think we still want to stop the container if it does not have a stop label?

Atomic/stop.py
```diff
@@ -17,6 +18,8 @@ def cli(subparser):
     stopp.set_defaults(_class=Stop, func='stop')
     util.add_opt(stopp)
     stopp.add_argument("container", help=_("container name or ID"))
+    stopp.add_argument("--display", default=False, action="store_true",
+                            help=_("preview the command that %s would execute") % sys.argv[0])
```
rhatdan: Nit Fix tab.

Atomic/backends/_docker.py
baude: This was supposed to be in a conditional checking for args.display.  Agree and fixed.

Atomic/stop.py
baude: Fixed

Atomic/backends/_docker.py
```diff
@@ -237,7 +242,27 @@ def start_container(self, name):
...
+            except TypeError:
+                cmd = atomic.gen_cmd(con_obj.stop_args + atomic.quote(args.args))
+            cmd = atomic.sub_env_strings(cmd)
+            atomic.display(cmd)
+            if args.display:
+                return 0
+            # There should be some error handling around this
+            # in case it fails.  And what should then be done?
+            return util.check_call(cmd, env=atomic.cmd_env())
+        elif args.display:
```
rhatdan: Remove this, we don't display errors on other missing labels. (Run, Uninstall, Install).  No label showing up should tell the user that their is no label.

rhatdan:
Do we care to run the STOP Label on system containers? @giuseppe

Atomic/backends/_docker.py
baude: Removed the write out statement.

giuseppe:
No, we don't use labels with system containers.  System containers are started and stopped via their systemd service

rhatdan:
Ok
@rh-atomic-bot r+ 5fb2d2c

PR #900, opened 2017-02-16
add --lvsize option to atomic storage modify

shishir-a412ed:

### DIFF
--- a/Atomic/storage.py
+++ b/Atomic/storage.py
@@ -65,6 +65,7 @@ def cli(subparser):
     modifyp.add_argument('--rootfs', dest="rootfs", default=None,
                          help='Mountpath where logical volume for container storage would be mounted. e.g. /var/lib/containers')
     modifyp.add_argument('--lvname', dest="lvname", default=None, help='logical volume name for container storage. e.g container-root-lv')
+    modifyp.add_argument('--lvsize', dest="lvsize", default=None, help='logical volume size for container storage.')
     modifyp.add_argument('--driver', dest="driver", default=None, help='The storage backend driver', choices=['devicemapper', 'overlay', 'overlay2'])
     modifyp.add_argument('--vgroup', dest="vgroup", default=None, help='The storage volume group')
     modifyp.add_argument("--graph", dest="graph",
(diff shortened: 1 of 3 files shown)


shishir-a412ed:
ping @rhatdan @baude 

PS: I am working on a separate PR which will add an integration test for all three options {--lvname, --lvsize and --rootfs}

Shishir

rhatdan:
man pages.

rhatdan:
Should we do any error checking?

shishir-a412ed:
@rhatdan 
So the way I wrote it:

1) If `--lvsize` is passed without passing both `--rootfs` and `--lvname`, it will just ignore it. Since `--lvsize` needs to be set along rootfs and lvname. This is also mentioned in man page.

2) If someone passes one of them {`--rootfs` or `--lvname`} along with `--lvsize`, it will error out based on existing checks.

3) If someone passes `--lvsize` in the wrong format. (How to pass the value in correct format is described in man page), docker-storage-setup would throw an error, which atomic should print to STDERR.

I think all the error handling is done either at atomic level or d-s-s.

Shishir

rhatdan:
It will throw an error at a much later time.  I would rather have it tell the user right a way they did a typo or an invalid format.

shishir-a412ed:
@rhatdan PTAL.

Atomic/storage.py
```diff
@@ -171,6 +172,11 @@ def modify(self):
             if self.args.driver:
                 self._driver(self.args.driver)
             if self.args.rootfs and self.args.lvname:
+                if self.args.lvsize:
+                    cmd = 'bash -c ". /usr/lib/docker-storage-setup/libdss.sh; check_data_size_syntax {0}"'.format(self.args.lvsize)
+                    if os.system(cmd) != 0:
```
rhatdan: Use


``` 
try:
    util.check_call(cmd)
except IOError:
     raise ValueError("Invalid format for --lvsize")
```

rhatdan:
Need successfull and Failure test cases.

Atomic/storage.py
shishir-a412ed: @rhatdan This is not working. In case of an `invalid format` it just throws an empty STDOUT.
However it does return an error code of `1`. But the message `Invalid format for --lvsize` doesn't get printed. Any idea ?

rhatdan: Is the exception happening?

shishir-a412ed: Yeah it is happening. I just changed `except IOError` to `except subprocess.CalledProcessError` and it works. PTAL.

shishir-a412ed:
@rhatdan I am working on a separate PR that will have test for all 3 flags. Let's merge this one, I will open another one for the test.

rhatdan:
@rh-atomic-bot r+ fc5954e

PR #902, opened 2017-02-21
depend on version 2 of docker-py

TomasTomecek:
which is now available as 'docker' in PyPI and as a `python{2,3}-docker`
in Fedora.

Related: https://github.com/projectatomic/atomic/pull/894

I'm assuming that I shouldn't make the change in `.redhat-ci` file.


.redhat-ci.Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN dnf install -y \
         python-dbus \
         pylint \
         python-slip-dbus \
-        python-docker-py \
+        python2-docker \
```
baude: are these rpm changes already propagated to f25, rawhide, centos, and rhel?

rhatdan: Not in RHEL/Centos and maybe not in F25.

baude: @rhatdan exactly, i think this only applies to rawhide at this point.

TomasTomecek: Correct, this is only rawhide now. There is no plan to add the package to F25.

rhatdan: @jlebon Any way to handle this?

jlebon: But the container is based on F24 (should probably be updated to F25 now), so this isn't going to resolve, right?

Sounds instead like we should set up a rawhide tester (though of course, this issue there is that you'll have to deal with false positives).

TomasTomecek: It could also be possible to pull the package directly from koji. As a workaround before having F26.

jlebon: Hmm, if we go that way, I think I'd rather make it a new testsuite (they're delimited by `---` in the YAML). It's best if the F25/24 ones reflect as close as possible the actual state we release in.

TomasTomecek: Is that something I should do?

jlebon: So another issue here is that F25 has Python v3.5, while rawhide has v3.6. Are we expecting incompatibilities here? Given that we'll be branching soon, maybe let's leave the redhat-ci testsuites for now and circle around to it when we add support for F26. I opened https://github.com/jlebon/redhat-ci/issues/28 to track that work.

rhatdan: Yes I think we should wait for F26 to make this change.  I don't believe 3.5 vs 3.6 should be an issue.

TomasTomecek: Okay, so I'll leave this open and let's revisit once F26 moves forward.

TomasTomecek:
Since Fedora 26 is released, what can I do to complete this?

jlebon:
@TomasTomecek You should be able to rebase and modify `.papr.sh` accordingly.

TomasTomecek:
bot, retest this please

TomasTomecek:
The CI failure:

```
### TIMED OUT AFTER 1800s
```

giuseppe:
I think we can ignore the fedora/26/atomic failure.

ashcrow:
bot, retest this please

ashcrow:
Looks ready to merge.

TomasTomecek:
\o/

ashcrow:
cc @baude

ashcrow:
@baude / @rhatdan: **bump**

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #903, opened 2017-02-21
Fix uninstall bug: BZ 1425495

baude:
The bug https://bugzilla.redhat.com/show_bug.cgi?id=1425495 demostrates
a case where atomic attempts to "add" a list and str resulting in a
TypeError exception.  This should resolve that BZ.

### DIFF
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -375,7 +375,10 @@ class DockerBackend(Backend):
 
         cmd = []
         if uninstall_command:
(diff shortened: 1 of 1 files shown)


baude:
@miabbott please ack if this fixed the bug

miabbott:
This change fixed the issue on my RHEL Atomic Host where I first saw the problem.

rhatdan:
@rh-atomic-bot r+ cb6ed42

PR #907, opened 2017-02-22
Enable atomic push in dbus and add integrations tests

baude:
For potential cockpit use, we have enabled atomic push.  It returns
a 0 if successful or throws a dbus Exception upon failure.  Added
a series of integrations tests as well.


baude:
[(5) [Dbus] Add support for atomic push](https://trello.com/c/3SItlEXr/236-5-dbus-add-support-for-atomic-push)

rhatdan:
Should we add a dbus interface to return registry choices for cockpit to display?

baude:
@rhatdan We sure can, but ideally we should be able to detect this on the fly if we can ever get the registry API to just change a bit.   Let me know what you want and I'll do it.

rhatdan:
@rh-atomic-bot r+

rhatdan:
@rh-atomic-bot r+ b956711

PR #912, opened 2017-02-24
Atomic/util.py: default_docker detects wrong library BZ# 1381696

baude:
When Atomic unmount is called on an overlayfs based dockerd,
the default docker library is incorrectly determined resulting in
failures to unmount.

We now first ask the running dockerd what its root library directory
is and use that if dockerd is running.

This resolves https://bugzilla.redhat.com/show_bug.cgi?id=1381696


rhatdan:
@rh-atomic-bot r+ 1803b18

baude:
@rh-atomic-bot retry

Issue #910, opened 2017-02-24
atomic-1.16.1 storage reset is broken


rhatdan:
I get this nice sequence.

```
# atomic storage reset
The docker daemon does not appear to be running.

# systemctl start docker

# atomic storage reset
Docker daemon must be stopped before resetting storage
```

rhatdan:
@baude PTAL
I think we are forcing docker to always be running, which is wrong,  atomic should be able to run without docker.

PR #913, opened 2017-02-24
Atomic/storage.py: Prevent circular dep on dockerd

baude:
atomic storage reset requires that the dockerd be inactive.  However,
calling the d.info() requires that docker be active.  Adding the
NoDockerDaemon exception allows it to progress without the daemon.

This fixes https://github.com/projectatomic/atomic/issues/910


rhatdan:
@rh-atomic-bot r+ 0adf422

rhatdan:
@rh-atomic-bot r=rhatdan 0adf422

PR #915, opened 2017-02-27
Fix issues: Atomic storage import and Atomic storage modify.

shishir-a412ed:


rhatdan:
Storage tests are failing.

shishir-a412ed:
@rhatdan PTAL.

rhatdan:
@rh-atomic-bot r+ da96a69

mvollmer:
Hmm, I was hoping that `atomic storage` would one day completely wrap docker-storage-setup, or even replace it.  What is the reason for requiring the user to run it?

shishir-a412ed:
@mvollmer That's the idea. `atomic storage` consuming `docker-storage-setup`.
Right now docker-storage-setup (now container-storage-setup) is being transformed at a rapid pace with new features being added to upstream/master.

`What is the reason for requiring the user to run it?` : Run what ? container-storage-setup or atomic storage ?

Shishir

mvollmer:
> Run what ? container-storage-setup or atomic storage ?

container-storage-setup.  For example, you had to insert explicit invocations of docker-storage-setup into test-storage.sh.  The same needs to be done in other places that use `atomic storage`, such as Cockpit.

Why do you break the users of `atomic storage`?  I am sure you have a good reason, but I can't see it spelled out anywhere.

shishir-a412ed:
@mvollmer

container-storage-setup has full backward compatibility with docker-storage-setup. A user can still run docker-storage-setup. N we intend to maintain backward compatibility going forward. I am sorry if I am missing your point, but feel free to ping me on irc (shishir-a412ed) and we can sort it out. I hang out on #atomic and #aos.

rhatdan:
@mvollmer Think of container-storage-setup as being a low level tools for configuring container storage.  `atomic storage` will be the CLI for interacting with it.  We will execute the low level routines from the unit files.  Replacing a pretty complex bash script that executes lots of lvm commands with full python does not seem like a good use of our time.

mvollmer:
> container-storage-setup has full backward compatibility with docker-storage-setup.

Except for the name. :-)

My point is that previously this was enough to make the docker thinpool bigger:
```
    # atomic storage modify --add-device /dev/sdc
```
This would either work and the pool is bigger afterwards, or it would make no change to the configuration.

With this change here, one now has to do this to get the same effect:
```
   # atomic storage modify --add-device /dev/sdc
   # container-storage-setup                (or docker-storage-setup)
```
This is a change in the 'API' of `atomic storage modify`, and it is also worse: If /dev/sdc can not be added to the pool for some reason, docker-storage-setup will fail, but the configuration has changed and docker-storage-setup will continue to fail every time it runs.  The user has to manually clean up the configuration.

mvollmer:
> Think of container-storage-setup as being a low level tools for configuring container storage. atomic storage will be the CLI for interacting with it. 

Okay, so a user of `atomic storage` is not required to know that container-storage-setup exists and never needs to interact with it directly?

That's what I thought, but I am afraid that this PR is breaking that.

> We will execute the low level routines from the unit files.

When?  Would `atomic storage modify` still have immediate effect, or would one need to reboot or restart docker to see any actual change to the storage setup?

rhatdan:
Yes I agree this is a regression and we should fix it.  I think we should attempt to avoid telling users to run container-storage-setup.  This should all be handled in the atomic command or by telling the user he needs to restart the docker daemon.

shishir-a412ed:
@mvollmer @rhatdan 
container-storage-setup (previously docker-storage-setup) was designed to run once, setup the storage and exit. it was not designed as a command line tool to pass options and run again and again. 
atomic is a command line tool, with different set of expectations. The premise of the two tools is entirely different.

I run `atomic storage modify` once (it runs container-storage-setup once), I run `atomic storage modify` again (it runs container-storage-setup again). I don't think the current implementation of container-storage-setup was designed to handle this. It is a bash script. The error handling is at minimum. If the container-storage-setup fails, the storage is already messed up. N From that point on the user is only doing more damage to the storage configuration (Unless he resets his configurations). 
My point being they were initially (container-storage-setup, atomic) designed for different use cases.
Having said that, we are now moving in a direction where container-storage-setup can run more than once, do a lot more independent operations. (But it's gonna take a while). 

Shishir

shishir-a412ed:
@mvollmer 
Reply: https://github.com/projectatomic/atomic/pull/915#issuecomment-285608464
"Except for the name :)" You can use docker-storage-setup if you like, it would still run container-storage-setup and provide backward compatibility.

I agree `atomic storage modify` before used to update the config file (/etc/sysconfig/docker-storage-setup) and run docker-storage-setup to setup the storage configuration AND now it only updates the config file (instructing the user to run docker-storage-setup manually). I agree it is an API change and could be construed as even worse for the reasons you mentioned.

If you like we can go back to the old behavior (I can open a PR for that) however I think both the previous and current behavior don't integrate well with container-storage-setup (Reasons outlined in my previous comment). We need to redesign the whole thing (atomic storage modify and container-storage-setup) to integrate well with each other, rather than doing plumbing to fix issues (and not fixing the root cause of the problem: design).

Shishir

shishir-a412ed:
@rhatdan 
Reply: https://github.com/projectatomic/atomic/pull/915#issuecomment-285685796
"I think we should attempt to avoid telling users to run `container-storage-setup`." So what would be the expected behavior ?

a) Original behavior (`atomic storage modify` should run `container-storage-setup`, that would imply. everytime a user runs `atomic storage modify`, `container-storage-setup` would be called. Not sure if that's a good idea).
b) If not whats the new behavior you are contemplating ?

"This should all be handled in the atomic command or by telling the user he needs to restart the docker daemon."

How does restarting docker daemon helps in storage configuration ? Can you elaborate a bit more.
Like I suggested I don't think we should be doing more plumbing, rather fix the root cause.

Shishir

rhatdan:
We have an effort to support container-storage-setup update which we can use in atomic.

The behaviour should be fixed in the code of atomic and container-storage-setup, not by directing humans who might not even have access to the system to execute random commands.

rhatdan:
@shishir-a412ed think of the machine as being managed by cockpit.  All I can do is execute commands in atomic dbus and start/stop the container runtime.

If container-storage-setup can not be run a second time or I am not allowed to change some content that should be reported back the the user maybe with a hint to execute atomic storage reset.

shishir-a412ed:
@rhatdan Right, that brings back to my first point that both `container-storage-setup` and `atomic` needs to be redesigned to have a seamless integration between these two tools.

Let's discuss this in office, when we have further meetings on `container-storage-setup`. Vivek is already putting a formal design document that has more details.

Shishir

mvollmer:
> Let's discuss this in office, when we have further meetings on container-storage-setup. Vivek is already putting a formal design document that has more details.

This PR has been released with atomic 1.16.2.  Will you revert it in 16.3?

Or in other words, should we (Cockpit) treat the changed behavior of `atomic storage modify` as a bug and just sit it out as a temporary regression, or do we need to adapt our code to work with both the new behavior and the old?

mvollmer:
> You can use docker-storage-setup if you like, it would still run container-storage-setup and provide backward compatibility.

Ahh, that's cool.

shishir-a412ed:
@mvollmer For now I think it would be best to revert to old behavior. So that Cockpit can use atomic without breaking the existing API (We can introduce that change in 1.16.3) i.e. `atomic storage modify` will update the config file and call container-storage-setup. I will open a PR for that.

/cc @rhatdan 

Shishir

mvollmer:
> For now I think it would be best to revert to old behavior. So that Cockpit can use atomic without breaking the existing API (We can introduce that change in 1.16.3) i.e. atomic storage modify will update the config file and call container-storage-setup. I will open a PR for that.

Nice, thanks!

Issue #916, opened 2017-02-27
atomic scan incorrectly copies ostree content


cgwalters:
See https://github.com/ostreedev/ostree/pull/491#issuecomment-282802207

Looking at the code, I think there should be a `def get_transient_mountpoint_base()` defined per backend.  Then `syscontainers.py` can provide that as `return /ostree/repo/tmp/`, the docker backend can have it as `/run/atomic` as it is now.

scan would then combine the base with its chosen container names.

This would allow the ostree backend to do what it does best - create lots of hardlinks.

baude:
Is there some subtlety I am not catching between /run/atomic and /ostree/repo/tmp ?  Can you elaborate?

Also, the mountpoint dir likely will need to be somewhat dynamic because multiple scans can occur simultaneously, so we can use a namedtemp dir or something for that.

cgwalters:
They're on different filesystems.  `/run` is a tmpfs.  `/ostree/repo/tmp` is on the same physical filesystem as `/ostree/repo`, and hence `link()` works.

cgwalters:
I understand the dynamic nature of the mountpoint directory.  I think the backend code should provide a *prefix*.  (That's what I meant with the `_base()` in the API but I wasn't clear enough)

baude:
I'm working my way through this.  The first step is to perform the backend refactor for scan.py.

The next complicating problem is that atomic scan is designed around passing in an in dir that has a bunch of chroots mounted "inside it" (a dir of dirs).  If the backends are going to have different prefix'd mount points, it challenges that base rule and all the coding assumptions around it.  I think I can work around this but more or less recording it here so I remember it and make sure I deal with it.

giuseppe:
I have some WIP here:

https://github.com/projectatomic/atomic/pull/917

It is WIP as I have not really tested it well.  As we don't really have the mount logic in the new backend stuff, I've worked directly in scan.py and mount.py.

PR #917, opened 2017-02-27
scan: optimize the OSTree case

giuseppe:
create hard links instead of copying files when possible.

Closes: https://github.com/projectatomic/atomic/issues/916

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -140,14 +140,33 @@ class Mount(Atomic):
     def _info(self):
         return self.d.info()
 
-    def mount(self):
+    def _try_ostree_mount(self, best_mountpoint_for_storage):
+        if best_mountpoint_for_storage:
+            mountpoint = os.path.join(self.syscontainers.get_ostree_repo_location(), "tmp/atomic-mount", str(os.getpid()), self.image)
+            if os.path.exists(mountpoint):
+                shutil.rmtree(mountpoint)
+            os.makedirs(mountpoint)
+        else:
+            mountpoint = self.mountpoint
+
+        d = OSTreeMount(self.args, mountpoint, live=self.live, shared=self.shared)
+        if d.mount(self.image, self.options):
+            self.mountpoint = mountpoint
+            return True
+
+        return False
+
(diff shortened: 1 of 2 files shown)


baude:
@giuseppe lets talk tomorrow.  im refactoring Scan right now. Would like to complete that and then we can fix the ostree mount point issues.

Atomic/mount.py
```diff
@@ -140,14 +140,33 @@ def set_args(self, args):
     def _info(self):
         return self.d.info()
 
-    def mount(self):
+    def _try_ostree_mount(self, best_mountpoint_for_storage):
+        if best_mountpoint_for_storage:
+            mountpoint = os.path.join(self.syscontainers.get_ostree_repo_location(), "tmp", str(os.getpid()), self.image)
```
cgwalters: Using `getpid()` in here is going to make it hard to avoid leaking them.  I'd make it `tmp/atomic-mount` or something.  Then it's easy to have a cleanup function delete that directory.  This is what we do in rpm-ostree.  Now, *inside* that directory, maybe you do want a pid.

giuseppe: thanks, I have pushed a fixup patch to change this to `/tmp/atomic-mount`.

giuseppe:
@baude The changes to scan.py are quite limited.  Are these patches going to conflict with the refactor you are working on?  I am ok to resolve these conflicts on top of your work

giuseppe:
@baude rebased on top of your series

giuseppe:
one issue I've seen is that if the image is present in both repositories, scan fails with an unfriendly:

```
ValueError: Found more than one Image with name 7968321274dc6b6171697c33df7815310468e694ac5be0ec03ff053bb135e768; please specify with --storage.
```

Not sure how to handle this, if mount fails, iterate over all the backends until the image is found?

baude:
@giuseppe ill look into that issue deeper but outside the context of this PR.  Sound good?

giuseppe:
@baude, sure just wanted to point it out so that we could discuss it

baude:
LGTM!

baude:
@rh-atomic-bot r+ 	f005964

PR #919, opened 2017-02-28
Refactor Scan

baude:
Refactor scan to use the backends as well as container and
image objects.

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -6,6 +6,7 @@ import os
 from shutil import rmtree
 import json
 import sys
+from Atomic.backendutils import BackendUtils
 
 def cli(subparser):
     # atomic scan
@@ -37,6 +38,7 @@ class Scan(Atomic):
     def __init__(self):
         super(Scan, self).__init__()
         self.scan_dir = None
+        self.scan_list = []
         self.rootfs_paths = []
         self.cur_time = datetime.now().strftime('%Y-%m-%d-%H-%M-%S-%f')
         self.chroot_dir = '/run/atomic/{}'.format(self.cur_time)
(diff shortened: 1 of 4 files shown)


Atomic/scan.py
```diff
@@ -178,50 +177,70 @@ def set_scanner():
...
+        #        docker_object = docker_object if docker_object is not None \
+        #            else (next((item for item in images if item['Id'] == self.get_input_id(scan_input)), None))
+        #        if docker_object is not None:
+        #            docker_object['input'] = scan_input
+        #            scan_list.append(docker_object)
+        #    if len(scan_list) < 1:
+        #        raise ValueError("You must provide at least one container or image for atomic "
+        #                         "scan. See 'atomic scan --help' for more information")
+        #
+        #return scan_list
```
giuseppe: let's drop completely this dead code

Atomic/scan.py
```diff
@@ -178,50 +177,70 @@ def set_scanner():
...
-                    docker_object['input'] = scan_input
-                    scan_list.append(docker_object)
-            if len(scan_list) < 1:
-                raise ValueError("You must provide at least one container or image for atomic "
-                                 "scan. See 'atomic scan --help' for more information")
-
+            for scan_target in self.args.scan_targets:
+                try:
+                    # get_backend_and_container throws a ValueError when it cannot find anything
+                    _, scan_obj = beu.get_backend_and_container_obj(scan_target)
```
giuseppe: could we rewrite it as `_, scan_list = beu.get_backend_and_container_obj(scan_target)`

and move the second try/except block to the same level as the first one?  The only other change required is:
```
if scan_list != []:
   _, scan_list = beu.get_backend_and_image_obj(scan_target)
```

What do you think?

Atomic/scan.py
baude: Missed that, of course.

giuseppe:
@rh-atomic-bot r+ dc6def0

PR #920, opened 2017-02-28
Atomic/util.py Fix Try/Catch exceptions for dockerd

baude:
Fix a cirucular dependancy in atomic storage reset where certain
functions require a dockerd but storage reset requires dockerd to
be stopped.


baude:
Merging for @shishir-a412ed because s3 is broked.

PR #921, opened 2017-02-28
Atomic/storage.py: Reduce number of times NoDockerDaemon is called

baude:
Instead of calling the default_docker_lib function a bunch of times
in the arg parser, we call it one and use the result in the parser.

Also, tighten up some exit conditions where self.d.close() is being
called when dockerd is not running.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -53,11 +53,9 @@ input, is_python2 = check_if_python2() # pylint: disable=redefined-builtin
 
 
 def get_docker_conf():
+    dconf = []
     with AtomicDocker() as c:
-        try:
-            dconf = c.info()
-        except (NoDockerDaemon, requests.exceptions.ConnectionError):
-            raise ValueError("This Atomic function requires an active docker daemon.")
+        dconf = c.info()
     return dconf
 
 def get_registries():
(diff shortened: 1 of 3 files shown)


Atomic/usersys.py
```diff
@@ -0,0 +1,11 @@
+from Atomic.util import output_json
+from Atomic.discovery import RegistryInspect
+#atomic-registry.usersys.redhat.com:5000/auth/baude_auth
```
TomSweeneyRedHat: Do you want this and the other commented lines in here?

Atomic/util.py
```diff
@@ -447,7 +445,7 @@ def default_docker():
 def default_docker_lib():
     try:
         return get_docker_conf()["DockerRootDir"]
-    except ValueError:
+    except (NoDockerDaemon, requests.ConnectionError):
```
TomSweeneyRedHat: Any value to adding a warning here with verbiage similar to the error at the old line 60?

Atomic/usersys.py
rhatdan: Yes remove this line

Atomic/util.py
rhatdan: We want to make atomic agnostic to container runtime.  Do not want it complaining about docker not running.

Atomic/usersys.py
baude: This whole file is just a test file.  Wasn't meant to be in the commit.  Removing...

baude:
@rhatdan should be good to go now.

rhatdan:
@rh-atomic-bot r+ 6c48d5e

PR #922, opened 2017-03-01
syscontainers: when using --rootfs create a symlink to rootfs

giuseppe:
In this way it is still possible to access $DESTDIR/rootfs as in the
case --rootfs is not specified.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -505,6 +505,9 @@ class SystemContainers(object):
                 config['root']['path'] = remote_rootfs
             with open(destination_path, 'w') as config_file:
                 config_file.write(json.dumps(config, indent=4))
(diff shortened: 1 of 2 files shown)


giuseppe:
yuqi-zhang a simple change for the --rootfs case.  What do you think?

rhatdan:
LGTM

yuqi-zhang:
LGTM, will merge when tests pass

cgwalters:
Note you actually don't need to wait for the per-PR tests - [our instance of Homu](https://homu-projectatomic-ci.svc.ci.openshift.org/) reruns them for the merge.

This is a nice advantage of Homu - you can fearlessly drop an `r+` in a comment and know the tests have to pass.

yuqi-zhang:
Cool,

@rh-atomic-bot r+ 23e417d

PR #923, opened 2017-03-01
Fix filtering

yuqi-zhang:
Change the format of "Created" for images to match containers,
fix filtering for partial keywords, and update auto-complete.


Atomic/images.py
```diff
@@ -259,7 +259,8 @@ def _filter_include_image(self, image_obj):
                                  "Please choose from {}".format(var, filterables))
             if var == "type":
                 var = "str_backend"
-            if getattr(image_obj, var, None).lower() != value.lower():
+
+            if value.lower() not in str(getattr(image_obj, var, None)).lower():
```
giuseppe: since we are touching this line, I wonder about the missing case here.  Do we really want to compare the value to the string "none"?

Should we do something like this instead?

`hasattr(image_obj, var) and value.lower() in image_obj.var.lower()`

rhatdan: That is closer to the original.

rhatdan:
@yuqi-zhang could you take this oppurtunity to fix the man pages and bash completions to list the valid values for filtering?

Atomic/images.py
yuqi-zhang: Yes, good point, I will fix that. Although I would like to note that for all the filterable items, if the user specifies a filter that does not exist, it would get caught and an error would be returned. So for all filters that get here, the image_obj would have already initialized that attribute on creation (which incidentally defaults to None). Thus hasattr would always be true.

yuqi-zhang:
Will do

yuqi-zhang:
Bash and man pages have been updated. could you PTAL @giuseppe @rhatdan

giuseppe:
LGTM, I wait for @rhatdan since he had some comments

rhatdan:
@rh-atomic-bot r+ d173f73

giuseppe:
@rh-atomic-bot r+ d173f73

yuqi-zhang:
Can you take a look at what's happening there ^ @jlebon

giuseppe:
@yuqi-zhang, could you try with a rebase from master and a `push -f`?  We can try with a fresh checksum :)

jlebon:
Yeah, I think this is just a case of homu being confused. Will check the logs to see what happened there. But for now yeah, just a rebase or a `commit --amend` should do the trick.

jlebon:
Don't mind me, just triggering the error again.
@rh-atomic-bot r+ d173f73

yuqi-zhang:
Rebased and squashed.

giuseppe:
@rh-atomic-bot r+ 7e8d46f

PR #927, opened 2017-03-03
syscontainers: do not resolve the image name if Docker is down

giuseppe:
there are 2 small improvements here:

1) if the image cannot be resolved, use the provided image name.  It won't cause an error when Docker is down.
2) Print a warning if the image is not ready to be used as a system container, suggest the user to manually modify the generated config.json file before trying to run the service.


rhatdan:
@rh-atomic-bot r+ 0f1ef19

Issue #929, opened 2017-03-04
(atomic sign) does not work at all


mtrmac:
1. ```
   Traceback (most recent call last):
     File "./atomic", line 189, in <module>
       sys.exit(_func())
     File "/home/mitr/t/atomic/Atomic/sign.py", line 80, in sign
       manifest = ri.get_manifest()
     File "/home/mitr/t/atomic/Atomic/discovery.py", line 41, in get_manifest
       assert(self.fqdn is not None)
   AssertionError
   ```
   Apparently it is necessary to call `ri.inspect()` before `ri.get_manifest()`; though a clean fix would probably be different, at the very least it would be nice to skip `util.skopeo_inspect` if `self.registry` is specified in the input argument.

1. After adding a `_ = ri.inspect()` in there:
   ```
   $ ATOMIC_CONF=tests/unit/fixtures/atomic.conf ATOMIC_CONFD=/var/empty/ sudo -E ./atomic --debug sign --sign-by=mitr@volny.cz -d . docker.io/library/busybox:latest
   …
   Traceback (most recent call last):
     File "./atomic", line 189, in <module>
       sys.exit(_func())
     File "/home/mitr/t/atomic/Atomic/sign.py", line 85, in sign
       manifest_file.write(manifest)
   TypeError: argument 1 must be string or buffer, not dict
   ```
   It seems `RegistryInspect.get_manifest` should use `return_json=False`, not `True`.

1. Also (after the `_ = ri.inspect()` workaround), only with `--debug`:
   ```
   $ ATOMIC_CONF=tests/unit/fixtures/atomic.conf ATOMIC_CONFD=/var/empty/ sudo -E ./atomic --debug sign --sign-by=mitr@volny.cz -d . busybox:latest
   …
   Traceback (most recent call last):
     File "./atomic", line 189, in <module>
       sys.exit(_func())
     File "/home/mitr/t/atomic/Atomic/sign.py", line 79, in sign
       _ = ri.inspect()
     File "/home/mitr/t/atomic/Atomic/discovery.py", line 34, in inspect
       self.fqdn = self.find_image_on_registry()
     File "/home/mitr/t/atomic/Atomic/discovery.py", line 70, in find_image_on_registry
       util.write_out(i)
     File "/home/mitr/t/atomic/Atomic/util.py", line 189, in write_out
       _output(sys.stdout, output, lf)
     File "/home/mitr/t/atomic/Atomic/util.py", line 202, in _output
       fd.write(output + lf)
[...]

PR #932, opened 2017-03-06
Atomic/discovery.py: Fix exception with FQDN in atomic sign

baude:
When running atomic sign, we were throwing an exception because
the fqdn has not been determined yet.  Made fqdn a property instead
of a variable.  The property will define itself.

### DIFF
--- a/Atomic/discovery.py
+++ b/Atomic/discovery.py
@@ -29,7 +29,7 @@ class RegistryInspect():
     @property
     def fqdn(self):
         if not self._fqdn:
-            self.fqdn = self.assemble_fqdn(include_tag=True) if self.registry else self.find_image_on_registry()
+            self._fqdn = self.assemble_fqdn(include_tag=True) if self.registry else self.find_image_on_registry()
         return self._fqdn
 
     @fqdn.setter
@@ -37,8 +37,6 @@ class RegistryInspect():
         self._fqdn = value
 
     def inspect(self):
-        if not self.fqdn:
-            _ = self.fqdn
         if self.registry:
             inspect_data = util.skopeo_inspect("docker://{}".format(self.fqdn), return_json=True)
         else:
--- a/Atomic/discovery.py
+++ b/Atomic/discovery.py
@@ -21,17 +21,27 @@ class RegistryInspect():
         self.digest = digest
         self.orig_input = orig_input
         self._remote_inspect = None
-        self.fqdn = None
+        self._fqdn = None
 
         if self.debug:
             util.output_json(self.registries)
 
+    @property
+    def fqdn(self):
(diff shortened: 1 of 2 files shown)


baude:
should address #1 in https://github.com/projectatomic/atomic/issues/929

Atomic/discovery.py
```diff
@@ -21,17 +21,26 @@ def __init__(self, registry=None, repo=None, image=None, tag=None, digest=None,
...
+    def fqdn(self, value):
+        self._fqdn = value
+
     def inspect(self):
         if self.registry:
-            self.fqdn = self.assemble_fqdn(include_tag=True)
             inspect_data = util.skopeo_inspect("docker://{}".format(self.fqdn), return_json=True)
         else:
-            self.fqdn = self.find_image_on_registry()
+            _ = self.fqdn
```
rhatdan: Is this line necessary?

baude: theoretically yes because it sets the fqdn (the old code did this) and I didn't want to break a possible case I hadn't identified.  More of being safe than anything.

Atomic/discovery.py
```diff
@@ -21,17 +21,26 @@ def __init__(self, registry=None, repo=None, image=None, tag=None, digest=None,
...
-        self.fqdn = None
+        self._fqdn = None
 
         if self.debug:
             util.output_json(self.registries)
 
+    @property
+    def fqdn(self):
+        if not self._fqdn:
+            self.fqdn = self.assemble_fqdn(include_tag=True) if self.registry else self.find_image_on_registry()
```
TomSweeneyRedHat: Maybe I'm misunderstanding, but should this be "self._fqdn = self.assemble_fqdn" here now?

baude: it could be or it can be as is given I have a property setter.

baude:
@rhatdan please merge

Atomic/discovery.py
```diff
@@ -21,17 +21,27 @@ def __init__(self, registry=None, repo=None, image=None, tag=None, digest=None,
...
+        if not self._fqdn:
+            self._fqdn = self.assemble_fqdn(include_tag=True) if self.registry else self.find_image_on_registry()
+        return self._fqdn
+
+    @fqdn.setter
+    def fqdn(self, value):
+        self._fqdn = value
+
     def inspect(self):
+        if not self.fqdn:
```
rhatdan: These two lines seem to be useless, since the first call will populate the self.fqdn?

I think you should just delete them and everything will be taken out.

baude: done

rhatdan:
@rh-atomic-bot r+ 2d947c4

PR #933, opened 2017-03-07
Small fix to image info

yuqi-zhang:
Use default storage if user does not specify, so it doesn't output
"Unable to associate string 'None' with backend".


baude:
lgtm

baude:
@rh-atomic-bot r+ 64341da

PR #936, opened 2017-03-13
Revert #915: Run container-storage-setup on atomic storage modify.

shishir-a412ed:

### DIFF
--- a/tests/integration/test_storage.sh
+++ b/tests/integration/test_storage.sh
@@ -69,18 +69,15 @@ trap teardown EXIT
 # If /etc/sysconfig/docker-storage-setup is missing, atomic should create the file.
 
 rm -f /etc/sysconfig/docker-storage-setup
-OUTPUT=$(${ATOMIC} storage modify --add-device $TEST_DEV_1 2>&1)
-grep -q "^DEVS=\"$TEST_DEV_1\"$" /etc/sysconfig/docker-storage-setup
 
 if [ -n "$VGROUP" ]; then
     cat >>/etc/sysconfig/docker-storage-setup <<EOF
 MIN_DATA_SIZE=0G
 EOF
+    # Add a device to volume group backing root filesystem.
 
-    # Executing docker-storage-setup should add the device
-    # and put it into the volume group.
-
-    /usr/bin/docker-storage-setup 1>/dev/null||exit 1
+    OUTPUT=$(${ATOMIC} storage modify --add-device $TEST_DEV_1 2>&1)
+    grep -q "^DEVS=\"$TEST_DEV_1\"$" /etc/sysconfig/docker-storage-setup
     [ $(pvs --noheadings -o vg_name $TEST_DEV_1_pvs) == $VGROUP ]
 
     # Removing it should undo all that.
@@ -92,7 +89,6 @@ EOF
     # Adding it again straight away should work.
 
     ${ATOMIC} storage modify --add-device $TEST_DEV_1
-    /usr/bin/docker-storage-setup 1>/dev/null||exit 1
     [ $(pvs --noheadings -o vg_name $TEST_DEV_1_pvs) == $VGROUP ]
     grep -q "^DEVS=\"$TEST_DEV_1\"$" /etc/sysconfig/docker-storage-setup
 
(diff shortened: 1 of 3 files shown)


shishir-a412ed:
ping @rhatdan 

/cc @mvollmer 
PR for this: https://github.com/projectatomic/atomic/pull/915#issuecomment-285787316

Shishir

rhatdan:
LGTM

shishir-a412ed:
@rhatdan Integration test `test_dbus.py` is failing, however the error doesn't seem like related to this patch. WDYT ?

/cc @baude @jlebon

rhatdan:
I see the atomic_top.py test failing, not dbus?

shishir-a412ed:
@rhatdan Previously it was `test_dbus.py`. I did a second run (restarted the CI) and now it is `atomic_top.py`. The other error in `Fedora/25/Cloud` seems legit as it is failing in `test_storage.sh`.
Taking a look into that one.

Shishir

tests/integration/test_storage.sh
```diff
@@ -69,20 +69,20 @@ trap teardown EXIT
 # If /etc/sysconfig/docker-storage-setup is missing, atomic should create the file.
 
 rm -f /etc/sysconfig/docker-storage-setup
-OUTPUT=$(${ATOMIC} storage modify --add-device $TEST_DEV_1 2>&1)
+if [ -z "$VGROUP" ];then
+   VGROUP="test-atomic-storage"
+   OUTPUT=$(${ATOMIC} storage modify --add-device $TEST_DEV_1 --vgroup $VGROUP 2>&1)
```
mvollmer: This is not strictly part of reverting #915, right?  I think setting the vgroup requires a reset first.

shishir-a412ed: "I think setting the vgroup requires a reset first."

I am still fixing the tests, however I don't see the relevance.
The volume group (VG) is being set here first time. why would it require a `reset` first ?

mvollmer: > I am still fixing the tests

Why did you change the tests as part of this PR beyond just reverting #915?  Can we leave that out and worry about it in a separate PR?

mvollmer: > The volume group (VG) is being set here first time. why would it require a reset first ?

The error from docker-storage-setup is
```
ERROR: Docker has been previously configured for use with devicemapper graph driver. Not creating a new thin pool as existing docker metadata will fail to work with it. Manual cleanup is required before this will succeed.
INFO: Docker state can be reset by stopping docker and by removing /var/lib/docker directory. This will destroy existing docker images and containers and all the docker metadata.
```

This is what `atomic storage reset` will do.

The reset is necesarry because there is already a docker storage pool, and docker-storage-setup can not move it seamlessly into a different volume group.

Whe I initially wrote the tests I couldn't do any reset since the tests couldn't be destructive back then.  Can we do it now, though?  The reset will remove the atomic-test-* images which are used by other integration tests.

shishir-a412ed: `Why did you change the tests as part of this PR beyond just reverting #915? Can we leave that out and worry about it in a separate PR?`

Because the tests started failing. Was this environment (fedora/25/cloud) recently introduced in the CI? Not sure how the existing test would pass on this environment. Since the test will try to find the volume group backing the rootfilesystem, and if it's not present it will look for `VG=` which was not defined in the exisiting tests. So d-s-s would fail. Obviously just setting VG won't work since docker is already set to run on loopback devices. We need to clear that first.

I believe `atomic storage reset` could come handy here like you mentioned.

shishir-a412ed:
For now, I have reverted the test back to how it is currently in `atomic/master` (This is not necessarily a good thing since the current test doesn't test a lot of possible storage configurations. e.g what happens when VG backing rootfilesystem is not present. But I will deal with that in a separate PR). For now this patch revert #915 thereby maintaining backward compatibility as requested by @mvollmer .

@rhatdan PTAL.

tests/integration/test_storage.sh
shishir-a412ed: I misread the tests. If the VG backing root filesystem is not present, the tests are skipped.
Hence it was not failing.

rhatdan:
@rh-atomic-bot r+ c807356

jlebon:
@rh-atomic-bot r=rhatdan c807356

rhatdan:
Is this a flake?

@rh-atomic-bot r=rhatdan c807356

rhatdan:
@rh-atomic-bot retest

jlebon:
@rh-atomic-bot retry

PR #938, opened 2017-03-14
Run cmd

baude:


Atomic/objects/container.py
```diff
@@ -100,3 +100,7 @@ def name(self, value):
     @property
     def user_command(self):
         return self._user_command
+
+    @user_command.setter
+    def user_command(self, value):
+        self._user_command = value
```
rhatdan: Add Return

rhatdan:
Just a nit and LGTM

baude:
updated per comments.

rhatdan:
@rh-atomic-bot r+ ea5368e

PR #942, opened 2017-03-20
Atomic/mount.py: Allow different binary paths

baude:
Different distributions put their binaries in different paths.  So for things
like 'mount', we need to be able to account for '/usr/bin/mount' or '/bin/mount'
and the like.

### DIFF
--- a/Atomic/mount.py
+++ b/Atomic/mount.py
@@ -36,10 +36,16 @@ from gi.repository import GLib  # pylint: disable=no-name-in-module
 
 # Module for mounting and unmounting containerized applications.
 
-DMSETUP_PATH='/usr/sbin/dmsetup'
-LSBLK_PATH='/usr/bin/lsblk'
-MOUNT_PATH='/usr/bin/mount'
-FINDMNT_PATH='/usr/bin/findmnt'
+def path_exists(paths):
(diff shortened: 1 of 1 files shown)


rhatdan:
@rh-atomic-bot r+ f308cbd

Atomic/mount.py
```diff
@@ -36,10 +36,16 @@
...
 
-DMSETUP_PATH='/usr/sbin/dmsetup'
-LSBLK_PATH='/usr/bin/lsblk'
-MOUNT_PATH='/usr/bin/mount'
-FINDMNT_PATH='/usr/bin/findmnt'
+def path_exists(paths):
+    for path in paths:
+        if os.path.exists(path):
+            return path
+    raise ValueError("Unable to find command in {}".format(paths))
```
TomSweeneyRedHat: Should this be "Unable to find path in {}", instead of 'find command'?

rhatdan: This tool is looking for commands, function should probably be renamed to commands_exists()
Which would clear it up.  Bottom line is the /usr/bin/mount or /bin/mount does not exist.

baude:
@rh-atomic-bot retry

PR #944, opened 2017-03-21
syscontainers: do not delete previous deployment on update

giuseppe:
Only overwrite files coming from the new version.  This permits the
Docker container to support upgrades of the container as the previous
running containers shim processes don't lose their mount points.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -292,7 +292,7 @@ class SystemContainers(object):
             return self._do_checkout(repo, name, img, upgrade, values, destination, unitfileout, tmpfilesout, extract_only, remote)
         except (ValueError, OSError) as e:
             try:
-                if not upgrade and not extract_only:
+                if not extract_only and not upgrade:
                     shutil.rmtree(destination)
             except OSError:
                 pass
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -292,7 +292,7 @@ class SystemContainers(object):
             return self._do_checkout(repo, name, img, upgrade, values, destination, unitfileout, tmpfilesout, extract_only, remote)
         except (ValueError, OSError) as e:
             try:
-                if not extract_only:
+                if not upgrade and not extract_only:
                     shutil.rmtree(destination)
             except OSError:
                 pass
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@baude PTAL

Atomic/syscontainers.py
```diff
@@ -292,7 +292,7 @@ def _checkout(self, repo, name, img, deployment, upgrade, values=None, destinati
             return self._do_checkout(repo, name, img, upgrade, values, destination, unitfileout, tmpfilesout, extract_only, remote)
         except (ValueError, OSError) as e:
             try:
-                if not extract_only:
+                if not upgrade and not extract_only:
```
TomSweeneyRedHat: nit, just to remain consistent with the lines below, should this be "if not extract_only and not upgrade:"

Atomic/syscontainers.py
```diff
@@ -407,13 +407,12 @@ def _do_checkout(self, repo, name, img, upgrade, values, destination, unitfileou
                     pass
             rootfs = os.path.join(destination, "rootfs")
 
-        if os.path.exists(destination):
-            shutil.rmtree(destination)
-
```
TomSweeneyRedHat: Should we keep this rmtree  logic but hide it behind an if not upgrade?

if os.path.exists(destination) and not upgrade:
    shutil.rmtree(destination)

giuseppe: this was used only with `upgrade`.  `install` would fail earlier if the destination is not empty:

```
        if self.get_checkout(name):
            util.write_out("%s already present" % (name))
            return
```

Atomic/syscontainers.py
giuseppe: sure.  I pushed a fixup patch to address it.

giuseppe:
@rh-atomic-bot r=TomSweeneyRedHat

@TomSweeneyRedHat now you can approve the patch, if you are not already familiar with rh-atomic-bot:

https://github.com/jlebon/redhat-ci/blob/master/README.md

The `r+ <commit sha>` is what you'll need

giuseppe:
@rh-atomic-bot r-

rhatdan:
@rh-atomic-bot r+  	611306c

PR #945, opened 2017-03-21
Fix outdated container image message (BZ 1434430)

baude:
When running an image with atomic run, if the container already exists,
we were accidently displaying a message stating that the container's
image was old.  This was due to an invalid comparison between the container's
image id and the image's id.  The comparison was failing because we were
comparing two different variables and not the ids.

This was reported in bugzilla https://bugzilla.redhat.com/show_bug.cgi?id=1434430

### DIFF
--- a/Atomic/objects/container.py
+++ b/Atomic/objects/container.py
@@ -14,7 +14,7 @@ class Container(object):
         self.deep = False
         self._backend = backend
         self.runtime = backend.backend
-        self.image_id = None
+        self.image = None
         self.image_name = None
         self._command = None
         self.state = None
@@ -104,3 +104,4 @@ class Container(object):
     @user_command.setter
     def user_command(self, value):
         self._user_command = value
+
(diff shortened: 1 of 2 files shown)


Atomic/objects/container.py
```diff
@@ -104,3 +104,7 @@ def user_command(self):
     @user_command.setter
     def user_command(self, value):
         self._user_command = value
+
+    @property
+    def image(self):
+        return self.image_id
```
rhatdan: Missing Return, otherwise LGTM
Do we need to get this fix into RHEL build today?

rhatdan:
test_dbus is failing with no image_id defined.

rhatdan:
@rh-atomic-bot r+  	b463821

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot r+ b463821

PR #947, opened 2017-03-21
Fix environment variable not being displayed

yuqi-zhang:
'atomic images info' on a system container image was not displaying
the environment variables in the container anymore. This was due to
get_backend_and_image_obj() function decomposing the img_obj.image
to not be the full image name. img_obj.name should be used instead.


rhatdan:
Makes Sense to me.

@giuseppe @baude PTAL

giuseppe:
@rh-atomic-bot r+ 96e32b6

giuseppe:
@rh-atomic-bot retry

PR #949, opened 2017-03-23
Fixes for system containers generate RPM

giuseppe:
Add some tests for the generate RPM feature of system containers and fix some issues that I've found while using these tests.

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -62,3 +62,4 @@ context: fedora/25/cloud
 packages:
   - atomic
   - python3-coverage
+  - rpm-build
--- a/Atomic/backends/_ostree.py
+++ b/Atomic/backends/_ostree.py
@@ -37,13 +37,12 @@ class OSTreeBackend(Backend):
         container.id = container_id
         container.runtime = info['Runtime']
         container.image_name = info['Image']
-        container.image_id = info['ImageID']
+        container.image = info['ImageID']
         container.created = info['Created']
         container.status = container.state = runtime['status']
         container.input_name = container_id
         container.original_structure = info
         container.deep = True
-        container.image = info['Image']
         container.running = False if container.status == 'inactive' else True
         return container
 
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -203,7 +203,7 @@ class Containers(Atomic):
         containers = []
         for con_obj in con_objects:
             _con = {'id': con_obj.id,
-                    'image_id': con_obj.image_id,
+                    'image_id': con_obj.image,
                     'image_name': con_obj.image_name,
                     'command': con_obj.command,
                     'created': con_obj.created,
--- a/Atomic/rpm_host_install.py
+++ b/Atomic/rpm_host_install.py
@@ -1,0 +1,195 @@
+import os
+from . import util
+import tempfile
+import shutil
+
+RPM_NAME_PREFIX = "atomic-container"
+
+class RPMHostInstall(object):
+
+    @staticmethod
+    def rm_add_files_to_host(old_installed_files, exports, prefix="/", files_template=None, values=None, rename_files=None):
+        # if any file was installed on the host delete it
+        if old_installed_files:
+            for i in old_installed_files:
+                try:
+                    os.remove(i)
+                except OSError:
+                    pass
+
+        if not exports:
+            return []
+
+        templates_set = set(files_template or [])
+
+        # if there is a directory hostfs/ under exports, copy these files to the host file system.
+        hostfs = os.path.join(exports, "hostfs")
+        new_installed_files = []
+        if os.path.exists(hostfs):
+            for root, _, files in os.walk(hostfs):
+                rel_root_path = os.path.relpath(root, hostfs)
+                if not os.path.exists(os.path.join(prefix, rel_root_path)):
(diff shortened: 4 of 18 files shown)


giuseppe:
bot, retest this please

rhatdan:
Great work.  LGTM

giuseppe:
I am adding some more cleanups @TomasTomecek would like to use the same feature for Docker containers, so I am moving the RPM generator in its own file.

giuseppe:
@TomasTomecek I have added some more patches to move the code to a new file.

It is not the nicest API, but it is a starting point from what I refactored out from syscontainers.py.  We can clean it up incrementally with your feedbacks.

The entrypoint is `generate_rpm`.  For the Docker part, you won't probably need the arguments that have a default value (values, installed_files, installed_files_template, rename_files).

Beside that, it makes sense to share the same structure `/exports` and `/exports/hostfs` so to keep the possibility of having the same image for Docker and for system containers.

TomasTomecek:
@giuseppe Thank you!

> Beside that, it makes sense to share the same structure /exports and /exports/hostfs so to keep the possibility of having the same image for Docker and for system containers.

Absolutely. I want to reuse as much code as possible.

Unfortunately there is some further refactoring required: if I want to use functionality from `mount.py` inside `rpm_host_install.py`, the code halts on import loop:

```
  File "./atomic", line 33, in <module>
    import Atomic
  File "Atomic/__init__.py", line 3, in <module>
    from .atomic import Atomic
  File "Atomic/atomic.py", line 6, in <module>
    from .syscontainers import SystemContainers
  File "Atomic/syscontainers.py", line 17, in <module>
    from .rpm_host_install import RPMHostInstall, RPM_NAME_PREFIX
  File "Atomic/rpm_host_install.py", line 5, in <module>
    from . import mount
  File "Atomic/mount.py", line 23, in <module>
    from . import Atomic
ImportError: cannot import name Atomic
```

I'm not sure how to refactor it -- likely `syscontainers` import from `atomic.py` needs to be taken out.

giuseppe:
@TomasTomecek should the Docker code be added somewhere else?  I think `rpm_host_install.py` should not be an user of `mount` but that we add the should be used only for the part in common between OSTree and Docker.
As I see it, we need to add the `--system-package` option for Docker (which differently from OSTree, I think it should default to no, as it is expensive to mount the image each time a container is installed and probably only few containers will have `/exports`, or even better we add a LABEL used both by Docker and system containers so users don't need to be aware of these details), then from the Docker backend we mount the image and once it is mounted the functions from `rpm_host_install.py` are used.

Do you think it makes sense?

giuseppe:
@TomasTomecek I pushed a small fixup patch that enables to do this:

```diff
diff --git a/Atomic/install.py b/Atomic/install.py
index 855f137..8527b51 100644
--- a/Atomic/install.py
+++ b/Atomic/install.py
@@ -4,6 +4,11 @@ from . import util
 from .util import add_opt
 from .syscontainers import OSTREE_PRESENT
 from Atomic.backendutils import BackendUtils
+import os
+import shutil
+from .mount import DockerMount
+from .rpm_host_install import RPMHostInstall, RPM_NAME_PREFIX
+import tempfile
 
 try:
     from . import Atomic
@@ -121,6 +126,23 @@ class Install(Atomic):
                 return self.syscontainers.install(self.image, self.name)
             be.pull_image(self.args.image, remote_image_obj, debug=self.args.debug)
             img_obj = be.has_image(self.image)
+
+
+        if self.args.system_package in ['build', 'yes']:
+            temp_dir = tempfile.mkdtemp()
+            os.makedirs(os.path.join(temp_dir, "root"))
+            m = DockerMount(os.path.join(temp_dir, "root"))
+            m.mount(self.image)
+            try:
+                origin_name, path = RPMHostInstall.generate_rpm("test", self.image, img_obj.labels, os.path.join(temp_dir, "root/rootfs/exports"), temp_dir)
+                shutil.move(path, origin_name)
+            finally:
+                m.unmount()
+                shutil.rmtree(temp_dir)
+            sys.exit(0)
+
         install_args = img_obj.get_label('INSTALL')
         if not install_args:
             return 0
```

Of course, this is just a proof of concept and cannot be used as it is (also should not probably stay here in install.py).  Just wanted to show what we can do with the new code in common.

giuseppe:
There are a few fixes that are nice to have as soon as possible and would be nice to have before https://github.com/projectatomic/atomic/pull/952.  Is this ready to 🚢?

TomasTomecek:
@giuseppe Thank you, that diff was very helpful! I opened https://github.com/projectatomic/atomic/pull/955, feel free to take a look once you have time.

giuseppe:
@TomasTomecek thanks for the feedback.  I saw the `/etc` related fix, that is indeed a good thing to have.  Though, let's keep working on the new PR you have opened, and if there are no blockers let's merge this PR.  It will also help @ashcrow and me for the oneshot feature in system containers

baude:
looks like we need a rebase and to pass tests still ?

giuseppe:
I rebased and now I get some weird errors, taking a look...

giuseppe:
@baude tests are passing again

baude:
lgtm

baude:
@rh-atomic-bot r+  	a22fda4

PR #950, opened 2017-03-23
Atomic/install.py: Record installs for later use

baude:
When installing an image, we now write a small bit of json
to /var/lib/atomic/install.json.  The json format is:

{
	<image_name>: {
			     id: <image_id>,
			     install_date: <install_date_in_utc
			 }
}

This will be used in update, run, etc to ensure that any image
with an INSTALL label is first installed.

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -21,6 +21,9 @@ import requests
 import ipaddress
 import socket
 from Atomic.backends._docker_errors import NoDockerDaemon
+import fcntl
+import time
+
 # Atomic Utility Module
 
 ReturnTuple = collections.namedtuple('ReturnTuple',
@@ -29,6 +32,9 @@ ATOMIC_CONF = os.environ.get('ATOMIC_CONF', '/etc/atomic.conf')
 ATOMIC_CONFD = os.environ.get('ATOMIC_CONFD', '/etc/atomic.d/')
 ATOMIC_LIBEXEC = os.environ.get('ATOMIC_LIBEXEC', '/usr/libexec/atomic')
 ATOMIC_VAR_LIB = os.environ.get('ATOMIC_VAR_LIB', '/var/lib/atomic')
+if not os.path.exists(ATOMIC_VAR_LIB):
+    os.makedirs(ATOMIC_VAR_LIB)
+ATOMIC_INSTALL_JSON = os.environ.get('ATOMIC_INSTALL_JSON', os.path.join(ATOMIC_VAR_LIB, 'install.json'))
 
 GOMTREE_PATH = "/usr/bin/gomtree"
 BWRAP_OCI_PATH = "/usr/bin/bwrap-oci"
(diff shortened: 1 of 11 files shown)


Atomic/util.py
```diff
@@ -794,6 +795,54 @@ def load_scan_result_file(file_name):
...
+    @staticmethod
+    def read_install_data():
+        if os.path.exists(ATOMIC_INSTALL_JSON):
+            return load_scan_result_file(ATOMIC_INSTALL_JSON)
+        return {}
+
+    @staticmethod
+    def write_install_data(install_data):
+        with open(ATOMIC_INSTALL_JSON, 'w') as write_data:
+            json.dump(install_data, write_data)
```
giuseppe: this can be risky.  If something goes wrong in json.dump() we lose the entire db of the images.  We should probably write to a temporary file in the same directory first, and then move it to ATOMIC_INSTALL_JSON

rhatdan: Yes I agree create a tmpfile and rename.

rhatdan: Should we use file locks?  To make sure that two processes accessing the file at the same time don't overight?  Or should we create a different file for each installed image?

atomic
```diff
@@ -124,6 +124,8 @@ def create_parser(help_text):
                         help=_("show atomic version and exit"))
     parser.add_argument('--debug', default=False, action='store_true',
                         help=_("show debug messages"))
+    parser.add_argument('-i', '--ignore', default=False, action='store_true',
+                        help=_("allows you to ignore certain errors"))
```
rhatdan: This is pretty vague.  Either specify what  you are ignoring, or don't give the user the option.

Atomic/util.py
baude: I'll implement a file locking mechanism on write.

baude:
updated with a file locking mechanism

Atomic/util.py
```diff
@@ -794,6 +798,67 @@ def load_scan_result_file(file_name):
...
+    def read_install_data():
+        if os.path.exists(ATOMIC_INSTALL_JSON):
+            return json.loads(open(os.path.join(ATOMIC_INSTALL_JSON), "r").read())
+        return {}
+
+    @staticmethod
+    def write_install_data(install_data):
+        with open(ATOMIC_INSTALL_JSON, 'w') as write_data:
+            time_out = 0
+            f_lock = False
```
giuseppe: should the lock protect the entire (read + modify + write) operation?  Otherwise with two processes trying to write at the same time, we can lose what was written by the process that terminated first if they read the same initial file content.
An alternative can probably be to keep one separate file for each container, like:

`ATOMIC_VAR_LIB/install-data/${NAME}.json`

so we don't require any sort of synchronization.  Would it work?

Prune can cleanup the files that are left there without a related container still active (if the user manually uninstalled it, or the previous rm failed for any reason).

rhatdan: I kind of like the separate file per container.

baude: The problem with separate files is it makes it significantly more complicated to look up images by hash, etc.  And eventually we come down to the same problem of reading or writing the same file.  I'll extend the locking.

baude:
@rhatdan implemented locking with @giuseppe and also addressed the ignore comment you made.  Ready for merging.

baude:
[Atomic CLI: Warn/Flag when user runs an image that hasn't been installed.](https://trello.com/c/4BENk3NT/348-atomic-cli-warn-flag-when-user-runs-an-image-that-hasn-t-been-installed)

Atomic/util.py
```diff
@@ -794,6 +800,117 @@ def load_scan_result_file(file_name):
...
+        if not f_lock:
+            raise ValueError("Unable to get file lock for {}".format(ATOMIC_INSTALL_JSON))
+
+    def release_lock():
+        fcntl.flock(install_data_file, fcntl.LOCK_UN)
+
+    def wrapper(*args, **kwargs):
+        get_lock()
+        ret = func(*args, **kwargs)
+        release_lock()
```
rhatdan: Should we do a try: finally block incase func raises an exception?

rhatdan:
LGTM Make sure @giuseppe Likes it.

giuseppe:
the decorator looks very neat!

@rh-atomic-bot r+ 831a9a0

PR #952, opened 2017-03-24
Allow systemcontainers to install, run, and uninstall in one go

ashcrow:
This change adds a new optional label called ``atomic.run``. When set to ``once`` the image will be extracted, run, and cleaned up. This currently requires the use of ``--system``.

The idea behind this new option is that there are cases where a system container would not be a full fledged service on a host. For example, the openshift-ansible installer isn't a service, but a containerized installer.

Example:

```
$ sudo atomic pull --storage ostree docker:testcontainer:latest
$ sudo atomic install --system testcontainer:latest
Extracting to /ostree/repo/tmp/atomic-container/15926/rootfs
HI
$ ls /ostree/repo/tmp/atomic-container/15926/rootfs
ls: cannot access '/ostree/repo/tmp/atomic-container/15926/rootfs': No such file or directory
$
```

/cc @sdodson @tbielawa

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -227,6 +227,101 @@ class SystemContainers(object):
         return None
 
     def install(self, image, name):
+        """
+        External container install logic.
+
+        :param image: The name of the image
+        :type image: str
+        :param name: The name of the checkout
+        :type name: str
+        :returns: Shell call result
+        :rtype: int
+        """
+        return_value = None
+        # If we don't have a dockertar file or a reference to a docker engine image
+        if not image.startswith('dockertar:/') and not (image.startswith("docker:") and image.count(':') > 1):
+            labels = self.inspect_system_image(image).get('Labels', {})
+            # And we have a run-once label
+            if labels.get('atomic.run') == 'once':
+                # Execute the _run_once method and set the return_value
+                return_value = self._run_once(image, name)
+        # If we don't have a return_value then use the traditional install
+        if return_value is None:
+            return_value = self._install(image, name)
+        # Return
+        return return_value
+
+    def _run_once(self, image, name):
+        """
+        Runs the container once and then removes it.
+
+        :param image: The name of the image
+        :type image: str
+        :param name: The name of the checkout
+        :type name: str
+        :returns: Shell call result
+        :rtype: int
+        """
+        # Create a temporary directory to house the oneshot container
+        base_dir = os.path.join(self.get_ostree_repo_location(), "tmp/atomic-container", str(os.getpid()))
+        os.makedirs(base_dir)
+        try:
+            rootfs = os.path.sep.join([base_dir, 'rootfs'])
+            # Extract the image to a temp directory.
+            self.extract(image, rootfs)
+
+            # This part should be shared with install.
+            values = {}
+            if self.args.setvalues is not None:
+                setvalues = SystemContainers._split_set_args(self.args.setvalues)
+                for k, v in setvalues.items():
+                    values[k] = v
+
+            # Check for config.json in exports
+            destination_config = os.path.sep.join([base_dir, 'config.json'])
+            template_config_file = os.path.sep.join([rootfs, 'exports', 'config.json.template'])
+            template_tmpfiles = os.path.sep.join([rootfs, 'exports', 'tmpfiles.template'])
+            tmpfiles_destination = None
+            # If there is a config.json, use it
+            if os.path.exists(os.path.sep.join([rootfs, 'exports', 'config.json'])):
+                shutil.copy(os.path.sep.join([rootfs, 'exports', 'config.json']),
+                            destination_config)
+            # Else, if we have a template, populate it
+            elif os.path.exists(template_config_file):
+                with open(template_config_file, 'r') as infile:
+                    util.write_template(template_config_file, infile.read(), values, destination_config)
+            # Otherwise, use a default one
+            else:
+                self._generate_default_oci_configuration(destination_config)
+
+            # If we have a tmpfiles template, populate it
+            if os.path.exists(template_tmpfiles):
+                with open(template_tmpfiles, 'r') as infile:
+                    tmp = os.path.sep.join([base_dir, 'tmpfiles.conf'])
+                    util.write_template(template_tmpfiles, infile.read(), values, tmp)
+                    self._systemd_tmpfiles("--create", tmp)
+                    tmpfiles_destination = tmp
+
+            # Get the start command for the system container
+            (start_command, _) = self._generate_systemd_startstop_directives(name)
+            # Move to the base directory to start the system container
+            os.chdir(base_dir)
+            # ... and run it. We use call() because the actual
+            # run may be expected to fail.
+            return util.call(start_command)
+        finally:
+            if tmpfiles_destination:
+                try:
+                    self._systemd_tmpfiles("--remove", tmpfiles_destination)
+                except subprocess.CalledProcessError:
+                    pass
+            # Remove the temporary checkout
+            shutil.rmtree(base_dir)
+
+    def _install(self, image, name):
+        """
+        Internal container install logic.
+        """
         repo = self._get_ostree_repo()
         if not repo:
             raise ValueError("Cannot find a configured OSTree repo")
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -265,6 +265,7 @@ class SystemContainers(object):
         # Create a temporary directory to house the oneshot container
         base_dir = os.path.join(self.get_ostree_repo_location(), "tmp/atomic-container", str(os.getpid()))
         os.makedirs(base_dir)
+        tmpfiles_destination = None
         try:
             rootfs = os.path.sep.join([base_dir, 'rootfs'])
             # Extract the image to a temp directory.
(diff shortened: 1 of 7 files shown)


Atomic/syscontainers.py
```diff
@@ -210,6 +210,40 @@ def _find_rpm(tmp_dir):
...
+
+                # Get the start command for the system container
+                (start_command, _) = self._generate_systemd_startstop_directives(name)
+                # Move to the proper directory to start the system container
+                os.chdir(os.path.join(ATOMIC_VAR, name))
+                # ... and run it. We use call() because the actual run may be expected to fail.
+                util.call(start_command)
+            finally:
+                # Uninstall the image post run
+                self.uninstall(name)
```
giuseppe: when I firstly suggested this I forgot we had an `extract`.  Sorry :)

At some point we will need an OSTree temporary storage (also to fix mount) but for now I think it would be better if we can do something like:

```python
try:
    self.extract(img, tmpdir)
    generate_systemd_directives(...)
    os.chdir("$tmpdir/rootfs")
    util.call(...)
finally:
   shutil.rmtree(tmp)
```

`tmpdir` is better created under the OSTree repository so we are sure to use hard links and the checkout is faster.  The `_try_ostree_mount` function in mount.py does that (we will need to copy/move it to syscontainers).  Anyway, I'll have to fix this for the RPM stuff as well so feel free to ignore this part.

What do you think?

I see in your TODO list that there is to create the runc directives for --user.  That should be already done by generate_systemd_startstop_directives now to use bwrap-oci so there should not be anything to do for this.

ashcrow: > What do you think?

SGTM. I'll update this afternoon.

> I see in your TODO list that there is to create the runc directives for --user. That should be already done by generate_systemd_startstop_directives now to use bwrap-oci so there should not be anything to do for this.

Perfect!

rhatdan:
We keep making changes to system containers that force the user to have some kind of knowledge of how to install the image, can we remove the option and just put something like a label into the container to do a --oneshot?  Force the image developer to define how to install his image and allow the atomic command to react.  I want the same thing for --system.

If I do 
atomic install foobar

And foobar is intended to be a --oneshot --system container  I will be prone to forget those options.

ashcrow:
@rhatdan that's a really good point. If ``--system`` is going to make that change ``--oneshot`` should be done that was as well. @giuseppe if you're on board with this as well we can work on this together if you'd like.

giuseppe:
@ashcrow, sure! A container that is designed to be "oneshot" probably makes no sense at all to have it normally installed so a label that indicates this behaviour can make this easier for the user.

ashcrow:
@giuseppe I'll pull the option code out of this PR, take a look at how labels are used, and see if I can't put something together. I'll probably bug you a bit :laughing:

ashcrow:
I created an ``atomic.run`` label which, if used **and** set to ``once`` will extract, run, and clean up after the extract.

## Labels
```
LABEL "Name"="atomic-test-system"\
      "atomic.run"="once"\
      "atomic.type"="system"
```

## Run
```
$ sudo python atomic install --system o:latest
Extracting to /tmp/oIg9Ulv/rootfs
HI
$ ls /tmp/oIg9Ulv/rootfs
ls: cannot access '/tmp/oIg9Ulv/rootfs': No such file or directory
```

## Question

@giuseppe I noticed that using ``atomic.type`` as ``system`` didn't react the same was as using the ``--system`` option. Is this by design?

giuseppe:
@ashcrow Nice!

We should probably add a check for the type of container also for already pulled images, I guess you are pulling it directly from the local Docker engine?  We do that now for images pulled from a registry.

I've added a fixup commit here: https://github.com/giuseppe/atomic/tree/oneshot

I have not tested it, just wanted to share it quickly.

I've done some small changes, also added the possibility of generating the configuration from the config.json.template file and support for systemd tmpfiles, if the container needs to bind directories/files not present on the file system when it starts.

Some code is deliberately duplicated in the new function, as it will save some merge conflicts with https://github.com/projectatomic/atomic/pull/949.

I also like the new path for `_run_once` as in the future I'd like we treat these containers separately: we will have a "layers storage" that we can use for mounting images and at the same time for running one shot containers since they are managed by atomic, we can setup the overlayfs for mounting the layers, and proceed with the cleanup once it is done.

Another change is to use a tmpdirectory under the OSTree repository, so we use hard links instead of copying the files.  I think at this point all the features we need to support are there.

What do you think?

ashcrow:
@giuseppe I'll pull in and test.

yuqi-zhang:
@ashcrow with regards to "atomic.type=system" vs "install --system" there are some discrepancies. I added a PR here to address some issues: https://github.com/projectatomic/atomic/pull/954

ashcrow:
@yuqi-zhang thanks for the info!

@giuseppe I'm thinking that ``inspect_system_image`` won't end up being enough. The failures CI is finding is related to image names such as "docker:atomic-test-system:latest" which the ``inspect_system_image`` doesn't seem made for. Any pointers before I dig deeper in the code?

giuseppe:
@ashcrow let's add another exception for now `not image.startswith('dockertar:/') and not (image.startswith("docker:") and image.count(':') > 1)`.

`_install` has the code for pulling the image and then use the canonical name for the image, we can move the new code there once https://github.com/projectatomic/atomic/pull/949 gets merged and we don't end up in merge conflicts.

giuseppe:
I've added another fixup commit that solves the issue and a few other things.  There is only one missing part now, and that we need to set the correct `values` or the installation with template files will fail.  Now this code is in `_do_checkout` .

This requires some refactoring as the code in common must be moved out of `_do_checkout`, but I'd prefer if we do this after: https://github.com/projectatomic/atomic/pull/949 as it changes a lot of stuff around there.

The next failure in the tests is fixed by https://github.com/projectatomic/atomic/pull/949/commits/10ba9dc9c8998b9b7521acac9ff92b311e9d3ce0 (which is part of https://github.com/projectatomic/atomic/pull/949).  No idea how this could get broken.

giuseppe:
once 949 gets merged, we can use this commit here to refactor out the generate values function and use from here as well:
https://github.com/giuseppe/atomic/tree/move-read-values

ashcrow:
Local testing worked as expected.

giuseppe:
yes, I've done some testing as well and it worked for me (except handling template files).  We should probably add the oneshot image to the testing images as well.

ashcrow:
@giuseppe I'll add that shortly. I'll also stage a ``Labels`` document for the atomic-system-containers repo.

giuseppe:
@ashcrow thanks, could the image include a `config.json.template` file instead of `config.json`?

ashcrow:
@giuseppe the test image? Sure.

ashcrow:
Added a ``Dockerfile`` for the change and rebased.

ashcrow:
I'll add to integration tests an look at your added patch in the morning @giuseppe.

ashcrow:
Rebased.

giuseppe:
@ashcrow if you cherry-pick "syscontainers: refactor out values generation to a function" from https://github.com/giuseppe/atomic/tree/move-read-values we can add the last missing piece and generate VALUES to be feed to the template generator

ashcrow:
@giuseppe will do. I'm trupping my way through the bash for testing right now. Once I get a good result I'll cherry pick.

ashcrow:
@giuseppe I think this is ready. Anything else you think that needs to be done in this PR?

giuseppe:
@ashcrow just filling default values (like $NAME, $DESTDIR...), I've added a couple of patches here to do that:

https://github.com/giuseppe/atomic/tree/oneshot

If you are fine with them, could you cherry pick them?

Assuming tests pass after these two patches, I think it is ready to go!

ashcrow:
@giuseppe sure. Checking them out now.

ashcrow:
@giuseppe I did a fixup for one of the commits cherry-picke'd in. Let me know if you are OK with me squashing it in your commit: https://github.com/projectatomic/atomic/pull/952/commits/87dc5e16802f3afda225cf3711c2c761a8d7e064

Atomic/syscontainers.py
```diff
@@ -578,36 +578,46 @@ def _amend_values(self, values, manifest, name, image, image_id, destination, pr
...
             values = {}
 
-        remote_path = self._resolve_remote_path(remote)
-
-        imgs = self._resolve_image(repo, img)
-        if imgs is None:
-            raise ValueError("Image %s not found" % img)
+        # Get the rev or raise out of the method
+        try:
+            rev = self._resolve_image(repo, img)[0]
```
giuseppe: should be `_, rev = self._resolve_image(repo, img)[0]`

giuseppe:
I am fine with this PR.  If @rhatdan likes this as well, we can merge it

ashcrow:
Squshed fixups.

@rhatdan PTAL

ashcrow:
I'll rebase this again tomorrow.

@rhatdan PTAL when you have a chance.

giuseppe:
@ashcrow we will need to take in consideration the change here:

https://github.com/projectatomic/atomic/commit/f5b00bc171d0f324f1b40cceee7b8850f3c01761

That code is moved to a function now (just some reordering is needed)

ashcrow:
This PR is ready for review.

giuseppe:
LGTM

Atomic/syscontainers.py
```diff
@@ -495,7 +655,7 @@ def _do_checkout(self, repo, name, img, upgrade, deployment, values, destination
         else:
             if not os.path.exists(rootfs):
                 os.makedirs(rootfs)
-
+        # XXX
```
baude: ? leftover

Atomic/syscontainers.py
```diff
@@ -227,6 +227,121 @@ def build_rpm(self, repo, name, image, values, destination):
         return None
 
     def install(self, image, name):
+        """
+        External container install logic.
```
baude: I know we are not outwardly changing at of the interfaces to the users which usually prompts someone to do updates a docs/ file.  I'm thinking however it might be worth noting this change and the labels in the docs somwhere.

baude:
@ashcrow @giuseppe This might be out of scope but I would really love a mechanism for systemcontainers that would allow us to copy files from the system container to the host filesystem during install.  This seems like the right place for something like that.  Not sure.

Atomic/syscontainers.py
ashcrow: Yes, good catch. Will update.

Atomic/syscontainers.py
ashcrow: Right now there are 3 (or more?) places that document labels. I'm planning on adding it in the [Atomic System Containers repo](https://github.com/projectatomic/atomic-system-containers/blob/master/LABELS.md) and, as it is used, should propagate to wherever becomes the proper place. Or at least that is the theory.

ashcrow:
Removed the leftover and rebased (for good measure).

⬆️

rhatdan:
@baude Ready to merge?

baude:
@rh-atomic-bot r+  	6887eab

PR #955, opened 2017-03-28
Implement `install --storage=docker`

TomasTomecek:
This is now ready for review.

@giuseppe feel free to take changes I did in `rpm_host_install.py` into your #949. Right now there should be just two bugfixes.

TODO:

 * [x] clean up all temporary files
 * [x] store metadata about installed container images
 * [x] `--system_package` is ignored https://github.com/projectatomic/atomic/pull/955#issuecomment-296151720
 * [x] structure the code better
 * [x] intergration test
 * [x] document the behavior
 * [x] uninstall support


TomasTomecek:
@giuseppe I'm done with most of the changes, please take a look. It's not ready for merge, but definitely ready for a comprehensive review.

I'm assuming I should probably implement update and uninstall before merging this.

cgwalters:
What's the high level rationale for this?  Are we not using the ostree storage so that admins can use `docker rmi` on things as well?

Atomic/syscontainers.py
```diff
@@ -923,7 +923,12 @@ def _get_containers_at(self, checkouts, are_preinstalled, containers=None):
             if fullpath.endswith(".0") or fullpath.endswith(".1"):
                 continue
 
-            with open(os.path.join(fullpath, "info"), "r") as info_file:
+            try:
+                fd = open(os.path.join(fullpath, "info"), "r")
+            except FileNotFoundError:
+                # Something got wrong?
+                continue
+            with fd as info_file:
```
rhatdan: Don't you need an fd.close() now?

rhatdan: Why not just do a 
if not os.path.exists(os.path.join(fullpath, "info")):
           continue
with open ...

rhatdan:
I think we should have a design discussion about what you guys are trying to do here.  I would like to see a higherlevel User CLI DIscussion of how a user interacts with this.  

We are talking about Standalone containers and system containers and from a user point of view,  I don't see any difference here. 

Bottom line for all of these changes, it should be the responsibility for the packager/developer to put information into his container image that the atomic command can interpret and act on.  The end user should not need to make informed decisions on this if at all possible.  Meaning options are optional, not required.  `atomic install foobar` should do the right thing.

Atomic/syscontainers.py
TomasTomecek: I did it this way because exceptions are more pythonic. I don't mind rewriting the code if you prefer in your codebase using `if`s over exceptions.

The descriptor would not leak: if the file does not exist, the descriptor would not be created; if the file exists, the descriptor will be closed within with statement:

```python
In [10]: fd = open("/etc/passwd", "r")

In [11]: with fd as x:
   ....:     x.read(3)
   ....:

In [12]: x
Out[12]: <closed file '/etc/passwd', mode 'r' at 0x7f76252cf420>

In [13]: fd
Out[13]: <closed file '/etc/passwd', mode 'r' at 0x7f76252cf420>
```

giuseppe:
IIUIC, @TomasTomecek would like to have the possibility for Docker containers to copy files to the host as we already do for system containers.  I still believe that this class of containers should be handled using system containers.
If this is really needed though, my observation was that at least we need to add a new label since it is expensive to `mount/check-if-there-are-files-to-copy/umount` the image every time a Docker container is installed.  I proposed to add "atomic.has_install_files".  If this label is found then in the Docker case we proceed with what this PR does.  System containers won't honor the label as checking if there are files to copy is very cheap once the container is checked out to the file system.

TomasTomecek:
Thank you @giuseppe, I like your proposal.

@rhatdan @cgwalters are you okay with it?

rhatdan:
I'm fine with it as long as you guys document the behaviour.  Main goal is this is simple for users, second goal is this is not to difficult for developers of these standalone containers.

TomasTomecek:
@rhatdan thank you for the answer, I completely agree. I added it as an item to the initial todo list.

TomasTomecek:
@giuseppe I'm assuming that semantics of `atomic.has_install_files` will be: take files from container, create rpm, install it. Do we also want this to be configurable via `--system-package`?

```
  --system-package SYSTEM_PACKAGE
                        control how to install the package. It accepts `auto`,
                        `yes`, `no`, `build`
```

giuseppe:
I think that for the Docker backend `atomic.has_install_files` should change the behavior of `--system-package=auto`.  When the label is present, then `--system-package=auto` will behave as if `--system-package=yes` was specified, otherwise it should work as `--system-package=no`, as we don't mount the image to check whether the files to copy to the host are present or not.

Also, the Docker backend should fail when `--system-package=build` is used since it is not supported.

TomasTomecek:
got it, thank you; definitely makes sense

TomasTomecek:
@giuseppe I made some progress on this PR. What do I need to do to get further?

TomasTomecek:
I'm stuck on the CI failure: it works for me locally and errors out in CI: I pushed a commit with plenty of prints to help me identify the issue. It would super-helpful if I could reproduce the CI environment locally.

TomasTomecek:
Finally reproduced locally. The issue is in devicemapper graph backend.

TomasTomecek:
@baude @rhatdan @giuseppe 

This should be ready for review now. I removed every occurrence of word 'standalone' -- we shouldn't create new container classes just like that.

TomasTomecek:
bot, retest this please

Atomic/install.py
```diff
@@ -132,6 +160,42 @@ def install(self):
...
             be.syscontainers.set_args(self.args)
             return be.install(self.image, self.name)
+        if util.InstallData.image_installed(img_obj):
+            raise ValueError("Image {} is already installed.".format(self.image))
+
+        installation = None
+        if storage == 'docker' and not args_system:
+            if self.args.system_package == 'build':
+                raise ValueError("'--system-package=build' is not supported for docker backend")
+            labels = self.get_labels()
```
baude: We should be using the refactored approach to this where you get labels from the image_object.

iirc img_object.get_label or img_object.labels

Atomic/atomic.py
```diff
@@ -301,6 +313,15 @@ def is_iid(self):
                 return True
         return False
 
+    def get_image_id(self):
```
baude: We should be using the id from the image or container objects for this.  See Atomic/objects/image.py for example

Atomic/atomic.py
```diff
@@ -79,13 +79,25 @@ def docker_binary(self):
             self.docker_cmd = util.default_docker()
         return self.docker_cmd
 
-    def get_label(self, label, image=None):
+    def get_labels(self, image=None):
```
baude: Here again, would prefer the use of the labels from the image or container object.

Atomic/install.py
TomasTomecek: can do! I wasn't sure about the correct way.

Atomic/atomic.py
TomasTomecek: Yup, I will remove this.

Atomic/atomic.py
TomasTomecek: Thank you for the pointer, I was aware of that.

TomasTomecek:
(will clean up the commits once lgtm)

Atomic/backends/_docker.py
```diff
@@ -399,11 +399,18 @@ def uninstall(self, iobject, name=None, **kwargs):
         if cmd:
             return util.check_call(cmd, env=atomic.cmd_env())
 
+        install_data = util.InstallData.get_install_data_by_id(iobject.id)
+        system_package_nvra = install_data.get("system_package_nvra", None)
+        if system_package_nvra:
+            if os.path.exists("/usr/bin/dnf"):
+                util.check_call(["dnf", "remove", "-y", system_package_nvra])
+            else:
+                util.check_call(["yum", "remove", "-y", system_package_nvra])
```
giuseppe: this code is almost the same in `syscontainers.py`.  We should refactor this to `rpm_host_install.py` and make both Docker and system containers use the same functions.

Atomic/install.py
```diff
@@ -80,8 +86,31 @@ def cli(subparser):
 
 
 class Install(Atomic):
-    def __init__(self):
-        super(Install, self).__init__()
+    def build_rpm_for_docker_backend(self, image_id, temp_dir, labels):
```
giuseppe: this looks like Docker specific stuff, we should move it to `backends/_docker.py`

Atomic/install.py
```diff
@@ -133,23 +162,51 @@ def install(self):
...
+            if self.args.system_package == 'build':
+                raise ValueError("'--system-package=build' is not supported for docker backend")
+            labels = img_obj.labels or {}
+            # this means --system-package=yes
+            # we actually don't care about value of the label
+            if 'atomic.has_install_files' in labels:
+                temp_dir = tempfile.mkdtemp()
+                try:
+                    installation = self.build_rpm_for_docker_backend(img_obj.id, temp_dir, labels)
+                    self.syscontainers.install_rpm(installation.destination_path)
```
giuseppe: same issue here: the code is Docker specific and should stay in `backends/_docker.py`.

Also here we are using a syscontainers function for Docker containers. the refactor I suggested earlier will help to clean this up as well

Atomic/syscontainers.py
```diff
@@ -181,7 +181,7 @@ def install_user_container(self, image, name):
         # Same entrypoint
         return self.install(image, name)
 
-    def _install_rpm(self, rpm_file):
+    def install_rpm(self, rpm_file):
```
giuseppe: let's move this (and `uninstall_rpm`) to `rpm_host_install.py`.

Atomic/syscontainers.py
```diff
@@ -861,7 +861,7 @@ def _do_checkout(self, repo, name, img, upgrade, deployment, values, destination
     def _get_preinstalled_containers_path(self):
         return ATOMIC_USR
 
-    def _get_system_checkout_path(self):
+    def get_system_checkout_path(self):
```
giuseppe: it seems that it is used only in `syscontainers.py`, are you planning on using it somewhere else?  Otherwise I think it should still be private

Atomic/backends/_docker.py
TomasTomecek: good suggestion, can do! (I was lazy to refactor it)

Atomic/syscontainers.py
TomasTomecek: One of the previous iterations of my code used this, can revert to be private again.

TomasTomecek:
Just addressed all @giuseppe's comments.

Atomic/syscontainers.py
```diff
@@ -1061,13 +1053,23 @@ def _get_containers_at(self, checkouts, are_preinstalled, containers=None):
             if fullpath.endswith(".0") or fullpath.endswith(".1"):
                 continue
 
-            with open(os.path.join(fullpath, "info"), "r") as info_file:
+            try:
+                fd = open(os.path.join(fullpath, "info"), "r")
+            except IOError:
+                # Something got wrong?
+                continue
+            with fd as info_file:
```
giuseppe: please drop this part, if something went wrong we must report the error, or am I missing something?

Atomic/syscontainers.py
```diff
@@ -1061,13 +1053,23 @@ def _get_containers_at(self, checkouts, are_preinstalled, containers=None):
...
                 created = info["created"] if "created" in info else 0
                 image = info["image"] if "image" in info else ""
 
-            with open(os.path.join(fullpath, "config.json"), "r") as config_file:
+            try:
+                fd = open(os.path.join(fullpath, "config.json"), "r")
+            except IOError:
+                # Something got wrong?
+                continue
+            with fd as config_file:
```
giuseppe: same here

Atomic/syscontainers.py
TomasTomecek: I ran into this during first iterations. No problem dropping it.

TomasTomecek:
I addressed all the issues.

giuseppe:
@TomasTomecek it looks very neat now and it is fine for me.  Waiting for @baude or @rhatdan for the Docker part.

Atomic/backends/_docker.py
```diff
@@ -362,7 +412,25 @@ def get_dangling_images(self):
         return self._get_images(get_all=True, quiet=True, filters={"dangling": True})
 
     def install(self, image, name, **kwargs):
```
baude: Do you think instead of passing this, we should consider call this method rpm_install?

TomasTomecek: Makes sense to me.

baude:
@rh-atomic-bot r+  	63a9df9

TomasTomecek:
Thank you to everyone helping me here.

PR #958, opened 2017-03-31
images: support creating tags

giuseppe:
Add a new command "atomic images tag" to create a tag of a local image

### DIFF
--- a/Atomic/tag.py
+++ b/Atomic/tag.py
@@ -1,0 +1,35 @@
+from . import Atomic
+from . import util
+from Atomic.backendutils import BackendUtils
+
+ATOMIC_CONFIG = util.get_atomic_config()
+storage = ATOMIC_CONFIG.get('default_storage', "docker")
+
+class Tag(Atomic):
+    def __init__(self):
+        super(Tag, self).__init__()
+        self.be = None
+
+    def tag_image(self):
+        """
+        Tag an image with a different name
+        :return: 0 if the tag was created
+        """
+        if self.args.debug:
+            util.write_out(str(self.args))
+
+        beu = BackendUtils()
+
+        backend = None
+        if self.args.storage:
+            backend = beu.get_backend_from_string(self.args.storage)
+        else:            
+            backend, _ = beu.get_backend_and_image_obj(self.args.src, required=False)
+
+        if not backend:
+            raise ValueError("Cannot find image {}.".format(self.args.src))
+
+        backend.tag_image(self.args.src, self.args.target)
+
+        # We need to return something here for dbus
+        return 0
--- a/atomic_dbus.py
+++ b/atomic_dbus.py
@@ -241,7 +241,7 @@ class atomic_dbus(slip.dbus.service.Object):
         args = self.Args()
         args.all=True
         images.set_args(args)
-        i = images.display_all_image_info()
+        i = images.images()
         return json.dumps(i)
 
     # atomic containers section
--- a/atomic_dbus.py
+++ b/atomic_dbus.py
@@ -253,6 +253,7 @@ class atomic_dbus(slip.dbus.service.Object):
         args.delete_targets = images
         args.remote = remote
         args.force = force
+        args.assumeyes = True
         i.set_args(args)
         return i.delete_image()
 
--- a/atomic_dbus.py
+++ b/atomic_dbus.py
@@ -246,7 +246,7 @@ class atomic_dbus(slip.dbus.service.Object):
     # atomic containers section
     # The ImagesDelete method will delete one or more images on the system.
     @slip.dbus.polkit.require_auth("org.atomic.readwrite")
-    @dbus.service.method("org.atomic", in_signature='asbb', out_signature='')
+    @dbus.service.method("org.atomic", in_signature='asbb', out_signature='i')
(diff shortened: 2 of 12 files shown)


baude:
@giuseppe what would you think of adding --tag to pull?  Then the tag wouldn't have to be a secondary step in the kickstart?

atomic_dbus.py
```diff
@@ -256,6 +259,19 @@ def ImagesDelete(self, images, force, remote):
         i.set_args(args)
         return i.delete_image()
 
+    # atomic containers section
+    # The ImagesTag method will create a tag from an existing image.
+    @slip.dbus.polkit.require_auth("org.atomic.readwrite")
+    @dbus.service.method("org.atomic", in_signature='sss', out_signature='i')
+    def ImagesTag(self, storage, src, dest):
```
baude: You should add a ImagesTag to the client file too.

tests/integration/test_system_containers.sh
```diff
@@ -256,8 +256,10 @@ test \! -e ${ATOMIC_OSTREE_CHECKOUT_PATH}/${NAME}.0
 test \! -e ${ATOMIC_OSTREE_CHECKOUT_PATH}/${NAME}.1
 
 ${ATOMIC} pull --storage ostree docker:atomic-test-secret:latest
-# Move directly the OSTree reference to a new one, so that we have different names and info doesn't error out
-mv ${ATOMIC_OSTREE_REPO}/refs/heads/ociimage/atomic-test-secret_3Alatest ${ATOMIC_OSTREE_REPO}/refs/heads/ociimage/atomic-test-secret-ostree_3Alatest
+
+${ATOMIC} images tag --storage ostree atomic-test-secret:latest atomic-test-secret-ostree:latest
+${ATOMIC} --assumeyes images delete -f --storage ostree atomic-test-secret:latest
```
baude: I think a dbus test and possible a unittest would be appropriate here too?

Atomic/tag.py
```diff
@@ -0,0 +1,35 @@
...
+        backend = None
+        if self.args.storage:
+            backend = beu.get_backend_from_string(self.args.storage)
+        else:            
+            backend, _ = beu.get_backend_and_image_obj(self.args.src, required=False)
+
+        if not backend:
+            raise ValueError("Cannot find image {}.".format(self.args.src))
+
+        backend.tag_image(self.args.src, self.args.target)
```
baude: @giuseppe as I understood the conversation yesterday, they wanted to be able to delete any trace of the original image pulled which likely would include the registry name.  So does tag need something like a --remove which would remove the original image after being tagged so that all that is left is the newly tagged image?

giuseppe: This feature is slightly more generic that the specific request.  I think that an additional `images delete` is not a big issue

@imcleod @3dav1x

tests/integration/test_system_containers.sh
giuseppe: yes sure

atomic_dbus.py
giuseppe: Thanks, I will fix that

Atomic/tag.py
baude: @giuseppe actually i have wanted something like this too when building/pushing/testing images.  I agree no biggie but would be nice.

giuseppe:
bot, please retry

giuseppe:
bot, please retry

giuseppe:
@baude tests added and passing now.  I had to fix some other issues in the dbus code

atomic_dbus.py
```diff
@@ -240,22 +244,37 @@ def ImagesList(self):
...
+        args.storage = storage
+        args.assumeyes = True
         i.set_args(args)
         return i.delete_image()
 
+    # atomic containers section
+    # The ImagesTag method will create a tag from an existing image.
+    @slip.dbus.polkit.require_auth("org.atomic.readwrite")
+    @dbus.service.method("org.atomic", in_signature='sss', out_signature='i')
+    def ImagesTag(self, src, target, storage='docker'):
```
rhatdan: Should not have default here.  Default would be on other side.  dbus is going to force the client to pass three fields.

atomic_dbus.py
```diff
@@ -240,22 +244,37 @@ def ImagesList(self):
...
+        i = images.images()
         return json.dumps(i)
 
     # atomic containers section
     # The ImagesDelete method will delete one or more images on the system.
     @slip.dbus.polkit.require_auth("org.atomic.readwrite")
-    @dbus.service.method("org.atomic", in_signature='asbb', out_signature='')
-    def ImagesDelete(self, images, force, remote):
+    @dbus.service.method("org.atomic", in_signature='asbbs', out_signature='i')
+    def ImagesDelete(self, images, force, remote, storage='docker'):
```
rhatdan: Should not have default here. Default would be on other side. dbus is going to force the client to pass three fields.

giuseppe:
@rhatdan added two fixup commits to address your comments

rhatdan:
:+1:

giuseppe:
@rhatdan can this get merged?

rhatdan:
@rh-atomic-bot r+ f3b0b31

PR #962, opened 2017-04-03
syscontainers: fix overridable vars

yuqi-zhang:
Allow RUN_DIRECTORY and STATE_DIRECTORY to be overridable if the user
sets them as a value or they are present in manifest.json, as they are
in TEMPLATE_OVERRIDABLE_VARIABLES.


yuqi-zhang:
@giuseppe I'm not sure if we decided to change that behaviour at some point, but if not this should fix it to "--set" -> manifest.json value -> default value priority order.

giuseppe:
@yuqi-zhang yes it looks fine to me, it is probably going to cause a merge conflict in https://github.com/projectatomic/atomic/pull/952 but we will deal with it in the other PR

Thanks for fixing it!

@rh-atomic-bot r+  	b9a7447

giuseppe:
@rh-atomic-bot r+ b9a7447

yuqi-zhang:
Want to give this a shot again @giuseppe

giuseppe:
absolutely!

@rh-atomic-bot r+ b9a7447

PR #964, opened 2017-04-03
Honor proxy usage

baude:
If HTTP[S]_PROXY is defined, honor it in python requests usage
as well as pass it on to skopeo.

### DIFF
--- a/Atomic/push.py
+++ b/Atomic/push.py
@@ -19,6 +19,10 @@ def cli(subparser):
         "push", aliases=['upload'], help=_("push latest image to repository"),
         epilog="push the latest specified image to a repository.")
     pushp.set_defaults(_class=Push, func='push')
+    pushp.add_argument("--anonymous",
+                     default=False,
+                     action="store_true",
+                     help=_("Push images without a username or password"))
     # making it so we cannot call both the --pulp and --satellite commands
     # at the same time (mutually exclusive)
     pushgroup = pushp.add_mutually_exclusive_group()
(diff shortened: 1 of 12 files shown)


baude:
[(?) Atomic CLI support for HTTP/HTTPS_PROXY env vars](https://trello.com/c/qa3obU4v/353-atomic-cli-support-for-http-https-proxy-env-vars)

Atomic/push.py
```diff
@@ -69,6 +69,8 @@ def cli(subparser):
                               "use an alternate user's GPG keyring for signing. "
                               "Useful when running with sudo, "
                               "e.g. set to '~/.gnupg'."))
+    pushp.add_argument("--insecure", dest="insecure", default=False,
```
giuseppe: should this new option be added to bash autocomplete and to the documentation?

rhatdan:
It seems that there are multiple pull requests combined here.  Can we separate them out.  I agree with giuseppe on the need for man page and command completion.

Atomic/util.py
```diff
@@ -124,7 +123,8 @@ def subp(cmd, cwd=None, newline=False):
     proc = subprocess.Popen(cmd, cwd=cwd,
                             stdout=subprocess.PIPE,
                             stderr=subprocess.PIPE, close_fds=True,
-                            universal_newlines=newline)
+                            universal_newlines=newline,
+                            env=os.environ)
```
rhatdan: What do these have to do with the --insecure flag?

baude: When we call skopeo, we need to pass the proxy information which is done via environment variables.

Atomic/push.py
baude: It is already in the bash completion, I'll add to the docs.

baude:
@rhatdan there are not multiple requests.  This PR covers proxy work and that we need to be able to allow the user to indicate if the registry is insecure because we cannot authoritatively be sure.

baude:
@rhatdan I added the ability to read atomic.conf.  Ready for re-review

baude:
bot, retest this please

ddarrah:
LGTM - this will really help simplify CI.

rhatdan:
@rh-atomic-bot r+ c0a844f

Issue #963, opened 2017-04-03
requires root privileges and touches /var during build


cgwalters:
https://ci.centos.org/view/Atomic/job/fahc-rdgo/1443/artifact/build-logs/failed/atomic/build.log

```
+ make PYTHON=/usr/bin/python3 PYLINT=true all
/usr/bin/python3 setup.py build
Traceback (most recent call last):
  File "setup.py", line 5, in <module>
    import Atomic as _Atomic
  File "/builddir/build/BUILD/atomic-v1.16.3-52-f88187d5e42c496b4ed7a2ffc3a891847def925d/Atomic/__init__.py", line 1, in <module>
    from .pulp import PulpServer, PulpConfig
  File "/builddir/build/BUILD/atomic-v1.16.3-52-f88187d5e42c496b4ed7a2ffc3a891847def925d/Atomic/pulp.py", line 10, in <module>
    from . import util
  File "/builddir/build/BUILD/atomic-v1.16.3-52-f88187d5e42c496b4ed7a2ffc3a891847def925d/Atomic/util.py", line 37, in <module>
    os.makedirs(ATOMIC_VAR_LIB)
  File "/usr/lib64/python3.5/os.py", line 241, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/var/lib/atomic'
```

jlebon:
@baude Basically, the `mkdir` shouldn't be triggered by importing the module. E.g. you can simply move it to create when needed.

baude:
PTAL https://github.com/projectatomic/atomic/pull/966

PR #966, opened 2017-04-06
requires root privileges and touches /var during build

baude:
Fixes issue #963

https://github.com/projectatomic/atomic/issues/963

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -6,6 +6,7 @@ import sys
 import json
 import subprocess
 import collections
+from contextlib import contextmanager
 from fnmatch import fnmatch as matches
 import os
 import selinux
@@ -800,44 +801,27 @@ def load_scan_result_file(file_name):
     return json.loads(open(os.path.join(file_name), "r").read())
 
 
-def file_lock(func):
-    lock_file_name = "{}.lock".format(os.path.join(os.path.dirname(ATOMIC_INSTALL_JSON), "." + os.path.basename(ATOMIC_INSTALL_JSON)))
-
-    def get_lock():
-        '''
-        Obtains a read-only file lock on the install data
-        :return: 
-        '''
-        time_out = 0
-        f_lock = False
-        # Create the temporary lockfile if it doesn't exist
-        if not os.path.exists(lock_file_name):
-            open(lock_file_name, 'a').close()
-
-        with open(lock_file_name, "r") as f:
-            while time_out < 10.5: # Ten second attempt to get a lock
-                try:
-                    fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
-                    f_lock = True
-                    break
-                except IOError:
-                    time.sleep(.5)
-                    time_out += .5
+@contextmanager
+def file_lock(path):
+    lock_file_name = "{}.lock".format(path)
+    time_out = 0
+    f_lock = False
+    with open(lock_file_name, "a") as f:
+        while time_out < 10.5: # Ten second attempt to get a lock
+            try:
+                fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                f_lock = True
+                break
+            except IOError:
+                time.sleep(.5)
+                time_out += .5
         if not f_lock:
-            raise ValueError("Unable to get file lock for {}".format(ATOMIC_INSTALL_JSON))
+            raise ValueError("Unable to get file lock for {}".format(lock_file_name))
 
-    def release_lock():
-        with open(lock_file_name, "r") as f:
-            fcntl.flock(f, fcntl.LOCK_UN)
-
-    def wrapper(*args, **kwargs):
-        get_lock()
-        ret = func(*args, **kwargs)
-        release_lock()
-        return ret
-
-    return wrapper
-
+        # Call the user code
+        yield
+        # Now unlock
+        fcntl.flock(f, fcntl.LOCK_UN)
 
 # Records additional data for containers outside of the native storage (docker/ostree)
 class InstallData(object):
@@ -869,17 +853,14 @@ class InstallData(object):
         shutil.move(temp_file.name, ATOMIC_INSTALL_JSON)
 
     @classmethod
-    @file_lock
     def read_install_data(cls):
-        if os.path.exists(ATOMIC_INSTALL_JSON):
-            read_data = cls._read_install_data()
-            return read_data
-        return {}
+        with file_lock(ATOMIC_INSTALL_JSON):
+            return cls._read_install_data()
 
     @classmethod
-    @file_lock
     def write_install_data(cls, new_data):
-        cls._write_install_data(new_data)
+        with file_lock(ATOMIC_INSTALL_JSON):
+            cls._write_install_data(new_data)
 
     @classmethod
     def get_install_name_by_id(cls, iid, install_data=None):
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -827,40 +827,32 @@ def file_lock(path):
 class InstallData(object):
 
     @classmethod
-    def _read_install_data(cls):
-        try:
-            with open(ATOMIC_INSTALL_JSON, 'r') as f:
-                # Backwards compatibilty - we previously created an empty file explicitly;
-                # see https://github.com/projectatomic/atomic/pull/966
-                if os.fstat(f.fileno()).st_size == 0:
-                    return {}
-                return json.load(f)
-        except IOError as e:
-            if e.errno == errno.ENOENT:
-                return {}
-            raise e
-
-    @classmethod
-    def _write_install_data(cls, new_data):
-        install_data = cls._read_install_data()
-        for x in new_data:
-            install_data[x] = new_data[x]
-        temp_file = tempfile.NamedTemporaryFile(mode='w', delete=False)
-        json.dump(install_data, temp_file)
-        temp_file.close()
-        if not os.path.exists(ATOMIC_VAR_LIB):
-            os.makedirs(ATOMIC_VAR_LIB)
-        shutil.move(temp_file.name, ATOMIC_INSTALL_JSON)
-
-    @classmethod
     def read_install_data(cls):
         with file_lock(ATOMIC_INSTALL_JSON):
-            return cls._read_install_data()
+            try:
+                with open(ATOMIC_INSTALL_JSON, 'r') as f:
+                    # Backwards compatibilty - we previously created an empty file explicitly;
+                    # see https://github.com/projectatomic/atomic/pull/966
+                    if os.fstat(f.fileno()).st_size == 0:
+                        return {}
+                    return json.load(f)
+            except IOError as e:
+                if e.errno == errno.ENOENT:
+                    return {}
+                raise e
 
     @classmethod
     def write_install_data(cls, new_data):
         with file_lock(ATOMIC_INSTALL_JSON):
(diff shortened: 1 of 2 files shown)


baude:
bot, retest this please

baude:
bot, retest this please

rhatdan:
Tests still failing?

rhatdan:
LGTM if you fix tests failing.

baude:
bot, retest this please

Atomic/util.py
```diff
@@ -842,10 +842,16 @@ def wrapper(*args, **kwargs):
...
-        open(ATOMIC_INSTALL_JSON, 'a').close()
 
-    install_file_handle = open(ATOMIC_INSTALL_JSON, 'r')
+    @staticmethod
+    def get_install_file_handle():
+        if not os.path.exists(ATOMIC_VAR_LIB):
+            os.makedirs(ATOMIC_VAR_LIB)
+
+        if not os.path.exists(ATOMIC_INSTALL_JSON):
+            open(ATOMIC_INSTALL_JSON, 'a').close()
```
cgwalters: I don't understand why we create an empty file explicitly.  It's not valid JSON.  Why not simply have "absence of file" = empty?  Basically we don't explicitly write it when we're just opening for *read*.

Atomic/util.py
```diff
@@ -803,13 +801,15 @@ def load_scan_result_file(file_name):
...
-        open(lock_file_name, 'a').close()
+    def get_lock_file_handle() :
+        lock_file_name = "{}.lock".format(os.path.join(os.path.dirname(ATOMIC_INSTALL_JSON), "." + os.path.basename(ATOMIC_INSTALL_JSON)))
 
-    install_data_file = open(lock_file_name, "r")
+        # Create the temporary lockfile if it doesn't exist
+        if not os.path.exists(lock_file_name):
+            open(lock_file_name, 'a').close()
+
+        return open(lock_file_name, "r")
```
jlebon: Won't this leak though? Don't we want to `close()` in `release_lock()`?

Atomic/util.py
```diff
@@ -842,10 +842,16 @@ def wrapper(*args, **kwargs):
...
-    install_file_handle = open(ATOMIC_INSTALL_JSON, 'r')
+    @staticmethod
+    def get_install_file_handle():
+        if not os.path.exists(ATOMIC_VAR_LIB):
+            os.makedirs(ATOMIC_VAR_LIB)
+
+        if not os.path.exists(ATOMIC_INSTALL_JSON):
+            open(ATOMIC_INSTALL_JSON, 'a').close()
+
+        return open(ATOMIC_INSTALL_JSON, 'r')
```
jlebon: Here too.

Atomic/util.py
```diff
@@ -842,21 +842,31 @@ def wrapper(*args, **kwargs):
...
+            return install_data
         except ValueError:
             return {}
 
     @classmethod
     def _write_install_data(cls, new_data):
-        install_data = cls._read_install_data(cls.install_file_handle)
+        try:
+            install_data = cls._read_install_data()
+        except IOError:
```
jlebon: This is already caught by `_read_install_data`, no? So you should already be getting `{}` back in that case.

Atomic/util.py
```diff
@@ -842,21 +842,31 @@ def wrapper(*args, **kwargs):
...
 
 class InstallData(object):
-    if not os.path.exists(ATOMIC_INSTALL_JSON):
-        open(ATOMIC_INSTALL_JSON, 'a').close()
-
-    install_file_handle = open(ATOMIC_INSTALL_JSON, 'r')
 
     @staticmethod
-    def _read_install_data(file_handle):
+    def get_install_file_handle():
```
jlebon: Does this need to be a function? Seems like we can just merge it into `_read_install_data()` so that you can just use context managers there to open and close, e.g.:

```python
try:
  with open(ATOMIC_INSTALL_JSON, 'r') as f:
    return json.load(f)
except IOError as e:
  if e.errno == errno.ENOENT:
    return {}
  raise e
```

?

cgwalters: Agree with this, except I think for backcompat with the previous "empty file" model we should make it something like:

```
try:
  with open(ATOMIC_INSTALL_JSON, 'r') as f:
    if os.fstat(f.fileno).st_size == 0:
      return {}
    return json.load(f)
except IOError as e:
  if e.errno == errno.ENOENT:
    return {}
  raise e

**Comment by cgwalters** [comment]
I went ahead and patched things up as above - there were more fixes required in the decorator, and I didn't run any tests other than simply doing as *non-root*:

```
$ python setup.py build
```

We clearly need to tweak the redhat-ci default build process to be more mock-like.

File: Atomic/util.py [hunk 6]
```diff
@@ -805,32 +803,32 @@ def load_scan_result_file(file_name):
...
-                time.sleep(.5)
-                time_out += .5
+        # Create the temporary lockfile if it doesn't exist
+        if not os.path.exists(lock_file_name):
+            open(lock_file_name, 'a').close()
+
+        with open(lock_file_name, "r") as f:
+            while time_out < 10.5: # Ten second attempt to get a lock
+                try:
+                    fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
```
jlebon: Won't this lock be released when `f` is closed though?
Just tried this out:

```python
with open("lockfile", "r") as f:
    fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
with open("lockfile", "r") as f:
    fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
```

which works without issue. Whereas

```python
f = open("lockfile", "r")
fcntl.flock(f, fcntl.LOCK_EX | fcntl.LOCK_NB)

g = open("lockfile", "r")
fcntl.flock(g, fcntl.LOCK_EX | fcntl.LOCK_NB)
```

throws a BlockingIOError as expected.

cgwalters: OK, let's try making it a context manager.   There is also https://pypi.python.org/pypi/flockcontext ...dunno about adding the dep.

cgwalters:
OK, this passes the tests now.

Atomic/util.py
```diff
@@ -802,102 +801,79 @@ def load_scan_result_file(file_name):
...
-        get_lock()
-        ret = func(*args, **kwargs)
-        release_lock()
-        return ret
-
-    return wrapper
+            raise ValueError("Unable to get file lock for {}".format(lock_file_name))
 
+        # Call the user code
+        yield
```
jlebon: Can we yield the file handle here, so that... (continued in next comment)

Atomic/util.py
```diff
@@ -802,102 +801,79 @@ def load_scan_result_file(file_name):
...
+                if e.errno == errno.ENOENT:
+                    return {}
+                raise e
 
     @classmethod
-    @file_lock
     def write_install_data(cls, new_data):
-        cls._write_install_data(new_data)
+        install_data = cls.read_install_data()
+        with file_lock(ATOMIC_INSTALL_JSON):
```
jlebon: here, we can do the whole read, update, write under lock?

Atomic/util.py
```diff
@@ -802,102 +801,79 @@ def load_scan_result_file(file_name):
...
     def delete_by_id(cls, iid, ignore=False):
-        install_data = cls._read_install_data(cls.install_file_handle)
+        install_data = cls.read_install_data()
         try:
             id_key = InstallData.get_install_name_by_id(iid, install_data=install_data)
         except ValueError as e:
             if not ignore:
                 raise ValueError(str(e))
             return
         del install_data[id_key]
```
jlebon: Wait, this won't take effect, right? The `write` methods (both old and new) are merging entries, so can deletes actually occur?

Atomic/util.py
```diff
@@ -802,102 +801,79 @@ def load_scan_result_file(file_name):
...
 
     @classmethod
-    @file_lock
     def write_install_data(cls, new_data):
-        cls._write_install_data(new_data)
+        install_data = cls.read_install_data()
+        with file_lock(ATOMIC_INSTALL_JSON):
+            for x in new_data:
+                install_data[x] = new_data[x]
+                temp_file = tempfile.NamedTemporaryFile(mode='w', delete=False)
```
jlebon: Seems like this line and the rest below are supposed to be unindented by one level?

cgwalters: Hah, yes.

Atomic/util.py
cgwalters: Probably, but that's not a regression from this patch.

I could think of *many* more things to do to improve this, not least of which is moving it to its own file with documentation, and such...but can we do that kind of thing in a separate PR?

I'd really like to get FAHC going again.

(And really, do we even *really* need this InstallData thing?  Can't we (for system containers) use the ostree commit metadata?)

Atomic/util.py
cgwalters: Pushed a fixup for that ⬇️

cgwalters:
As far as I can tell right now, the *only* place that *reads* (not writes) InstallData right now is:

```
        if iobject.get_label('INSTALL') and not args.ignore and not util.InstallData.image_installed(iobject):
 ```

And that's just for a warning.  How about we just delete this code?  I'd like to deprecate the `INSTALL` label anyways in favor of the system container approach.

rhatdan:
I am fine with moving that way.

rhatdan:
I am fine with de-emphasizing it, not removing it.

cgwalters:
https://github.com/projectatomic/atomic/pull/969

cgwalters:
A good chunk of the `INSTALL` use cases should be system containers.  The rest I think are more in the [system extensions](https://github.com/projectatomic/atomic/pull/767) case - and there the underlying data store tracking content is RPM.  So in neither case do we need our own data store.   Right?

baude:
so the current RHEL rsyslog container is not a use case? same with the openscap image? We have spent years trying to get people to use labels.  Maybe I am missing something.

cgwalters:
Yeah, we need to make the rsyslog container a system container.

rhatdan:
I think we will need to continue to support INSTALL Labels but we really want to move most cases where an app is dropping content on the host to use something like system containers.  INSTALL Labels begin to fall apart when you don't require docker.

cgwalters:
`INSTALL` also forces the container author to own problems like SELinux labeling, handling any conflicts with files they're trying to install.  And the existence of the `UNINSTALL` label gets into deep issues around things like "what happens if I pull a newer version of the container with a different `UNINSTALL`?

Basically `INSTALL` was a good start at having *something* but I think we're past the prototype stage and need to have a more rigorous model.  Basically, rigorous = using RPM to track installed files and handle config files in `/etc`.  That way one centralized place is also taking care of SELinux labeling as well.  We already have a database for content installed on the host - RPM.

cgwalters:
There were things blocking this before, like rpm-ostree not supporting local RPM installs, etc. that we've fixed now.  Though to really complete that picture we need [livefs](https://github.com/projectatomic/rpm-ostree/pull/652) which I'm working on.

cgwalters:
Anyways I think this PR is in an OK state - it could be a lot better but I'd really like to just get FAHC going and revisit this in later PRs.

baude:
LGTM

rhatdan:
@rh-atomic-bot r+ 4e6de1f

cgwalters:
Note CI is currently down https://github.com/jlebon/redhat-ci/issues/33

Issue #965, opened 2017-04-05
Year is wrong for 2017 changelog entries


jberkus:
... they say 2016 instead.

PR #968, opened 2017-04-12
Fix changelog to correct year

yuqi-zhang:
As pointed out by https://github.com/projectatomic/atomic/issues/965


rhatdan:
Will see if the failures happen again on approval.
@rh-atomic-bot r+ c18c489

rhatdan:
@rh-atomic-bot retry

PR #969, opened 2017-04-13
Stop reading/writing "install data" in /var/lib/atomic

cgwalters:
This reverts: https://github.com/projectatomic/atomic/pull/950
See: https://github.com/projectatomic/atomic/pull/966

The idea here originally was good, but we have to be a *lot* more
careful in introducing data stores than this.  They need "schema design",
documentation, and most importantly IMO: lifecycle coupling.

If I e.g. `rm -rf /var/lib/docker` (as a lot of documentation tells
people to do), this install data would stick around.  Similarly
for deleting content directly in `/ostree/repo`.  You could say
that's a bug, but it'll happen.

The best way to do metadata is to use the underlying store.  OSTree
in particular already has an API for storing metadata.  Any containers
we write there should use that.

I'm not sure offhand if we can do something similar with Docker.
But basically I don't think it's worth carrying another data store
*just* to show a warning.


baude:
This was not something invented for just a warning.  This was a user-based request.  We have people trying to run containers before installing them and then reporting that as a bug.  The openscap image is a perfect example.

We knew this implementation was going to be imperfect but it was better that doing nothing.

cgwalters:
OK.  We can close this and return to fixing the existing implementation.

baude:
I'd like your ideas on how it can be improved.  Prefer to work this as an issue?

PR #973, opened 2017-04-17
Atomic/install.py: Fix what c134ee5 broke

baude:
The commit c134ee5 broke the ability for atomic install to pull an image
if it isnt present.

### DIFF
--- a/Atomic/install.py
+++ b/Atomic/install.py
@@ -104,26 +104,34 @@ class Install(Atomic):
             if not util.is_user_mode():
                 raise ValueError("--user does not work for privileged user")
             return self.syscontainers.install_user_container(self.image, self.name)
-        elif self.system or storage == 'ostree':
+        if self.ostree_uri(self.image):
             return self.syscontainers.install(self.image, self.name)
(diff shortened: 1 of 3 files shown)


yuqi-zhang:
Sorry for not catching this case. There is one minor problem with this fix being that remote ostree images still get ignored, i.e. if you do something like "atomic install --system gscrivano/flannel" (which has atomic.type=ostree) it would just say `Image gscrivano/flannel not found`. This is also the case if I use fqn's. So this fixes the docker part that I broke, but breaks the ostree part I fixed...

I will take another look at this today. Thanks for the fix!

miabbott:
Confirming what @yuqi-zhang said, this appears to fix the case of `atomic install` of a remote docker image, but breaks the case of `atomic install` of remote ostree image.

miabbott:
The latest commit works better for the various use cases.  👍

yuqi-zhang:
LGTM

baude:
@rhatdan shall i merge?

rhatdan:
@rh-atomic-bot r+ d6a5e48

baude:
@rh-atomic-bot retry

PR #974, opened 2017-04-19
Atomic/containers.py: Fix dbus client calls

baude:
No Arg existed for 'filters' when processing dbus calls to
containerslist.  This results in an AttributeError.  Changed
'filters' to 'filter' to align with ImagesList.


baude:
@rh-atomic-bot r+ 	b6bcf88

PR #978, opened 2017-04-24
.redhat-ci.sh: resize root LV

jlebon:
We have plenty of disk space here. So let's give the root LV a bit more
so that some of our new tests can run properly.


TomasTomecek:
Thank you!

TomasTomecek:
Tests are passing, can we merge this?

cgwalters:
This should have been mergable via homu, no?

PR #979, opened 2017-04-26
System containers storage

giuseppe:
## Description

Introduce system containers storage.  It is used to checkout each layer separately under $CHECKOUT/.storage/$LAYER_ID.

This is going to be used for 'atomic mount' and 'one shot' containers that are controlled from `atomic` and that can mount the image using overlay2

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1779,3 +1779,43 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         repo.prepare_transaction()
         repo.transaction_set_ref(None, get_image_branch(dest), rev)
         repo.commit_transaction(None)
+
+
+    def get_storage_path(self):
+        return os.path.sep.join([self._get_system_checkout_path(), ".storage"])
+
+
+    def _ensure_storage_for_image(self, repo, img):
+        # Get the rev or raise out of the method
+        try:
+            _, rev = self._resolve_image(repo, img)[0]
+        except (IndexError, TypeError):
+            raise ValueError("Image {} not found".format(img))
+        manifest = self._image_manifest(repo, rev)
+        if manifest is None:
+            raise ValueError("Image `%s` not present" % img)
+        layers = SystemContainers.get_layers_from_manifest(json.loads(manifest))
+
+        storage_path = self.get_storage_path()
+        layers_dir = []
+        for i in layers:
+            layer = SystemContainers._drop_sha256_prefix(i)
+            rootfs = os.path.sep.join([storage_path, layer])
+            layers_dir.append(rootfs)
+            if os.path.exists(rootfs):
+                continue
+            os.makedirs(rootfs)
+            rootfs_fd = None
+            try:
+                rootfs_fd = os.open(rootfs, os.O_DIRECTORY)
+                branch = "{}{}".format(OSTREE_OCIIMAGE_PREFIX, SystemContainers._drop_sha256_prefix(layer))
+                rev_layer = repo.resolve_rev(branch, True)[1]
+                if not rev_layer:
+                    raise ValueError("Layer not found: %s.  Please pull again the image" % layer.replace("sha256:", ""))
+
+                self._checkout_layer(repo, rootfs_fd, rootfs, rev_layer)
+            finally:
+                if rootfs_fd:
+                    os.close(rootfs_fd)
+
+        return layers_dir
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -864,11 +864,14 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             return ATOMIC_VAR
 
     def get_ostree_repo_location(self):
+        location = os.environ.get("ATOMIC_OSTREE_REPO")
+        if location is not None:
+            return location
+
         if self.user:
-            return "%s/.containers/repo" % HOME
+            return  "%s/.containers/repo" % HOME
         else:
-            return os.environ.get("ATOMIC_OSTREE_REPO") or \
-                self.get_atomic_config_item(["ostree_repository"]) or \
+            return self.get_atomic_config_item(["ostree_repository"]) or \
                 "/ostree/repo"
 
     def _get_ostree_repo(self):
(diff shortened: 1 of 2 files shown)


giuseppe:
CC @yuqi-zhang @ashcrow

ashcrow:
```Running test_system_containers.sh....           FAIL```

**Comment by ashcrow** [comment]
I'll do a review of the changes this afternoon my time.

File: Atomic/syscontainers.py [hunk 1]
```diff
@@ -872,11 +866,14 @@ def _get_system_checkout_path(self):
             return ATOMIC_VAR
 
     def get_ostree_repo_location(self):
+        location = os.environ.get("ATOMIC_OSTREE_REPO")
+        if location is not None:
+            return location
+
         if self.user:
-            return "%s/.containers/repo" % HOME
+            return  "%s/.containers/repo" % HOME
```
TomSweeneyRedHat: nit, extra space added after the return here

Atomic/syscontainers.py
```diff
@@ -872,11 +866,14 @@ def _get_system_checkout_path(self):
...
         else:
```
TomSweeneyRedHat: Nit.  The else isn't necessary.  You might want to consider dropping it and just doing the bare return here.

Atomic/syscontainers.py
```diff
@@ -1783,3 +1782,63 @@ def get_image_branch(img):
...
+        for i in os.listdir(storage):
+            rev_layer = repo.resolve_rev("%s%s" % (OSTREE_OCIIMAGE_PREFIX, i), True)[1]
+            if not rev_layer:
+                shutil.rmtree(os.path.join(storage, i))
+
+
+    def mount_from_storage(self, img, destination):
+        repo = self._get_ostree_repo()
+        if not repo:
+            raise ValueError("OSTree not supported")
```
TomSweeneyRedHat: would it be useful to also include the img in the error here?

giuseppe: I think this error is more generic, it checks whether OSTree storage is present or not.  It doesn't matter what image is used

TomSweeneyRedHat: Your call.  You're right the img isn't used looking up the repo, I just wondered if it would be helpful to know which img it was about to look up when things went south.

Atomic/syscontainers.py
```diff
@@ -583,6 +576,20 @@ def _amend_values(self, values, manifest, name, image, image_id, destination, pr
...
+        # system repo. It is needed to create the hard links.
+        if self.get_ostree_repo_location() != '/ostree/repo':
+            return destination
+        try:
+            sysroot = OSTree.Sysroot()
+            sysroot.load()
+            osname = sysroot.get_booted_deployment().get_osname()
+            destination = os.path.realpath(os.path.join("/ostree/deploy/", osname, os.path.relpath(destination, "/")))
+        except: #pylint: disable=bare-except
+            pass
```
ashcrow: nit: Seems like there should be a log line here (unless having ``OSTree.Sysroot()``or ``get_booted_deployment().get_osname()`` raising are not "important" errors.

Atomic/syscontainers.py
```diff
@@ -1354,6 +1353,8 @@ def prune_ostree_images(self):
         refs = {}
         app_refs = []
 
+        self.prune_storage(repo)
```
ashcrow: I feel like pruning when getting the repo might be a surprise. I realize this is an internal callable so it may not be a big deal.

Atomic/syscontainers.py
```diff
@@ -1783,3 +1784,66 @@ def get_image_branch(img):
...
+            rootfs = os.path.sep.join([storage_path, layer])
+            layers_dir.append(rootfs)
+            if os.path.exists(rootfs):
+                continue
+            os.makedirs(rootfs)
+            rootfs_fd = None
+            try:
+                rootfs_fd = os.open(rootfs, os.O_DIRECTORY)
+
+                rev_layer = repo.resolve_rev("%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer.replace("sha256:", "")), True)[1]
```
ashcrow: Not required, but ``.format()`` reads nicer than ``%()``.

Atomic/syscontainers.py
```diff
@@ -1783,3 +1784,66 @@ def get_image_branch(img):
...
+
+
+    def prune_storage(self, repo):
+        storage = self.get_storage_path()
+        if not os.path.exists(storage):
+            return
+        for i in os.listdir(storage):
+            rev_layer = repo.resolve_rev("%s%s" % (OSTREE_OCIIMAGE_PREFIX, i), True)[1]
+            if not rev_layer:
+                shutil.rmtree(os.path.join(storage, i))
```
ashcrow: A ``try``/``except`` may make sense here.

Atomic/syscontainers.py
giuseppe: the idea here is that it fails then we use the original destination (as we are not on Atomic Host)

Atomic/syscontainers.py
giuseppe: prune must be done explicitly with "atomic images prune".  I changed the function to be not public though

Atomic/syscontainers.py
giuseppe: done

Atomic/syscontainers.py
giuseppe: what portion would you like to enclose in the try/catch?  Shouldn't exceptions here be propagated?

giuseppe:
I addressed all the reported issues, anything more to do on this series?

giuseppe:
@rhatdan fine to merge this?

Atomic/syscontainers.py
rhatdan: @giuseppe can you fix this.

Atomic/syscontainers.py
rhatdan: Fix it.

rhatdan:
Fix the nits from @TomSweeneyRedHat  and then you can merge.

giuseppe:
I've done that in the last version of the PR

rhatdan:
@rh-atomic-bot r+ 	543167a

rhatdan:
@rh-atomic-bot r+ 335f076

rhatdan:
@rh-atomic-bot retry

Issue #943, opened 2017-03-21
pull --storage ostree results into "Not a regular file or symlink: loop1 (0)"


TomasTomecek:
This is basically a copy of https://github.com/projectatomic/atomic/pull/767#issuecomment-287818865.

Since I can reproduce with latest release of `atomic`, I am reporting as a new issue.

Environment is Fedora rawhide:

```
$ rpm -q docker atomic ostree
docker-1.13.1-4.git5be1549.fc27.x86_64
atomic-1.15.3-2.fc26.x86_64
ostree-2017.2-3.fc26.x86_64
```

The issue:

```
$ atomic --debug pull --storage ostree docker:modularitycontainers/nginx
Namespace(_class=<class 'Atomic.pull.Pull'>, assumeyes=False, debug=True, func='pull_image', image='docker:modularitycontainers/nginx', reg_type=None, storage='ostree')
g-io-error-quark: Not a regular file or symlink: loop1 (0)
Traceback (most recent call last):
  File "/usr/bin/atomic", line 188, in <module>
    sys.exit(_func())
  File "/usr/lib/python3.6/site-packages/Atomic/pull.py", line 53, in pull_image
    be.pull_image(self.args.image, remote_image_obj, debug=self.args.debug, assumeyes=self.args.assumeyes)
  File "/usr/lib/python3.6/site-packages/Atomic/backends/_ostree.py", line 115, in pull_image
    return self.syscontainers.pull_image(image)
  File "/usr/lib/python3.6/site-packages/Atomic/syscontainers.py", line 154, in pull_image
    self._pull_image_to_ostree(self._get_ostree_repo(), image or self.args.image, True)
  File "/usr/lib/python3.6/site-packages/Atomic/syscontainers.py", line 145, in _pull_image_to_ostree
    image = self._pull_docker_image(repo, image.replace("docker:", "", 1))
  File "/usr/lib/python3.6/site-packages/Atomic/syscontainers.py", line 1111, in _pull_docker_image
    return self._pull_docker_tar(repo, temptar.name, image)
  File "/usr/lib/python3.6/site-packages/Atomic/syscontainers.py", line 1132, in _pull_docker_tar
    self._pull_dockertar_layers(repo, imagebranch, temp_dir, input_layers, labels=labels)
  File "/usr/lib/python3.6/site-packages/Atomic/syscontainers.py", line 1298, in _pull_dockertar_layers
[...]

PR #980, opened 2017-04-26
syscontainers: do not commit special files

giuseppe:
## Description

Do not try to commit special files with OSTree.

## Related Issue Numbers
https://github.com/projectatomic/atomic/issues/943


baude:
@rh-atomic-bot r+  	7fd6f28

PR #982, opened 2017-04-26
bug: Escape %config in rpm_host_install.py

ashcrow:
## Description
When installing a system container the following error occurs:

```
%c requires int or char
```

## Related Bugzillas
None

## Related Issue Numbers
None


ashcrow:
cc @giuseppe

giuseppe:
@rh-atomic-bot r+  	396faff

PR #983, opened 2017-04-26
Refactor syscontainer tests

yuqi-zhang:
Current state of patch: most working, with commented-out error cases

The old tests (test_system_containers.sh) were disorganzied
and lacked coverage for many commands. This refactor attempts
to improve the tests with a focus on clarity and coverage.

Improvements:
  - split into 4 separate tests for clarity and independence
  - section each test to better know where failures happen
  - make use of shared setup functions
  - use ${WORK_DIR} for test files
  - better setup/teardown usage
  - improved coverage for update/rollback, image commands,
    environment variables and expected failures

Drawback:
  - Slower

Remaining Issues:
  - 'atomic.type'=system not tested (local installs do not recognize this label)
  - deleting a system image while its being used for a container causes lots of issues
  - is_dangling does not seem to work for system containers
  - verify code is not working for system container images, (docker id to ostree ref discrepancy)
  - if container update fails, a new checkout is still made but not fully populated
  - image list --no-trunc and --quiet does not work together

Future considerations:
  - images will not work with new runc (spec 1.0.0 RC3)
  - test for more system container failures in general


yuqi-zhang:
@giuseppe @baude PTAL

yuqi-zhang:
@giuseppe I think the tests are ready. For the individual cases I have commented out, I will fix the issues first as separate PR's.

Only warning is that these new tests are kinda slow (about 15 minutes to complete)

giuseppe:
@yuqi-zhang good improvements, just two questions:

1) do you know why it is failing on Fedora sometimes and we need to comment out that test?
2) is there anything we can do to make this faster somehow?

If possible, could you please amend together the patches that fix changes introduced in the same series?

yuqi-zhang:
@giuseppe 
1. I'm not sure. I was unable to reproduce that locally, but basically the idea behind that case was using the same port (8081) for both containers should have caused a failure, but it doesn't on fedora-cloud. I will look into this later
2. We could, either by shaving some test cases or putting more tests into one file. (the setup takes a long time since **pulling** is the real killer) Honestly I think keeping the tests separate helps a lot with locating specific failures, so if you want I can try to get rid of some test cases that are not very high priority?
3. I will squash and rebase

giuseppe:
no it is fine, have different tests in different files is a good thing.  We can optimize this part later, perhaps skipping the pull when we can

yuqi-zhang:
@giuseppe I've rebased. If you want you can merge this PR and I will fix the highlighted issues as a separate PR, since the tests are working as they stand.

giuseppe:
@rh-atomic-bot r+ ea2f0b3

yuqi-zhang:
@giuseppe could you trigger the r+ again? I want to see if it fails.

giuseppe:
@rh-atomic-bot retry

yuqi-zhang:
@rh-atomic-bot retry

giuseppe:
@rh-atomic-bot r+ 557af02

jlebon:
@rh-atomic-bot retry

PR #984, opened 2017-04-27
syscontainers: Fix rename with directories

giuseppe:
## Description

We saw some issues while using `renameFiles` with directories.  This PR addresses these problems


ashcrow:
Pulled locally to verify it renamed directories and it does properly.

giuseppe:

Since it is useful for the Docker container, if the series is fine for you, I think we can merge it

ashcrow:
@rh-atomic-bot r+

giuseppe:
it seems that Homu hangs when the commit is not specified, trying again:

@rh-atomic-bot r=ashcrow 57971cb

PR #994, opened 2017-05-05
resolves #993 fixes file bind mounts

### LINKED ISSUE #993: atomic install with system container fails with non directory bind mount
Currently in syscontainers.py when doing a _do_checkout missing bind mounts are evaluated via the config.json, unfortunately the method that creates the tmp files is creating all files with a directory (d) label.

The problem line is here:
https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L754
```
missing_bind_paths = self._check_oci_configuration_file(destination_path, remote_path, True)

...
...

if "mounts" in configuration:
            for mount in configuration["mounts"]:
                if not "type" in mount:
                    continue
                if "source" in mount and "bind" in mount["type"]:
                    source = mount["source"]
                    if include_all or not os.path.exists(source):
                        missing_source_paths.append(source)
        return missing_source_paths
```
The True designation causes the evaluated json to create the file with all directory designations:

```
d    /run/systemd   0700 0 0 - -
d    /etc/adjtime   0700 0 0 - -
d    /etc/hosts   0700 0 0 - -
d    /tmp   0700 0 0 - -
d    /etc/sysconfig   0700 0 0 - -
d    /etc/resolv.conf   0700 0 0 - -
d    /etc/shadow   0700 0 0 - -
d    /sys/fs/cgroup   0700 0 0 - -
d    /proc   0700 0 0 - -
d    /etc/hostname   0700 0 0 - -
d    /sysroot   0700 0 0 - -
d    /etc/passwd   0700 0 0 - -
d    /usr/share/zoneinfo   0700 0 0 - -
d    /var/lib/sss/pipes   0700 0 0 - -
d    /etc/nsswitch.conf   0700 0 0 - -
d    /var/log   0700 0 0 - -
d    /   0700 0 0 - -
d    /dev/rtc0   0700 0 0 - -
```

I will create a PR to change this.

dav1x:
## Description
Currently in syscontainers.py when doing a _do_checkout missing bind mounts are evaluated via the config.json, unfortunately the method that creates the tmp files is creating all files with a directory (d) label.

The problem line is here:
https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L754

## Related Bugzillas
- Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1448567

## Related Issue Numbers
- Fixes #993


yuqi-zhang:
Makes sense. I feel like that the `include_all` flag doesn't really make much sense in the first place, since when do we want to brand all mounts as a "missing" path? @giuseppe 

@dav1x thanks for the heads up. Creating a tmpfiles.template in the image can be used as a workaround for now if this is urgent, since it won't generate a tmpfiles.conf if a template exists.

giuseppe:
bot, test pull request

giuseppe:
I am tempted even to drop this feature now that we export files from the container.  In any case, this fixes a regression so let's start with this:

@rh-atomic-bot r+  	2d5a5c4

dav1x:
Thanks all!

PR #997, opened 2017-05-08
Re-pulling image error should exit 0

baude:
## Description
Bugzilla #1430708 recommends that if an atomic user attempts to pull
an image that is already present, we should not exit with a '1' which
indicates a failure; rather a 0.

## Related Bugzillas
- Bugzilla #1430708
-

## Related Issue Numbers
-
-

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -328,7 +328,7 @@ class DockerBackend(Backend):
         local_image = self.has_image(image)
         if local_image is not None:
             if self.already_has_image(local_image, remote_image_obj):
-                raise ValueError("Latest version of {} already present.".format(image))
+                raise util.ImageAlreadyExists(image)
         registry, _, _, tag, _ = util.Decompose(fq_name).all
         image = "docker-daemon:{}".format(fq_name)
         if not image.endswith(tag):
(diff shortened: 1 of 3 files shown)


Atomic/backends/_docker.py
```diff
@@ -328,7 +329,8 @@ def pull_image(self, image, remote_image_obj, **kwargs):
         local_image = self.has_image(image)
         if local_image is not None:
             if self.already_has_image(local_image, remote_image_obj):
-                raise ValueError("Latest version of {} already present.".format(image))
+                util.write_err("Latest version of {} already present.".format(image))
```
rhatdan: Should we create an exception that handles this, rather then writing the code here.

Something like

raise SuccessException("Latest version of {} already present.".format(image))
￼
That way it could be handled potentially by something like atomic-dbus or other callers.

TomSweeneyRedHat:
There's a comment near line 400 where the self.pull_image is called from that will need to be tweaked based on the final change.

Atomic/backends/_docker.py
TomSweeneyRedHat: Maybe I'm misunderstanding the desired end game here, but I thought you'd just replace the old 'raise ValueError' call here with a 'return 0' then let the self_inspect fire in the update call as a double (triple) check.

baude:
@rh-atomic-bot r+  	46dc65e

baude:
Merging so I can cut a new release of Atomic.

Issue #870, opened 2017-02-06
requiring active docker daemon for 'atomic images info'?


miabbott:
Similar to #869 

I only have an image pulled to ostree storage.  But I can't get the LABEL info of it unless docker is running.

```
# git branch
* master
# git rev-parse HEAD
dd182c19cd8f31ec60a577a5429ab80a1cb435fb
# ./atomic images list
   REPOSITORY       TAG      IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
   gscrivano/etcd   latest   54e7e6cbd01d   2017-02-06 19:33                  ostree    

# systemctl is-active docker
inactive
# ./atomic images info gscrivano/etcd
This Atomic function requires an active docker daemon.
# systemctl start docker
# docker images -a
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
# ./atomic images info gscrivano/etcd
Image Name: docker.io/gscrivano/etcd:latest


Template variables with default value, but overridable with --set:
ETCD_ADVERTISE_CLIENT_URLS: 
ETCD_CERT_FILE: 
ETCD_CLIENT_CERT_AUTH: false
ETCD_CORS: 
ETCD_DEBUG: false
ETCD_DISCOVERY: 
ETCD_DISCOVERY_FALLBACK: proxy
ETCD_DISCOVERY_PROXY: 
ETCD_DISCOVERY_SRV: 
ETCD_ELECTION_TIMEOUT: 1000
ETCD_ENABLE_PPROF: false
ETCD_HEARTBEAT_INTERVAL: 100
ETCD_INITIAL_ADVERTISE_PEER_URLS: 
ETCD_INITIAL_CLUSTER: 
ETCD_INITIAL_CLUSTER_STATE: new
ETCD_INITIAL_CLUSTER_TOKEN: etcd-cluster
ETCD_KEY_FILE: 
ETCD_LISTEN_CLIENT_URLS: 
ETCD_LISTEN_PEER_URLS: 
ETCD_LOG_PACKAGE_LEVELS: 
ETCD_MAX_SNAPSHOTS: 5
ETCD_MAX_WALS: 5
ETCD_NAME: 
ETCD_PEER_CERT_FILE: 
ETCD_PEER_CLIENT_CERT_AUTH: false
ETCD_PEER_KEY_FILE: 
ETCD_PEER_TRUSTED_CA_FILE: 
ETCD_PROXY: off
[...]

PR #1003, opened 2017-05-15
Incorporate registries parsing tool

baude:
## Description
Ideally atomic will be able to run multiple container
runtimes.  In and effort to do so, we must have a concept
of a global registries configuration file which will allow
us to pull images with atomic and skopeo. The tooling
that parses the global configuration file is:

https://github.com/projectatomic/registries

This is step one in the implementation of this tooling
and direction.

## Related Issue Numbers
- #870 
-

## Pull Request Checklist:

- [x] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -40,6 +40,11 @@ BWRAP_OCI_PATH = "/usr/bin/bwrap-oci"
 RUNC_PATH = "/bin/runc"
 SKOPEO_PATH = "/usr/bin/skopeo"
 
+try:
+    from subprocess import DEVNULL  # pylint: disable=no-name-in-module
+except ImportError:
+    DEVNULL = open(os.devnull, 'wb')
+
 def gomtree_available():
     return os.path.exists(GOMTREE_PATH)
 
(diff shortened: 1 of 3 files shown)


Atomic/util.py
```diff
@@ -63,22 +63,61 @@ def get_docker_conf():
         dconf = c.info()
     return dconf
 
+
+def registries_tool_path():
+    if os.path.exists("/usr/libexec/registries"):
```
rhatdan: Can you make this a constant that is available via /etc/atomic.conf.   Would like to see us able to change this on the fly especially for testing.

Atomic/util.py
```diff
@@ -63,22 +63,61 @@ def get_docker_conf():
...
+        return "/usr/libexec/registries"
+    try:
+        return subprocess.check_output(['which', 'registries'])
```
rhatdan: Add --skip-alias

Atomic/util.py
```diff
@@ -63,22 +63,61 @@ def get_docker_conf():
...
+        rconf = dconf['RegistryConfig']['IndexConfigs']
+        # docker.io is special
+        if 'docker.io' in rconf:
+            registries.append({'hostname': 'registry-1.docker.io', 'name': 'docker.io', 'search': True, 'secure': True})
+            # remove docker.io
+            del(rconf['docker.io'])
+        for i in rconf:
+            search_bool = True if i in search_regs else False
+            registries.append({'hostname': i, 'name': i, 'search': search_bool, 'secure': rconf[i]['Secure'] })
+    elif registries_tool_path() is not None:
```
rhatdan: Could I have a bogus line in docker conf that has no regististys, and I would still execure registires_too_path?  Or is this valid to have no registries defined.

Atomic/util.py
```diff
@@ -63,22 +63,61 @@ def get_docker_conf():
...
+        registries_json = load_registries_from_yaml()
+        _registries = registries_json.get("registries", [])
+        _insecure_registries = registries_json.get('insecure_registries', [])
+        _blocked_registries = registries_json.get('block_registries', [])
+        registries = [{'search': True, 'hostname': x, 'name': x, 'secure': True} for x in _registries if x not in _blocked_registries]
+        registries += [{'search': True, 'hostname': x, 'name': x, 'secure': True} for x in _insecure_registries if x not in _blocked_registries]
```
rhatdan: Should we be checking to see if we have duplicate registries?

Atomic/util.py
```diff
@@ -63,22 +63,61 @@ def get_docker_conf():
...
+        if 'docker.io' not in _registries and 'docker.io' not in _blocked_registries:  # Always add docker.io unless blocked
+            registries.append({'hostname': 'registry-1.docker.io', 'name': 'docker.io', 'search': True, 'secure': True})
```
rhatdan: Sad...

tests/unit/test_registries.py
```diff
@@ -0,0 +1,81 @@
...
+        if len(results) != len(answer):
+            raise AssertionError("Length of the lists differ.")
+        for registry in results:
+            self.assertDictEqual(registry, next(item for item in answer if item["hostname"] == registry['hostname']))
+
+    def test_no_registries(self):
+        with patch('Atomic.util.is_backend_available') as dockerd:
+            dockerd.return_value = False
+            with patch('Atomic.util.registries_tool_path') as reg_path:
+                reg_path.return_value = "/usr/libexec/registries"
```
rhatdan: Is registries going to be shipped by a different package or by atomic?  Should we test the local registries executable versus the one installed in /usr/libexec/registies

Atomic/util.py
baude: Sure ...

Atomic/util.py
baude: It is considered valid and is the default configuration; and then docker.io is always injected by the binaries.

tests/unit/test_registries.py
baude: It will be a subpackage of atomic last we spoke about it.  I had planned to exercise it via an integration test but not until it is in the spec.  These unit tests just test the path through get_registries to make sure the outputs don't regress.

Atomic/util.py
baude: great idea

baude: done

Atomic/util.py
baude: done

Atomic/util.py
baude: ... but true

Atomic/util.py
baude: if there is a dup between registries (secure) and insecure_registries, we drop a ValueError

baude:
@rhatdan updated to answer or resolve your comments ...

Atomic/util.py
rhatdan: If we have registries.conf, we should remove this as the default.  Otherwise registries will never work.

TomSweeneyRedHat:
LGTM

yuqi-zhang:
Can confirm this fixes system container install/info issues. One minor thing is that it outputs `which: no registries in (/sbin:/bin:/usr/sbin:/usr/bin)` multiple times when I run a command, not sure if that is intended.

baude:
stderr is now --> devnull

yuqi-zhang:
Can confirm that the messages are gone

baude:
@rhatdan i think we're good

rhatdan:
@baude I think the registries.conf always wins in the latest patch.  I was thinking this should be

atomic.conf wins if the ADMIN wrote anything in there.  By default there is nothing.
If there are no registries specified in atomic.conf, then registries works.
If there are no registries specified anywhere then docker.io wins.

Does the user need to specify docker.io to win, or do we just set this up by default in registries.conf, so it is easier to remove the name?

rhatdan:
@rh-atomic-bot r+ 6dbf951

jlebon:
@rh-atomic-bot retry

Issue #981, opened 2017-04-26
syscontainers: Avoid requiring rpmbuild


cgwalters:
See http://pkgs.fedoraproject.org/cgit/rpms/atomic.git/commit/?id=8c4a38a7f187c45926bdf60ae937f7a6090cbeae
:arrow_right: https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2017-April/msg00059.html

I haven't looked, but offhand, I suspect it won't be that hard to construct a binary RPM without doing a "build".

giuseppe:
we allow containers to provide a customized `.spec` file.  Should we limit this for now and allow only the arbitrary copy of files to the host?  We could probably take the metadata for dependencies/conflicts (which is probably the only part that is really useful) from the image labels as we already do in case a spec file is not specified.

@rhatdan @ashcrow @TomasTomecek

TomasTomecek:
I propose to raise this issue with Software Management team. I personally dislike constructing binary RPMs using some hacky method - this could lead into some unpredictable behavior (does RPM software keep binary RPMs compatible?).

baude:
how about building from within a container?

giuseppe:
we would require to create a container everytime we want to install a container that exports files, I think this feature should be lower level than that and be better tied to the system.  I like the idea of Colin to have a `lightweight rpmbuild` that only packs these files into an .rpm.  We probably don't need to support the build from a `.spec` file part we have now and only a small subset.

baude:
Given that atomic packages would appear broken with the removal of rpmbuild, can we come up with something quickly? I'm willing to help if others are limited on time.

giuseppe:
the best short term fix is probably to not require rpmbuild and not create the rpm when it is not present, in this case we can still copy files to the host but we will not track them with an .rpm file (which is already implemented as fallback). What do you think?

rhatdan:
Can we open up a discussion with the rpmbuild maintainers to see if we can get the lightweight rpmbuild.

rhatdan:
I agree with the fallback to no rpmbuild for now, when it is not present.

giuseppe:
I played a bit with it and this looks the minimal implementation I could get for building an RPM out of a tree directory:

https://gist.github.com/giuseppe/43cdb1e86f3042c4bcfd25cb68a249af

It still miss some pieces we need, but it is already able to build an rpm that `rpm -i` accepts

giuseppe:
[...]

PR #1005, opened 2017-05-17
Replace rpmbuild dependency with new rpmwriter module

giuseppe:
## Description

Replace rpmbuild dependency with a simpler module to write directly an RPM file.

We lose the capability of building an RPM from a spec file, but this was really an over-engineered solution, so better drop it before it is too late.

Users who want a more complex RPM file, then they must provide it as part of the image itself '/exports/container.rpm'

## Related Issue Numbers

https://github.com/projectatomic/atomic/issues/981

### DIFF
--- a/Atomic/rpm_host_install.py
+++ b/Atomic/rpm_host_install.py
@@ -1,5 +1,7 @@
+import re
 import os
 from . import util
+from . import rpmwriter
 import tempfile
 import shutil
 
@@ -68,53 +70,10 @@ class RPMHostInstall(object):
 
         return new_installed_files
 
-    @staticmethod
-    def _generate_spec_file(destdir, name, summary, license_, image_id, version="1.0", release="1", url=None,
-                            source0=None, requires=None, conflicts=None, provides=None, description=None,
-                            installed_files=None, include_containers_file=True):
-        spec = "%global __requires_exclude_from ^.*$\n"
-        spec = spec + "%global __provides_exclude_from ^.*$\n"
-        spec = spec + "%define _unpackaged_files_terminate_build 0\n"
-
-        fields = {"Name" : "%s-%s" % (RPM_NAME_PREFIX, name), "Version" : version, "Release" : release, "Summary" : summary,
-                  "License" : license_, "URL" : url, "Source0" : source0, "Requires" : requires,
-                  "Provides" : provides, "Conflicts" : conflicts}
-        for k, v in fields.items():
-            if v is not None:
-                spec = spec + "%s:\t%s\n" % (k, v)
-
-        spec = spec + ("\n%%description\nImage ID: %s\n" % image_id)
-        if description:
-            spec = spec + "%s\n" % description
-
-        spec = spec + "\n%files\n"
-        installed_files_in_etc = []
-        for root, _, files in os.walk(os.path.join(destdir, "etc")):
-            rel_path = os.path.relpath(root, destdir)
-            for f in files:
-                p = os.path.join("/", rel_path, f)
-                installed_files_in_etc.append(p)
-                spec += "%%config \"%s\"\n" % p
-
-        if include_containers_file:
-            spec += "/usr/lib/containers/atomic/%s\n" % name
-        for root, _, files in os.walk(os.path.join(destdir, "usr/lib/systemd/system")):
-            for f in files:
-                spec = spec + "/usr/lib/systemd/system/%s\n" % f
-        for root, _, files in os.walk(os.path.join(destdir, "usr/lib/tmpfiles.d")):
-            for f in files:
-                spec = spec + "/usr/lib/tmpfiles.d/%s\n" % f
-        if installed_files:
-            for i in installed_files:
-                if i not in installed_files_in_etc:
-                    spec = spec + "%s\n" % i
-
-        return spec
 
     @staticmethod
     def generate_rpm_from_rootfs(rootfs, temp_dir, name, image_id, labels, include_containers_file, display=False, installed_files=None, defaultversion='1'):
         rpm_content = os.path.join(temp_dir, "rpmroot")
-        spec_file = os.path.join(temp_dir, "container.spec")
 
         included_rpm = os.path.join(rootfs, "rootfs", "exports", "container.rpm")
         if os.path.exists(included_rpm):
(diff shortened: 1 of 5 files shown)


giuseppe:
@cgwalters @baude this is the WIP for replacing rpmbuild, I need to add tests

giuseppe:
I've adapted the tests to work with the new module.  There are quite a few tests to cover the rpm generation. Dropping the WIP and ready for review

giuseppe:
The CI failure is unrelated to the PR

ashcrow:
```
+ '[' /var/mnt /sysroot/ostree/deploy/centos-atomic-host/var/mnt ']'
./tests/integration/test_storage.sh: line 39: [: /var/mnt: unary operator expected
+ wipefs -a /dev/vdb
```

ashcrow:
Related: https://github.com/ostreedev/ostree/pull/859

baude:
@giuseppe assuming https://github.com/projectatomic/atomic/pull/1008 merges, just rebase and you should be good.

giuseppe:
rebased on master so that the storage test is disabled ⬆️

cgwalters:
How did you come up with `rpmwriter.py`?  Did you just reverse-engineer librpm and translate to Python?  A few comments about how you did it would help maintenance down the line.

Down the line...I'd like to change things so that the "synthesize an RPM" code lives in `rpm-ostree` in that case.  Basically have atomic pass the metadata as JSON/GVariant/whatever  + path to rootfs.

For now though this seems pretty sane to me.

giuseppe:
@cgwalters I've used the documentation here as a starting point: http://ftp.rpm.org/max-rpm/s1-rpm-file-format-rpm-file-format.html

I got stuck a few times so I added more debug messages to `rpm` that helped me, but in general the file format documentation + [rpmtag.h](https://github.com/rpm-software-management/rpm/blob/master/lib/rpmtag.h) was enough to write `rpmwriter.h`.  I've added a fixup patch to add more info about it

Atomic/rpmwriter.py
```diff
@@ -0,0 +1,369 @@
...
+    RPMTAG_FILELINKTOS = 1036
+    RPMTAG_FILEFLAGS = 1037
+    RPMTAG_FILEUSERNAME = 1039
+    RPMTAG_FILEGROUPNAME = 1040
+
+    RPMTAG_PROVIDENAME		= 1047
+    RPMTAG_REQUIRENAME		= 1049
+    RPMTAG_REQUIREVERSION	= 1050
+    RPMTAG_CONFLICTFLAGS	= 1053
+    RPMTAG_CONFLICTNAME		= 1054
```
TomSweeneyRedHat: micro nit.  You've an extra tab it looks like between the variable and the equals for a few of these definitions.    I personally like this style and would ask that you do the same for the group of variables above too (lines 40->62)

Atomic/rpmwriter.py
```diff
@@ -0,0 +1,369 @@
...
+        self.license_ = license_
+        self.changelog = changelog
+        self.url = url
+        self.group = group
+        self.stderr = stderr
+        self.whitelist = set(whitelist) if whitelist is not None else None
+
+    def add_header(self, tag, typ, count, value, pad=1):
+        try:
+            # this fails on python3
```
TomSweeneyRedHat: Should a TODO comment be added to fix in python3?

giuseppe: no, this is the desired behavior.  The try inhibits the exception to be raised as there is no `unicode` in python3

Atomic/rpmwriter.py
```diff
@@ -0,0 +1,369 @@
+import os
```
TomSweeneyRedHat: Do you need to have a shebang here declaring the version of python you're supporting?  Perhaps:

#!/usr/bin/env python2

Given your python3 issue noted below?

giuseppe: atomic must support both versions.  On RHEL we have python2, on Fedora python3

TomSweeneyRedHat: Sounds good, I misunderstood your comment on the python3 failure.  Given that it would stlll be good to have some kind of shebang statement at the top of the file.  Perhaps:

#!/usr/bin/env python

But certainly not a comment to hold the merge for.

Atomic/rpmwriter.py
TomSweeneyRedHat: Ah, I read it to mean you were having unexpected issues here.  Perhaps change the comment to something like:

"The next line is expected to always fail under python3, in that case just continue."

or something similar.  Another nit, not a hold the merge comment.

Atomic/rpmwriter.py
giuseppe: I've pushed a fixup patch to use the same style for all variables

Atomic/rpmwriter.py
giuseppe: done

Atomic/rpm_host_install.py
```diff
@@ -68,96 +70,60 @@ def rm_add_files_to_host(old_installed_files, exports, prefix="/", files_templat
...
-               "--build-in-place",
-               "--buildroot=%s" % rpm_content]
-        util.write_out(" ".join(cmd))
-        if not display:
-            util.check_call(cmd)
+
+        rpm_name = "atomic-container-%s" % name
+        rpm_out = os.path.join(result_dir, "%s.rpm" % rpm_name)
+        def split_name_version(pkg):
+            r = r"([abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789\-\._\+]+)(.*)"
```
ashcrow: nit: This regex could be shortened to ```([a-zA-Z0-9_\-\.\+]+)(.*)```

Atomic/rpm_host_install.py
```diff
@@ -199,10 +165,10 @@ def generate_rpm(name, image_id, labels, exports, destination, values=None, inst
             rpm_root = RPMHostInstall.generate_rpm_from_rootfs(destination, temp_dir, name, image_id, labels, include_containers_file=False, installed_files=installed_files, defaultversion=defaultversion)
             rpm_file = RPMHostInstall.find_rpm(rpm_root)
             if rpm_file:
-                dest_path = os.path.join(destination, "container.rpm")
+                orig_name = "atomic-container-%s.rpm" % name
```
ashcrow: nit: I recommend using ``{}`` and ``format`` with string formatting unless there is a need to hold compatibility with very old Python versions.

Atomic/rpmwriter.py
```diff
@@ -0,0 +1,369 @@
...
+    def add_provide(self, name):
+        self.provide.append(name)
+
+    def add_obsolete(self, name):
+        self.obsolete.append(name)
+
+    def add_conflict(self, name, version):
+        self.conflict.append([name, version])
+
+    def __init__(self, out, root, name, version, release, summary='', description='', license_='gpl2', changelog='', url='', group='', stderr=sys.stderr, whitelist=None):
```
ashcrow: I ❤️ GPL but why default to gpl2 if no license is provided?

giuseppe: rpm complains if no license is specified, so I've just used the same license of `atomic` itself

Atomic/rpm_host_install.py
giuseppe: thanks, pushed a fixup patch.

Atomic/rpm_host_install.py
giuseppe: change included in the fixup patch

TomSweeneyRedHat:
LGTM, assuming happy test results.  Thanks for the nit touch ups.

giuseppe:
@rhatdan what do you think of this change?

rhatdan:
@giuseppe Can this handle dependencies?  IE Can I build a "spec" file that says it requires "foobar" package be installed?

giuseppe:
@rhatdan the rpmwriter doesn't use a spec file to build the `.rpm` file. It creates the cpio archive given a `rootfs` and the metadata is read from labels and injected in the `.rpm`:

`LABEL atomic.requires=foo>1,bar<2`

We support: `summary`, `version`, `release`, `license`, `url`, `requires`, `provides`, `conflicts`, `description`.

I'll document this once the PR looks fine.

@TomasTomecek

Atomic/rpmwriter.py
TomasTomecek: What's the point of having shebang if the file is not meant to be run directly?

Atomic/rpmwriter.py
```diff
@@ -0,0 +1,369 @@
...
+
+# More information on the RPM file format can be found here:
+# http://ftp.rpm.org/max-rpm/s1-rpm-file-format-rpm-file-format.html
+# The same documentation was used to initially write this module.
+#
+# Only tags that are actually used in the code were added to the RpmWriter class.
+# The complete list is available here:
+# https://github.com/rpm-software-management/rpm/blob/master/lib/rpmtag.h
+
+class RpmWriter(object):
```
TomasTomecek: My opinion is that this module should be a separate project. If you look at it, it's not tight to `atomic` in any way. I think that plenty of other users and projects could benefit from a lightweight and friendler user-experience interaction with rpm builds. Even further, it would be awesome if rpm team reviewed this piece of code and ack'd it.

Atomic/rpmwriter.py
```diff
@@ -0,0 +1,369 @@
...
+        self._writebytearray(self._make_uint32(len(store)))
+        self._writebytearray(header_section)
+        self._writebytearray(store)
+
+    def _payload(self, out):
+        uncompressed_size = 0
+        def try_read(stdout, gzip_out):
+            written = 0
+            while True:
+                readable, _, _ = select.select([stdout], [], [], 0)
```
TomasTomecek: Wow, is `select` needed here? I would assume that simple pipe with reads and writes is just enough.

Atomic/rpmwriter.py
cgwalters: We're not *building* here though.  More like just repacking binaries.  What other use cases are there where one isn't using a spec file?  Or without wanting a dependency on `rpmbuild`?

Atomic/rpmwriter.py
```diff
@@ -0,0 +1,369 @@
...
+import struct
+import subprocess
+import select
+import gzip
+import hashlib
+import tempfile
+import shutil
+
+# More information on the RPM file format can be found here:
+# http://ftp.rpm.org/max-rpm/s1-rpm-file-format-rpm-file-format.html
```
TomasTomecek: Is this spec backwards and forward compatible?

Atomic/rpmwriter.py
giuseppe: I want it to not block on `_try_read`, just read as much as it is available.  In this way I can feed more data to the cpio process through `cpio_process.stdin.write(filename.encode() + b"\n")`

Atomic/rpmwriter.py
giuseppe: I've created a separate repository here: https://github.com/giuseppe/tinyrpmbuild, if someone is really interested in it.

I agree with @cgwalters, this is not a replacement for `rpmbuild`, it is really implementing the minimum necessary to get an installable `.rpm` file.  Not sure how many other users will have the same requirement of building an `.rpm` on the fly without using `rpmbuild`.

baude:
@rh-atomic-bot r+  	8537474

baude:
@rh-atomic-bot r+ 271ebdb

giuseppe:
@baude thanks for the r+, I've manually fixed the conflicts with the fixup patches.  Pushed a new version ⬆️

rhatdan:
@rh-atomic-bot retry

rhatdan:
@rh-atomic-bot r+  	be51f9c

PR #1008, opened 2017-05-18
tests/integration/test_storage.sh

baude:
## Description
Use findmnt to detect mounts.

### DIFF
--- a/tests/integration/test_storage.sh
+++ b/tests/integration/test_storage.sh
@@ -35,10 +35,7 @@ setup () {
     # Perform setup routines here.
     copy /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage-setup.atomic-tests-backup
     TEST_DEV_1=/dev/vdb
-    MNT=$(mount | awk '$1 ~/vdb/' | awk '{print $3}')
-    if [ ${MNT} ]; then
-	    umount $MNT
-    fi
+    findmnt -o SOURCE | grep "^$TEST_DEV_1" | uniq | xargs -r umount
     wipefs -a "$TEST_DEV_1"
     TEST_DEV_1_pvs=${TEST_DEV_1}1
 


ashcrow:
```
+ '[' TARGET SOURCE FSTYPE OPTIONS /var/mnt /dev/vdb ext4 rw,relatime,seclabel,data=ordered /sysroot/ostree/deploy/centos-atomic-host/var/mnt /dev/vdb ext4 rw,relatime,seclabel,data=ordered ']'
./tests/integration/test_storage.sh: line 38: [: too many arguments
+ wipefs -a /dev/vdb
```

jlebon:
So this is interesting. To be helpful, the ephemeral disk that OpenStack attaches already has a filesystem on it. E.g. `/dev/vdb` *is* an ext4 partition, which is rather unusual. So maybe what we can do here is, in the YAML, do a `umount /dev/vdb`. And then in the `env` stanza export e.g. a `TEST_STORAGE_DEV=/dev/vdb` env var, which gets picked up in `test_storage.sh`. That way, local devs can set their own block device there.

And then to make sure that all partitions on the device are unmounted, do something like (untested):

```sh
findmnt -o SOURCE | grep "^$TEST_STORAGE_DEV" | xargs umount
```

ashcrow:
Getting closer:

```
umount: /dev/vdb: not mounted
```

tests/integration/test_storage.sh
```diff
@@ -35,10 +35,13 @@ setup () {
     # Perform setup routines here.
     copy /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage-setup.atomic-tests-backup
     TEST_DEV_1=/dev/vdb
```
jlebon: What I mean is, this should be defined externally rather than hardcoded. It's not guaranteed to be `/dev/vdb` on every machine where this test is going to run.

baude: i guess that is the point of the destructive env variable

tests/integration/test_storage.sh
```diff
@@ -35,10 +35,14 @@ setup () {
...
-	    umount $MNT
-    fi
+    mount
+    cat /etc/fstab
+    findmnt -A
+    #if [ $(findmnt "$TEST_DEV_1") ]; then 
+#	    umount $MNT
+    #fi
+    findmnt -n $TEST_DEV_1 -o TARGET | xargs umount
+    #findmnt -o SOURCE | grep "^$TEST_DEV_1" | xargs umount
```
jlebon: Try something like:

```
findmnt -o SOURCE | grep "^$TEST_DEV_1" | uniq | xargs -r umount
```

baude:
@rh-atomic-bot r+  	17dec2d

PR #1004, opened 2017-05-17
atomic.conf: fix syntax error of the YAML format

chuanchang:
We use YAML format in atomic.conf file, so it should be KEY: VALUE
not KEY=VALUE.

Signed-off-by: Alex Jia <ajia@redhat.com>

## Related Bugzillas
-
-

## Related Issue Numbers
-
-

## Pull Request Checklist:

If your Pull request contains new features or functions, tests are required. If the PR is a bug fix and no tests exist, please consider adding some to prevent regressions.
- [ ] Unittests
- [ ] Integration Tests

PR #1009, opened 2017-05-22
atomic.conf test test and unit template updates

ashcrow:
## Description
- Updates the unit test template to be in line with other tests and encourage documentation
- Checks the ``atomic.conf`` file to ensure it's valid YAML

**Note** This test is currently expected to fail as the current atomic.conf is NOT valid YAML when the proxy items are checked.

## Related Issue Numbers
- #1004

### DIFF
--- a/tests/unit/test_conf.py
+++ b/tests/unit/test_conf.py
@@ -1,0 +1,46 @@
+#pylint: skip-file
+import os
+import unittest
+
+import yaml
+
+
+class TestAtomicUnit(unittest.TestCase):
+    """
+    Checks that the configuration file provided is valid.
+    """
+
+    def setUp(self):
+        """
+        Provide the conf file contents freshly for each test.
+        """
+        self.conf = open(os.path.sep.join(['atomic.conf']), 'r')
+
+    def tearDown(self):
+        """
+        Close the config file after every test.
+        """
+        self.conf.close()
+
+    def test_config_file_is_valid_yaml(self):
+        """
+        Verifies atomic.conf is valid YAML.
+        """
+        self.assertEquals(type(yaml.safe_load(self.conf)), dict)
+
+    def test_config_file_is_valid_yaml_with_items_uncommented(self):
+        """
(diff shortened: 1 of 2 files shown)


tests/unit/test_conf.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def setUp(self):
+        """
+        Provide the conf file contents freshly for each test.
+        """
+        self.conf = open(os.path.sep.join(['..', '..', 'atomic.conf']), 'r')
+
+    def tearDown(self):
+        """
+        Close the config file after ever test.
```
TomSweeneyRedHat: nit, "ever test" to "every test"

ashcrow: Woops!

tests/unit/test_conf.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def test_config_file_is_valid_yaml_with_items_uncommented(self):
+        """
+        Verifies atomic.conf is valid YAML when examples are uncommented.
+        """
+        # If a command line has a space after it, it's a comment
+        # If a comment line has no space after it, it's an example
+        uncommented = []
+        for line in self.conf.readlines():
+            if line.startswith('#') and len(line) > 2 and line[1] != ' ':
```
TomSweeneyRedHat: Maybe not enough tea yet today, but should it be checking "> 1" rather than "> 2"?   Otherwise wouldn't '#A' get skipped?

ashcrow: The test is just to make sure the line isn't ``#\n`` before checking the second element for a space.

ashcrow:
Looks like ``atomic.conf`` is not where I expect it to be in redhat-ci. Will update.

ashcrow:
Fixed.

ashcrow:
Rebased

ashcrow:
@rhatdan / @baude PTAL

rhatdan:
LGTM

giuseppe:
@rh-atomic-bot r+ 924617e

PR #1016, opened 2017-05-25
atomic containers list json fix

ashcrow:
## Description
The ``atomic containers list --json`` returns nothing when no containers will be returned. This causes some parsers, such as Python's ``json`` module, to fail reading the output. This change does the following:

When ``atomic containers list --json`` is used...
- empty results are returned as an empty json list: ``[]``
- and ``-q`` is also used, the quiet is honored over the ``--json`` request

## Related Bugzillas
- https://bugzilla.redhat.com/show_bug.cgi?id=1450307

## Related Issue Numbers
- https://github.com/openshift/openshift-ansible/pull/4272#issuecomment-304075181

## Pull Request Checklist:

- [x] Integration Tests

### DIFF
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -134,11 +134,14 @@ class Containers(Atomic):
 
 
         container_objects = self._ps()
-        if len(container_objects) == 0:
-            return 0
+        # If we were not asked for json output, return out with no output
+        # when there are no applicable container objects
+        if not self.args.json:
+            if len(container_objects) == 0:
+                return 0
 
-        if not any([x.running for x in container_objects]) and not self.args.all:
-            return 0
+            if not any([x.running for x in container_objects]) and not self.args.all:
+                return 0
 
         max_container_id = 12 if self.args.truncate else max([len(x.id) for x in container_objects])
         if self.args.quiet:
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -143,7 +143,14 @@ class Containers(Atomic):
             if not any([x.running for x in container_objects]) and not self.args.all:
                 return 0
 
-        max_container_id = 12 if self.args.truncate else max([len(x.id) for x in container_objects])
+        # Set to 12 when truncate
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@baude PTAL

Atomic/containers.py
```diff
@@ -134,13 +134,23 @@ def ps_tty(self):
...
+            if len(container_objects) == 0:
+                return 0
 
-        if not any([x.running for x in container_objects]) and not self.args.all:
-            return 0
+            if not any([x.running for x in container_objects]) and not self.args.all:
+                return 0
+
+        # Set to 12 when truncate
+        if self.args.truncate:
```
baude: Whats the purpose of expanding the short form of the expression previously used? Did it trigger a failure or doing for readability?

tests/integration/test_containers_list.sh
```diff
@@ -24,3 +24,7 @@ diff docker.ps.out atomic.ps.out
 ${ATOMIC} containers list -q -f runtime=Docker | sort > atomic.ps.out
 docker ps -q | sort > docker.ps.out
 diff docker.ps.out atomic.ps.out
+
+# Ensure that when json is requested and no containers are returned we still
+# get valid json ([])
+${ATOMIC} containers list --json -f container=idonotexist | grep "\[\]"
```
baude: should there be a comparison here to ensure the result is correct?

Atomic/containers.py
ashcrow: Readability was the primary reason. I felt adding the ```or [0]``` was one operation too many for a one liner.

tests/integration/test_containers_list.sh
ashcrow: If I understand the test scripts correctly this should catch improper return results. If the result itsn't an empty list then grep should return a non zero result failing the test.

baude:
@rh-atomic-bot r+  	1a40ea9

Issue #1006, opened 2017-05-17
'atomic run' using image with different tags returns error


miabbott:
Using git master (3c0cbc69eafffc6f9e192ebdaf48c05517ef130f)

1. I've pulled `fedora:25` and `fedora:26` from the Fedora registry
1. Run the `date` command using `fedora:25`
1. Run the `date` command using `fedora:26`
1. Blows up

```
# ./atomic images list
   REPOSITORY                          TAG  IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
   registry.fedoraproject.org/fedora   25   40e6be3d1688   2017-05-12 05:48   230.85 MB      docker    
   registry.fedoraproject.org/fedora   26   57ae3032ccae   2016-12-15 17:44   211.51 MB      docker    

# ./atomic run registry.fedoraproject.org/fedora:25 date
docker run -t -i --name fedora registry.fedoraproject.org/fedora:25 date
Wed May 17 18:34:12 UTC 2017

# ./atomic containers list --all
   CONTAINER ID IMAGE                COMMAND              CREATED          STATE      BACKEND    RUNTIME   
   d9f2626df2ad registry.fedoraproje date                 2017-05-17 18:34 exited     docker     docker    

# docker ps -a
CONTAINER ID        IMAGE                                  COMMAND             CREATED             STATUS                      PORTS               NAMES
d9f2626df2ad        registry.fedoraproject.org/fedora:25   "date"              18 seconds ago      Exited (0) 17 seconds ago                       fedora

# ./atomic run registry.fedoraproject.org/fedora:26 date                                                                                                                            
fedora
rpc error: code = 13 desc = invalid header field value "oci runtime error: exec failed: container \"d9f2626df2adcc2139b65e729c72700134452cb677ee8817265120611c6ce337\" does not exist\n"

# ./atomic --debug run registry.fedoraproject.org/fedora:26 date                                                                                                                    
fedora
rpc error: code = 13 desc = invalid header field value "oci runtime error: exec failed: container \"d9f2626df2adcc2139b65e729c72700134452cb677ee8817265120611c6ce337\" does not exist\n"

Traceback (most recent call last):
  File "./atomic", line 203, in <module>
    sys.exit(_func())
  File "/root/atomic/Atomic/run.py", line 87, in run
    return be.run(con_obj, atomic=self, args=self.args)
  File "/root/atomic/Atomic/backends/_docker.py", line 538, in run
    return self._start(iobject, args, atomic)
[...]

PR #1019, opened 2017-05-30
Atomic/backends/_docker.py: Error prevention with atomic run

baude:
## Description

There were two primary cases where a secondary atomic run with a command
would trigger an exception.  The first was reported in
https://github.com/projectatomic/atomic/issues/1006. Basically it can
be summarized as:

```
atomic run registry.fedoraproject.org/fedora:25 date  # works fine
atomic run registry.fedoraproject.org/fedora:26 date  # tries to run in the existing f25 container
```

The second case is as simple as:

```
atomic run registry.fedoraproject.org/fedora:25 date  # works fine
atomic run registry.fedoraproject.org/fedora:25 date  # fails
```

This fails because atomic starts the stopped f25 container and then attempts a docker exec.  The
exec fails because the 'date' command is short-lived and the container exits prior to the exec
being run.

We now catch those exceptions and notify the user.  We added a `--replace` option to run where
atomic will now delete the container in question and re-run it from the correct image.

## Related Bugzillas
-
-

## Related Issue Numbers
- #1006 
-

## Pull Request Checklist:

- [ ] Unittests
- [x] Integration Tests

### DIFF
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -18,6 +18,8 @@ from Atomic import Atomic
 from requests.exceptions import HTTPError
 from Atomic.backends._docker_errors import NoDockerDaemon
 from Atomic.discovery import RegistryInspectError
+from Atomic.atomic import AtomicError
+from subprocess import CalledProcessError
 
 try:
     from subprocess import DEVNULL  # pylint: disable=no-name-in-module
@@ -516,6 +518,16 @@ class DockerBackend(Backend):
             list_item += value
             return list_item
 
+        def replace_existing_container(_iobject, _requested_image, _args):
+            if _args.debug:
+                util.write_out("Removing the container {} and running with {}".format(_iobject.name,
+                                                                                      requested_image.fq_name))
+            self.delete_container(_iobject.id, force=True)
+            _iobject = _requested_image
+            if _args.command:
+                _iobject.user_command = _args.command
+            return _iobject
+
         atomic = kwargs.get('atomic', None)
         args = kwargs.get('args')
         # atomic must be an instance of Atomic
@@ -537,10 +549,37 @@ class DockerBackend(Backend):
                                "{}s\n\n removes all containers based on an "
                                "image.".format(iobject.name, iobject.image_name, iobject.name, iobject.name,
                                                iobject.image_name, iobject.image_name, iobject.image_name))
+
+            requested_image = self.has_image(args.image)
+            if requested_image is None:
+                requested_image = self.has_image(iobject.image)
+
             if iobject.running:
+                if args.replace:
+                    iobject = replace_existing_container(iobject, requested_image, args)
                 return self._running(iobject, args, atomic)
             else:
-                return self._start(iobject, args, atomic)
+                # Container with the name exists
+                image_id = iobject.image
+                if requested_image.id != image_id:
+                    if args.replace:
+                        iobject = replace_existing_container(iobject, requested_image, args)
+                    else:
+                        try:
+                            requested_image_fq_name = requested_image.fq_name
+                        except RegistryInspectError:
+                            requested_image_fq_name = args.image
+                        raise AtomicError("Warning: container '{}' already points to {}\nRun 'atomic run {}' to run "
+                                          "the existing container.\nRun 'atomic run --replace '{}' to replace "
+                                          "it".format(iobject.name,
+                                                      iobject.original_structure['Config']['Image'],
+                                                      iobject.name,
+                                                      requested_image_fq_name))
+                else:
+                    if args.replace:
+                        iobject = replace_existing_container(iobject, requested_image, args)
+                    else:
+                        return self._start(iobject, args, atomic)
 
         if iobject.get_label('INSTALL') and not args.ignore and not util.InstallData.image_installed(iobject):
             raise ValueError("The image '{}' appears to have not been installed and has an INSTALL label.  You "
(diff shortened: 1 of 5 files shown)


miabbott:
This change handles the cases mentioned above and the `--replace` flag is a great addition, IMO.

I think there may still be a problem with some of the logic.  It appears to have been a problem even before these changes, so this isn't a regression, but something that might be able to be handled here.

The hypothetical scenario is one where you have two images that share the same `image:tag` but are from different registries.  In my example below, I synthesized an image via `docker build`:

```
# cat strace/Dockerfile 
FROM registry.fedoraproject.org/fedora:25
RUN dnf -y install strace
# docker build -t myregistry.example.com/fedora:25 strace/
...
Successfully built d12d5e10bd8d
# ./atomic images list
   REPOSITORY                          TAG  IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
   myregistry.example.com/fedora       25   d12d5e10bd8d   2017-05-30 19:33   412.14 MB      docker    
   registry.fedoraproject.org/fedora   25   6c0c98f53e1e   2017-05-23 05:49   231.87 MB      docker
```

Now, when I `atomic run` the official Fedora image with the `date` command, the resulting container gets used for both images because they share the same `image:tag`.

```
# ./atomic run registry.fedoraproject.org/fedora:25 date
docker run -t -i --name fedora registry.fedoraproject.org/fedora:25 date
Tue May 30 19:34:26 UTC 2017
# ./atomic run registry.fedoraproject.org/fedora:25
Tue May 30 19:34:30 UTC 2017
# ./atomic run myregistry.example.com/fedora:25
Tue May 30 19:34:56 UTC 2017
```

I think `atomic` should be doing some more introspection to determine if the `atomic run` command is really asking for the same `image:tag` and therefore can re-use the container.

WDYT?

Atomic/run.py
```diff
@@ -45,6 +45,9 @@ def cli(subparser):
     runp.set_defaults(_class=Run, func='run')
     run_group = runp.add_mutually_exclusive_group()
     util.add_opt(runp)
+    runp.add_argument("--replace", "-r", dest="replace", default=False,
+                      action="store_true", help=_("Replaces an existing container by the same name"
+                                                  "if it exists"))
```
TomSweeneyRedHat: micro nit, missing period.  Ditto line 66

Atomic/backends/_docker.py
```diff
@@ -537,10 +549,31 @@ def add_string_or_list_to_list(list_item, value):
...
             if iobject.running:
+                if args.replace:
+                    iobject = delete_existing_container(iobject, requested_image, args)
                 return self._running(iobject, args, atomic)
             else:
-                return self._start(iobject, args, atomic)
+                # Container with the name exists
+                image_id = iobject.image
+                # See if we can find an image with the input name
+                if requested_image.id != image_id:
```
TomSweeneyRedHat: Ok, I've not had my first cup of tea today, so maybe messing up.  But the comment says we're looking for a match but the if is doing a not match.  Shouldn't this be '==' rather than '!='?

Atomic/run.py
baude: Fixed.

Atomic/backends/_docker.py
```diff
@@ -537,10 +549,34 @@ def add_string_or_list_to_list(list_item, value):
                                "{}s\n\n removes all containers based on an "
                                "image.".format(iobject.name, iobject.image_name, iobject.name, iobject.name,
                                                iobject.image_name, iobject.image_name, iobject.image_name))
+
+            requested_image = self.has_image(args.image)
             if iobject.running:
+                if args.replace:
+                    iobject = delete_existing_container(iobject, requested_image, args)
```
rhatdan: Should we call this replace_existing_container?

delete_* returning something new seems weird.

Atomic/backends/_docker.py
```diff
@@ -662,13 +698,15 @@ def _running(self, con_obj, args, atomic):
...
                 container_command = args.command if isinstance(args.command, list) else args.command.split()
-                return util.check_call(
-                    [atomic.docker_binary(), "exec", "-t", "-i", con_obj.name] +
-                    container_command)
+                try:
+                    return util.check_call([atomic.docker_binary(), "exec", "-t", "-i", con_obj.name] + container_command)
+                except CalledProcessError as e:
+                    if args.debug:
+                        util.write_out(str(e))
+                    raise AtomicError("Unable to execute your command.  Try re-running your command with the --replace"
```
rhatdan: I think you should add some more info here.  This error should only happen if the exec failed because the container is not running.
"Failed to execute the command inside the existing container.  In some situations this can happen because the entry point command of the container only runs for a short time.  You might want to replace the container by executing your command with --replace.  Note any updates to the existing container will be lost."

Atomic/backends/_docker.py
baude: Ok done, it doesnt actually replace the container but I think the intent is clearer.

Atomic/backends/_docker.py
baude: Fixed.

baude: done

baude:
Added a fix for @miabbott 's report as well ... it was a bug in the new code

miabbott:
I kicked this PR around on a Fedora 25 system and I think this is good to merge.

baude:
@rh-atomic-bot r+ 209f89c

PR #1027, opened 2017-06-07
Add configuration compliance scan to "atomic scan"

jan-cerny:
## Description
This PR enables scanning images and containers for configuration compliance with security profiles provided by SCAP Security Guide.

Note: This feature requires latest OpenSCAP Daemon from upstream installed in the underlying "rhel7/openscap" container.

## Related Bugzillas
- none

## Related Issue Numbers
- none

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -12,14 +12,15 @@ def cli(subparser):
     # atomic scan
     scanners = util.get_scanners()
     scanp = subparser.add_parser(
-        "scan", help=_("scan an image or container for CVEs"),
-        epilog="atomic scan <input> scans a container or image for CVEs")
+        "scan", help=_("scan an image or container for CVEs or configuration compliance"),
+        epilog="atomic scan <input> scans a container or image for CVEs or configuration compliance")
     scanp.set_defaults(_class=Scan, func='scan')
     scan_group = scanp.add_mutually_exclusive_group()
     scanp.add_argument("scan_targets", nargs='*', help=_("container image"))
     scanp.add_argument("--scanner", choices=[x['scanner_name'] for x in scanners], default=None, help=_("define the intended scanner"))
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scan type"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
+    scanp.add_argument("--scanner_args", default=None, help=_("Specify arguments to be passed to the scanner"))
     disp_group = scanp.add_mutually_exclusive_group()
     disp_group.add_argument("--verbose", action='store_true', default=False, help=_("Show more output from scanning container"))
     disp_group.add_argument("--json", action='store_true', default=False, help=_("Output results in JSON format"))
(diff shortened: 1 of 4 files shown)


rhatdan:
bot, add author to whitelist

Atomic/scan.py
```diff
@@ -12,14 +12,15 @@ def cli(subparser):
...
-        epilog="atomic scan <input> scans a container or image for CVEs")
+        "scan", help=_("scan an image or container for CVEs or configuration compliance"),
+        epilog="atomic scan <input> scans a container or image for CVEs or configuration compliance")
     scanp.set_defaults(_class=Scan, func='scan')
     scan_group = scanp.add_mutually_exclusive_group()
     scanp.add_argument("scan_targets", nargs='*', help=_("container image"))
     scanp.add_argument("--scanner", choices=[x['scanner_name'] for x in scanners], default=None, help=_("define the intended scanner"))
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scan type"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
+    scanp.add_argument("--profile_id", default=None, help=_("Select a SCAP Security Guide profile"))
```
rhatdan: Can we make this more generaal.  IE Remove SCAP.   
Option should just be --profile, I don't see any reason for _id

Atomic/scan.py
```diff
@@ -96,6 +97,9 @@ def set_scanner():
 
         scan_type = self.get_scan_type()
 
+        if self.args.profile_id is not None and scan_type != "configuration_compliance":
```
rhatdan: Should this type of intelligence be left to the scanning tool rather then hard coded in this way.  Should we allow different scanners to handle profiles differently?

rhatdan:
Missing command completions

Atomic/scan.py
jan-cerny: OK, I have added `_id` to distinguish the option from existing `atomic --profile` which is used for profiling. If this collision is OK, I also prefer just `--profile`.

Atomic/scan.py
jan-cerny: This check is intended to remind user that he specifies a profile for configuration compliance scan, but he's probably using CVE scan. It is not necessary to do here, so yes, I can leave the decision up to openscap daemon.

Atomic/scan.py
rhatdan: I will let @baude comment.    Is this nothing more then an args that gets passed to the scanner?  I am looking for a way to make this not OpenScap scanner specific and allow other scanners to pass in  additional args to their scanners.

jan-cerny: Yes, I just need a way how to pass an argument to the scanner.

If we want this to be more generic, we can have something like `--scanner_args` instead. That would be used e.g. `--scanner_args "profile=stig"`

rhatdan: Yes I think that would work better for the general case.

jan-cerny:
I have amended the commit. I can't figure out why the test failed, it seems that it just didn't have enough time. If this is caused by my commit, please let me know.

baude:
bot, retest this please.

baude:
bot, retest this please

baude:
@rh-atomic-bot r+  	7322e53

PR #1033, opened 2017-06-22
  Handle image deletion for used images and unnamed layer images 

peterbaouoft:
Before, when an image is referenced by a container, user is still
able to delete the image even without force command

To avoid that thing to happen, another column is added
in the delete message(shown below):

 IMAGE                     STORAGE       USED
 xxxxx                         xxxxxxx           xxxx

When user is trying to delete an image,
the USED column will have a YES or NO answer
to indicate the stauts of the image

The user will no longer be able to delete
"used" images(without force)
and an error message will be
shown to further indicate that.

Some tests are added for future regression.

## Related Bugzillas
-
-

## Related Issue Numbers
-
-

## Pull Request Checklist:

- [ ] Unittests
- [x] Integration Tests

### DIFF
--- a/Atomic/backends/_ostree.py
+++ b/Atomic/backends/_ostree.py
@@ -172,7 +172,8 @@ class OSTreeBackend(Backend):
 
     @staticmethod
     def get_dangling_images():
-        return []
+        syscontainers = SystemContainers()
+        return syscontainers.get_dangling_system_images()
 
     def make_remote_image(self, image):
         img_obj = self._make_remote_image(image)
--- a/Atomic/backends/_ostree.py
+++ b/Atomic/backends/_ostree.py
@@ -112,7 +112,29 @@ class OSTreeBackend(Backend):
     def pull_image(self, image, remote_image_obj=None, **kwargs):
         return self.syscontainers.pull_image(image)
 
+    # get containers that is using this image
+    def get_container_from_image(self,image):
+
+        # We get the system containers for its id and image
+        all_container_images = [(container.image,container.id) for container in self.get_containers()]
+
+        matching_container_ids = [container_id for (image_id, container_id) in all_container_images if image_id == image]
+
+        return matching_container_ids
+
     def delete_image(self, image, force=False):
+
+        # when force option is true, delete the containers associated with it
+        if force:
+            # to handle the case where partial id or image name was typed
+            img_obj = self.has_image(image)
+
+            if img_obj:
+                used_containers = self.get_container_from_image(img_obj.id)
+
+                for container in used_containers:
+                    self.delete_container(container)
+
         return self.syscontainers.delete_image(image)
 
     def version(self, image):
--- a/Atomic/backends/_ostree.py
+++ b/Atomic/backends/_ostree.py
@@ -112,25 +112,18 @@ class OSTreeBackend(Backend):
     def pull_image(self, image, remote_image_obj=None, **kwargs):
         return self.syscontainers.pull_image(image)
 
-    # get containers that is using this image
-    def get_container_from_image(self,image):
 
-        # We get the system containers for its id and image
-        all_container_images = [(container.image,container.id) for container in self.get_containers()]
-
-        matching_container_ids = [container_id for (image_id, container_id) in all_container_images if image_id == image]
-
+    def get_corresponding_containers_from_image_id (self,image_id):
+        all_container_and_image_ids = [(container.image,container.id) for container in self.get_containers()]
+        matching_container_ids = [container_id for (container_image_id, container_id) in all_container_and_image_ids if container_image_id == image_id]
         return matching_container_ids
 
     def delete_image(self, image, force=False):
-
-        # when force option is true, delete the containers associated with it
         if force:
             # to handle the case where partial id or image name was typed
             img_obj = self.has_image(image)
-
             if img_obj:
-                used_containers = self.get_container_from_image(img_obj.id)
+                used_containers = self.get_corresponding_containers_from_image_id(img_obj.id)
 
                 for container in used_containers:
                     self.delete_container(container)
@@ -170,10 +163,8 @@ class OSTreeBackend(Backend):
             layers.append(layer)
         return layers
 
-    @staticmethod
-    def get_dangling_images():
-        syscontainers = SystemContainers()
-        return syscontainers.get_dangling_system_images()
+    def get_dangling_images(self):
+        return self.syscontainers.get_dangling_system_images()
 
     def make_remote_image(self, image):
         img_obj = self._make_remote_image(image)
--- a/Atomic/backendutils.py
+++ b/Atomic/backendutils.py
@@ -156,11 +156,9 @@ class BackendUtils(object):
             con_objs += be.get_containers()
         return con_objs
 
-    def get_container_ids_from_image(self,image):
-        # We get the system containers for its id and image
+    def get_container_ids_from_image_id(self,image_id):
         all_container_images = [(container.image,container.id) for container in self.get_containers()]
-
-        matching_container_ids = [container_id for (image_id, container_id) in all_container_images if image_id == image]
+        matching_container_ids = [container_id for (container_image_id, container_id) in all_container_images if container_image_id == image_id]
         return matching_container_ids
 
     def get_container_obj_by_image_name(self, image_name, str_preferred_backend):
--- a/Atomic/backendutils.py
+++ b/Atomic/backendutils.py
@@ -156,6 +156,13 @@ class BackendUtils(object):
             con_objs += be.get_containers()
         return con_objs
 
+    def get_container_ids_from_image(self,image):
+        # We get the system containers for its id and image
+        all_container_images = [(container.image,container.id) for container in self.get_containers()]
+
+        matching_container_ids = [container_id for (image_id, container_id) in all_container_images if image_id == image]
+        return matching_container_ids
+
     def get_container_obj_by_image_name(self, image_name, str_preferred_backend):
         pass
 
--- a/Atomic/delete.py
+++ b/Atomic/delete.py
@@ -49,10 +49,12 @@ class Delete(Atomic):
         _image_names = []
         for del_obj in delete_objects:
             if del_obj.repotags:
-                _image_names.append(len(del_obj.repotags[0]))
+                length = 12 if "<none>" in del_obj.repotags[0] else len(del_obj.repotags[0])
+                _image_names.append(length)
             else:
                 _image_names.append(len(del_obj.id))
         max_img_name = max(_image_names) + 2
+
 
         # When found used imags, we stop action of trying to delete it
         # we then add it to a new list which will be processed
@@ -83,7 +85,7 @@ class Delete(Atomic):
                 used_images.append(del_obj)
             util.write_out(three_col.format(image, del_obj.backend.backend, image_status))
 
-        # always elminate out the unnamed not dangling layer
+        # always disable the intention that tries to delete unnamed layers which are not dangling
         delete_objects = [deletable for deletable in delete_objects if deletable not in unnamed_not_dangling_imgs]
 
         if not self.args.force:
@@ -105,7 +107,7 @@ class Delete(Atomic):
             # for the used images, output the error messages
             for used_img in used_images:
                 util.write_out("Error: conflict: unable to remove repository reference \"{}\"(must-force) - container(s) {} is using its referenced image {}".format(used_img.input_name,
-                                beu.get_container_ids_from_image(used_img.id),used_img.id[0:12]))
+                                beu.get_container_ids_from_image_id(used_img.id),used_img.id[0:12]))
 
         for undangling_layer in unnamed_not_dangling_imgs:
             util.write_out("Image {} is an unnamed layer of an image".format(undangling_layer.id[0:12]))
--- a/Atomic/delete.py
+++ b/Atomic/delete.py
@@ -69,15 +69,22 @@ class Delete(Atomic):
 
         three_col = "   {0:" + str(max_img_name) + "} {1:12}  {2}"
         util.write_out(three_col.format("IMAGE", "STORAGE", "USED"))
+
+        unnamed_not_dangling_imgs = []
         for del_obj in delete_objects:
             image = None if not del_obj.repotags else del_obj.repotags[0]
             image_status  = 'YES' if del_obj.used else 'NO'
             if image is None or "<none>" in image:
+                if not del_obj.is_dangling:
+                    unnamed_not_dangling_imgs.append(del_obj)
                 image = del_obj.id[0:12]
 
             if del_obj.used and not self.args.force:
                 used_images.append(del_obj)
             util.write_out(three_col.format(image, del_obj.backend.backend, image_status))
+
+        # always elminate out the unnamed not dangling layer
+        delete_objects = [deletable for deletable in delete_objects if deletable not in unnamed_not_dangling_imgs]
 
         if not self.args.force:
             # redefine the objects that needed to be deleted
@@ -99,6 +106,10 @@ class Delete(Atomic):
             for used_img in used_images:
                 util.write_out("Error: conflict: unable to remove repository reference \"{}\"(must-force) - container(s) {} is using its referenced image {}".format(used_img.input_name,
                                 beu.get_container_ids_from_image(used_img.id),used_img.id[0:12]))
+
+        for undangling_layer in unnamed_not_dangling_imgs:
+            util.write_out("Image {} is an unnamed layer of an image".format(undangling_layer.id[0:12]))
+
         # We need to return something here for dbus
         return 0
 
--- a/Atomic/delete.py
+++ b/Atomic/delete.py
@@ -54,18 +54,35 @@ class Delete(Atomic):
                 _image_names.append(len(del_obj.id))
         max_img_name = max(_image_names) + 2
 
+        # When found used imags, we stop action of trying to delete it
+        # we then add it to a new list which will be processed
+        used_images = []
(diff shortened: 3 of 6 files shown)


peterbaouoft:
It will be greatly appreciated if some one can look at it  :D

Atomic/delete.py
```diff
@@ -54,18 +54,35 @@ def delete_image(self):
                 _image_names.append(len(del_obj.id))
         max_img_name = max(_image_names) + 2
 
+        #When found used imags, we stop action of trying to delete it
```
yuqi-zhang: Space between "#" and the first word, also "imags"

Atomic/delete.py
```diff
@@ -54,18 +54,35 @@ def delete_image(self):
...
         else:
             util.write_out("The following images will be deleted.\n")
 
-        two_col = "   {0:" + str(max_img_name) + "} {1}"
-        util.write_out(two_col.format("IMAGE", "STORAGE"))
+        three_col = "   {0:" + str(max_img_name) + "} {1:12}  {2}"
+        util.write_out(three_col.format("IMAGE", "STORAGE","USED"))
         for del_obj in delete_objects:
             image = None if not del_obj.repotags else del_obj.repotags[0]
+            imageStatus  = 'YES' if del_obj.used else 'NO'
```
yuqi-zhang: image_status

Atomic/delete.py
```diff
@@ -54,18 +54,35 @@ def delete_image(self):
...
             if image is None or "<none>" in image:
                 image = del_obj.id[0:12]
-            util.write_out(two_col.format(image, del_obj.backend.backend))
+
+            if del_obj.used:
+                used_images.append(del_obj)
+            util.write_out(three_col.format(image, del_obj.backend.backend, imageStatus))
+
+        if not self.args.force:
+                #redefine the objects that needed to be deleted
```
yuqi-zhang: Spacing, and tabbing

tests/integration/test_system_containers_images.sh
```diff
@@ -88,6 +93,13 @@ ${ATOMIC} images prune
 ${ATOMIC} images list -f type=ostree --all > ${WORK_DIR}/images.all.out
 assert_matches "<none>" ${WORK_DIR}/images.all.out
 
+#Test that when image is being used, it can not be deleted without force option
```
yuqi-zhang: Space

Atomic/delete.py
```diff
@@ -77,6 +94,11 @@ def delete_image(self):
         for del_obj in delete_objects:
             del_obj.backend.delete_image(del_obj.input_name, force=self.args.force)
 
+        if not self.args.force:
+            # for the used images, output the error messages
+            for used_img in used_images:
+                util.write_out("Image {} is being used by a container currently".format(used_img.id[0:12]))
```
yuqi-zhang: Maybe "Image {} is currently being used by container {}" and display the container and image name?

Atomic/delete.py
```diff
@@ -54,18 +54,35 @@ def delete_image(self):
...
+
         if not self.args.assumeyes:
             util.write_out("Do you wish to delete the following images?\n")
         else:
             util.write_out("The following images will be deleted.\n")
 
-        two_col = "   {0:" + str(max_img_name) + "} {1}"
-        util.write_out(two_col.format("IMAGE", "STORAGE"))
+        three_col = "   {0:" + str(max_img_name) + "} {1:12}  {2}"
+        util.write_out(three_col.format("IMAGE", "STORAGE","USED"))
```
yuqi-zhang: Spacing

peterbaouoft:
will fix the previous commit tommorw, the reason was due to improperly handled image id

peterbaouoft:
@yuqi-zhang , I made a new commit, take a look when you have time? Thanks in advance :D

tests/integration/test_system_containers_images.sh
```diff
@@ -78,7 +81,9 @@ assert_not_matches '<none>' ${WORK_DIR}/images.out
 
 
 # 3. deleting/pruning local images
-${ATOMIC} --assumeyes images delete -f --storage ostree busybox
+${ATOMIC} --assumeyes images delete -f --storage ostree busybox > ${WORK_DIR}/image.delete.out
+assert_matches "NO" ${WORK_DIR}/image.delete.out
+assert_matches "USED" ${WORK_DIR}/image.delete.out
```
yuqi-zhang: Why check for the header by itself? Maybe test that the correct column has "NO".

tests/integration/test_system_containers_images.sh
```diff
@@ -88,11 +93,37 @@ ${ATOMIC} images prune
...
+${ATOMIC} install --system -n ${CONTAINER_NAME} docker.io/busybox
+${ATOMIC} --assumeyes images delete --storage ostree docker.io/busybox > ${WORK_DIR}/image.delete.out
+assert_matches "YES" ${WORK_DIR}/image.delete.out
+assert_matches "unable to remove repository reference" ${WORK_DIR}/image.delete.out
+assert_matches ${CONTAINER_NAME} ${WORK_DIR}/image.delete.out
+assert_matches "docker.io/busybox" ${WORK_DIR}/image.delete.out
+${ATOMIC} --assumeyes containers delete ${CONTAINER_NAME}
+
+# Test that when the named layer is still present, we are unable to delete the unnamedlayer
+UNNAMED_LAYER_ID=$(${ATOMIC} images list -a -f type=ostree| grep '<none>' | awk '{print $3}')
```
yuqi-zhang: This is probably fine, but I still caution against doing this because you're assuming the busybox image will always have 1 unnamed layer.

Atomic/syscontainers.py
```diff
@@ -1369,6 +1370,53 @@ def uninstall(self, name):
...
+        except (IndexError, TypeError):
+            raise ValueError("Image {} not found".format(img))
+        manifest = self._image_manifest(repo, rev)
+        if not manifest:
+            return None
+        layers = SystemContainers.get_layers_from_manifest(json.loads(manifest))
+        return layers
+
+    # define a function to handle layer image checking
+    def get_dangling_system_images(self):
```
yuqi-zhang: I think for this function, the method in prune_ostree_images is a little better. Perhaps that code can be adapted to this function, and the prune_ostree_images could be built on top of this instead?

Also, please format this function properly.

Atomic/syscontainers.py
```diff
@@ -1369,6 +1370,53 @@ def uninstall(self, name):
...
+
+        if not named_layer_images:
+            return all_images
+
+        existing_img_id = []
+
+        # for loop to add each layered image into the list
+        for n_img  in named_layer_images:
+            existing_img_id.append(n_img)
+            # get exiting layers from named images
```
yuqi-zhang: exiting

Atomic/delete.py
```diff
@@ -54,18 +54,42 @@ def delete_image(self):
...
+
         if not self.args.assumeyes:
             util.write_out("Do you wish to delete the following images?\n")
         else:
             util.write_out("The following images will be deleted.\n")
 
-        two_col = "   {0:" + str(max_img_name) + "} {1}"
-        util.write_out(two_col.format("IMAGE", "STORAGE"))
+        three_col = "   {0:" + str(max_img_name) + "} {1:12}  {2}"
+        util.write_out(three_col.format("IMAGE", "STORAGE", "USED"))
```
yuqi-zhang: This does not seem to be aligned properly

Atomic/backends/_ostree.py
```diff
@@ -150,7 +172,8 @@ def get_layers(self, image):
 
     @staticmethod
     def get_dangling_images():
-        return []
+        syscontainers = SystemContainers()
```
peterbaouoft: Oh, the reason for that is cuz it is a static method, so I am assuming no self is needed?

Atomic/backendutils.py
```diff
@@ -156,6 +156,13 @@ def get_containers(self):
             con_objs += be.get_containers()
         return con_objs
 
+    def get_container_ids_from_image(self,image):
```
peterbaouoft: ah, both backendutils and syscontainers.py have this sort of similar function

peterbaouoft: I did that because we can get both docker images and ostree images from the backendutils

Atomic/syscontainers.py
peterbaouoft: Oh, I considered that as well, but that time I was not sure about which is better, whether resolve_image, then ask for manifest, or resolve rev then manifest. And yea, I have noticed the speed of running atomic images list become slower than before, I am currently "working" on it to sort of making it have speed as usual.

baude: @peterbaouoft please work with me on the speed issue... i have some thoughts and ideas too.

Atomic/syscontainers.py
```diff
@@ -1177,7 +1177,8 @@ def _inspect_system_branch(self, repo, imagebranch):
             imgs = self._resolve_image(repo, imagebranch, allow_multiple=True)
             if imgs is None:
                 raise ValueError("Image %s not found" % imagebranch)
-            _, commit_rev = imgs[0]
+            # replace the old branch with the new branch found
+            imagebranch, commit_rev = imgs[0]
```
giuseppe: can this change be reverted?  It seems to not be used

Atomic/backends/_ostree.py
giuseppe: looking at it, it seems the function should not be static.  Could you please make it non static and use `self.syscontainers` as @yuqi-zhang suggested?

Atomic/delete.py
```diff
@@ -54,18 +54,42 @@ def delete_image(self):
...
+                if not del_obj.is_dangling:
+                    unnamed_not_dangling_imgs.append(del_obj)
                 image = del_obj.id[0:12]
-            util.write_out(two_col.format(image, del_obj.backend.backend))
+
+            if del_obj.used and not self.args.force:
+                used_images.append(del_obj)
+            util.write_out(three_col.format(image, del_obj.backend.backend, image_status))
+
+        # always elminate out the unnamed not dangling layer
```
giuseppe: small typo here: "eliminate"

Atomic/syscontainers.py
```diff
@@ -1369,6 +1370,53 @@ def uninstall(self, name):
...
+            # get exiting layers from named images
+            existing_image_layers = self.get_layers_from_image(n_img,repo)
+
+            if not existing_image_layers:
+                continue
+
+            for layer in existing_image_layers:
+                existing_img_id.append(layer.replace("sha256:", ""))
+
+        # have a comparsion between the list of images and actual images from layers
```
giuseppe: typo: "comparsion"

Atomic/syscontainers.py
giuseppe: yes, I agree with @yuqi-zhang.  It seems that this function is doing the same thing as we already do in `prune_ostree_images`.  Could we refactor `prune_ostree_images` into two function `_check_used_layers` and `prune_ostree_images` that uses it?  Then we can use `_check_used_layers` here.

Alternatively you can refactor it directly in `get_dangling_system_images` and adapt `prune_ostree_images` to use it.

Atomic/backends/_ostree.py
```diff
@@ -112,7 +112,29 @@ def get_containers(self):
...
+        # when force option is true, delete the containers associated with it
+        if force:
+            # to handle the case where partial id or image name was typed
+            img_obj = self.has_image(image)
+
+            if img_obj:
+                used_containers = self.get_container_from_image(img_obj.id)
+
+                for container in used_containers:
+                    self.delete_container(container)
```
giuseppe: Do we want this behaviour (cc @rhatdan)?  A Docker container can still run when the image is removed and in the system containers case we can safely delete the image and the container works as it is checked out from the storage via hard links.  It will probably be better to just delegate this to each backend.  This is a dangerous operation and the user must know in advance the risks associated with it.

Atomic/syscontainers.py
peterbaouoft: Ah, the reason why I changed that image branch is because when we are deleting an unnamed image layer of an ociimage, if the user types in partial id(not the 64 characters id) into the command line, we can correctly identify the existence of an unnamed layer image, but the image id of the unnamed layer will be interpreted as the commit_rev( which can be seen in line 1180 ish in syscontainers.py in _inspect_system_branch function).
As a result, it leads to some of the information for the unnamed layer to be processed incorrectly. I am not too sure about whether this change would leave any side effects over, but fortunately I have not came across those at the current stage. So I am suggesting maybe we can leave it for now, if errors came along in the future, we can possibly change it back?

Atomic/backends/_ostree.py
peterbaouoft: Sure :D

Atomic/syscontainers.py
peterbaouoft: @baude sounds great :D, and I am feeling excited :p, may be we can have a discussion soon on IRC about this? I am not too skillful at the moment, but hopefully I can help out a bit. :D

peterbaouoft: and yep, it feels to me that refactor the function seems to work the best here, currently working on it :D

tests/integration/test_system_containers_images.sh
peterbaouoft: Good point, my intention was to put it there so people can see the USED tag was there without downloading the repo and see it :p

tests/integration/test_system_containers_images.sh
peterbaouoft: ah, I was just trying to test if deleting any unnamed layer will cause any potential problem, so my intention was not going for to test delete multiple unnamed image layers

peterbaouoft:
gg, seems like to be another infinite loop

Atomic/syscontainers.py
giuseppe: sure

Atomic/backends/_ostree.py
```diff
@@ -112,25 +112,18 @@ def get_containers(self):
...
 
-    # get containers that is using this image
-    def get_container_from_image(self,image):
-
-        # We get the system containers for its id and image
-        all_container_images = [(container.image,container.id) for container in self.get_containers()]
-
-        matching_container_ids = [container_id for (image_id, container_id) in all_container_images if image_id == image]
 
+    def get_corresponding_containers_from_image_id (self,image_id):
```
giuseppe: spacing here: self,image_id

Atomic/backendutils.py
```diff
@@ -156,11 +156,9 @@ def get_containers(self):
             con_objs += be.get_containers()
         return con_objs
 
-    def get_container_ids_from_image(self,image):
-        # We get the system containers for its id and image
+    def get_container_ids_from_image_id(self,image_id):
```
giuseppe: here too, spacing

Atomic/syscontainers.py
```diff
@@ -1370,57 +1370,26 @@ def uninstall(self, name):
...
-
-        # have a comparsion between the list of images and actual images from layers
-        dangling_images = [image_id for image_id in all_images if image_id not in existing_img_id]
-        return dangling_images
-
-    def prune_ostree_images(self):
-        repo = self._get_ostree_repo()
-        if not repo:
-            return
+    def _check_used_layers(self,repo):
```
giuseppe: here as well, no space after "self,"

Atomic/syscontainers.py
```diff
@@ -1370,57 +1370,26 @@ def uninstall(self, name):
...
-        # all images including the dangling versions
-        all_images = [x['Id'] for x in self.get_system_images(get_all=True)]
+        dangling_image_ids = []
+        dangling_refs = self._check_used_layers(repo)
 
-        if not named_layer_images:
-            return all_images
+	# each ref essentially can be seen as image branch
+        for dangling_image_branch in dangling_refs:
+            dangling_img = self._inspect_system_branch(repo,dangling_image_branch)
```
giuseppe: spacing issue here as well

giuseppe:
the code looks better now, except some spacing issues.  I am still not sure about this functionality though, do we really want to kill containers when we delete their image even if it is not required?

Should this perhaps be controlled with an option (something like `atomic images delete --destroy-containers IMAGE)?

rhatdan:
If we were to add this his, the option would be --force.  But I think we should just fail and say you must stop the containers.

peterbaouoft:
@rhatdan,may I ask one question for clarification with the force option? So, I am assuming that we will disallow users to delete the image when the containers is running(for both docker and system). However, I am unsure about how do we handle the "not running" system containers(with forced option), do we handle it similarly as  docker containers?

peterbaouoft:
I am suggesting that currently we just disallow user to delete the image( from ostree backend) whenever the image is used by a container?( no matter whether the container is running or exited) How does that sound?

giuseppe:
there is nothing preventing to delete the image even when it is used.  Since a container from the ostree storage is checked out using hard links it is perfectly fine to delete an image and its layers from the storage and still have a container that uses it.  The files of the image will be removed completely once there is no image in the storage and no containers using it.

peterbaouoft:
Ah, Jerry and I have talked about this issue earlier. The reason why I suggested to not allow the deletion of an used image is that once we delete the image, the next time when we are trying to update the container which is using the image, we will have an error saying the old image(deleted) is not found. 

As a result, me and Jerry sort of decided to not allow the image deletion when it is used by containers. If it is not needed, I can just revert it back to the original implementation( just not caring about deleting used images).

peterbaouoft:
Is it better that I write up an issue to summarize the problems I am trying to solve, and come back to further work on this problem later? I am fine with all suggestions :D. Thanks in advance

peterbaouoft:
close this since this has been too long

PR #1035, opened 2017-06-23
syscontainers: some improvements for the copy to the host case

giuseppe:
Print what files are copied to the host and handle Ctrl-C during install


giuseppe:
@yuqi-zhang could you please take a look?

yuqi-zhang:
LGTM, only minor question is why 7bd2a2 and ded0419 seems to add/remove the same part of code?

giuseppe:
I think only the indentation was changed as I moved more code under the `try/except` block

yuqi-zhang:
Oh sorry, it was a duplicate check and unlink. LGTM then.

giuseppe:

yuqi-zhang:
@rh-atomic-bot r+ ddacca4

PR #1040, opened 2017-06-30
syscontainers: by default don't install rpm on AH

giuseppe:
We still don't support it, when --system-package=auto default to no on
Atomic host.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -167,6 +167,7 @@ class SystemContainers(object):
         return self.install(image, name)
 
     def _install_rpm(self, rpm_file):
+        # Ensure _should_be_installed_rpm is changed as well.
         if os.path.exists("/run/ostree-booted"):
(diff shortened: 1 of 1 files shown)


Atomic/syscontainers.py
```diff
@@ -167,6 +167,7 @@ def install_user_container(self, image, name):
         return self.install(image, name)
 
     def _install_rpm(self, rpm_file):
```
yuqi-zhang: Is this function actually being used anywhere?

giuseppe: good point, it should be dropped

giuseppe:
@yuqi-zhang a patch to drop it (and a fixup to correct the comment)

yuqi-zhang:
LGTM

giuseppe:

yuqi-zhang:
@rh-atomic-bot r+  ba8fbe3

PR #1043, opened 2017-07-05
syscontainers: some fixes for --user

giuseppe:
one fix for mounting images as an user and support for `--pid-file` in bwrap-oci.  I am still working on the bwrap-oci part but the CLI will look as used in the PR here.  Since we check if the `--pid-file` option is available before using it, we won't break with older versions


giuseppe:
ping @yuqi-zhang

yuqi-zhang:
LGTM, is it worth it to add some tests?

giuseppe:
they would require the last version of bwrap-oci and bubblewrap, both not released yet.

I am planning to do more work on this though, in particular on reporting the user container status correctly

giuseppe:
tests are failing because of https://github.com/projectatomic/atomic/pull/1047

giuseppe:
@yuqi-zhang tests are passing again now

giuseppe:
@rhatdan this is only a first pass and probably there will be more work to be done.  OK to merge this for now?

Atomic/syscontainers.py
```diff
@@ -407,10 +408,28 @@ def _generate_default_oci_configuration(self, destination):
...
+            return self._generate_user_systemd_startstop_directives(name, pidfile, unit_file_support_pidfile)
+        else:
+            return self._generate_system_systemd_startstop_directives(name, pidfile, unit_file_support_pidfile)
 
+    # --user case
+    def _generate_user_systemd_startstop_directives(self, name, pidfile=None, unit_file_support_pidfile=False):
+        if unit_file_support_pidfile:
+            has_pidfile_option = False
+            try:
+                has_pidfile_option = "--pid-file" in str(util.check_output([util.BWRAP_OCI_PATH, "--help"], stderr=DEVNULL))
```
rhatdan: Seems a little flimsy, can't we handle this with requires of a particular version?

Atomic/syscontainers.py
```diff
@@ -1879,3 +1902,19 @@ def mount_from_storage(self, img, destination, upperdir=None, workdir=None, debu
             util.write_out(cmd)
         stderr = None if debug else DEVNULL
         return util.check_call(cmd, stderr=stderr)
+
+    @staticmethod
+    def _correct_dir_permissions_for_user(path):
+        os.chmod(path, 0o777)
```
rhatdan: Is this creating a world writable/executable directory on path?

Atomic/syscontainers.py
giuseppe: yes we could do that.  Just there is no bwrap/bwrap-oci release yet so it would break with existing versions.  I thought of changing it once we have a new version released.

Atomic/syscontainers.py
giuseppe: ah sorry.  I've added a fixup patch that changes it to `0700`

Atomic/syscontainers.py
rhatdan: Ok, add a comment to this to fix the issue when bwrap/bwrap-oci is released.

Atomic/syscontainers.py
```diff
@@ -1905,14 +1905,14 @@ def mount_from_storage(self, img, destination, upperdir=None, workdir=None, debu
...
             if os.path.islink(root):
                 continue
             for d in dirs:
                 fullpath = os.path.join(root, d)
                 if not os.path.islink(fullpath):
-                    os.chmod(fullpath, 0o777)
+                    os.chmod(fullpath, 0o700)
             for f in files:
                 fullpath = os.path.join(root, f)
                 if not os.path.islink(fullpath):
```
rhatdan: Why are you changing the permissions on directories?  But not files or other objects?

giuseppe: the problem with the directories is that we would not be able to write any file there during the checkout (we would need something like `tar --delay-directory-restore`).  Similarly we would not be able to delete the checkout since we can't delete files that are under a not writeable directory.

giuseppe:
@rhatdan fine to merge this?

rhatdan:
LGTM

baude:
@rh-atomic-bot r+ 8e94371

PR #1046, opened 2017-07-12
Remove commissaire integration

mbarnes:
The Commissaire project has halted development for the time being and never shipped in any Fedora or Red Hat products, so this PR is just to clean up the leftover cruft.

The subcommands are simple passthroughs and are only visible when the Commissaire client library is present, so few if any users should miss them.  Should the Commissaire project start moving again, we can reconsider whether it makes sense to restore this.


rhatdan:
retest

rhatdan:
@rh-atomic-bot r+ b03d69d

rhatdan:
@mbarnes any idea why this is blowing up?
@jlebon ?

rhatdan:
Ok I guess it worked.

PR #1055, opened 2017-07-23
syscontainers: fix reading image id in run_once

eyusupov:
## Description

Attempting to install a system container from an image that has a "run-once" directive fails.
```
atomic install --system openshift/origin-ansible

string indices must be integers

```
This is not reproducible in the integration tests because the code in _get_image_id_from_manifest works fine in the absence of `Digest` or `config` strings in the undecoded JSON, and the digest is only calculated when the image is pushed to the registry.


rhatdan:
bot, add author to whitelist

rhatdan:
bot, add author to whitelist

rhatdan:
@giuseppe PTAL

rhatdan:
# atomic rhost (if commctl.cli is available)
    try:
        import commctl.cli
        rhostp = subparser.add_parser(
            "rhost", help=_("execute cluster host management commands"))
        commctl.cli.add_host_commands(rhostp)
    except ImportError:
        pass

rhost support is disabled unless you install  Commissaire.

```
man atomic-rhost
ATOMIC(1)                                                                                          May 2016                                                                                         ATOMIC(1)

NAME
       atomic-rhost - Manage cluster hosts

SYNOPSIS
       atomic rhost [OPTIONS] SUBCOMMAND

       atomic rhost will issue commands to a Commissaire server over HTTP.

       Commissaire  allows administrators of a Kubernetes, Atomic Enterprise or OpenShift installation to perform administrative tasks without the need to write custom scripts or manually intervene on sys‐
       tems.
```

giuseppe:
@eyusupov good catch and thanks for the patch!

@rh-atomic-bot r+ ecc9041

eyusupov:
You're welcome :)

PR #1056, opened 2017-07-27
papr: use new name and bump to f26

jlebon:
Rename the YAML file and its auxiliary files to the newly supported name.
Bump tests to use F26, and switch to using the Alpha stream, which is
more stable than continuous.

---

We're still using a Dockerfile here to reduce friction,
though many projects ditched that approach in favour of just
provisioning the container every time. PAPR now makes use of
caching & repo injection to makes this much faster.


cgwalters:
Rebased :surfing_man:

ashcrow:
@jlebon do you still want to peruse this?

jlebon:
Oh wow, didn't realize this was still open. Nope, we can close this. It was incorporated into another PR. That said, we probably want to bump to f27 now...

PR #1062, opened 2017-08-07
Introduce building hardened images

jan-cerny:
## Description

I would like to propose an extension of Atomic scan command. It enables to create images that are compliant with a security policy (STIG, OSPP, PCI-DSS or others available in SCAP Security Guide) from existing images. 

The pull request adds a new option `--harden`  to `atomic scan`. It works like this: First, it performs a scan. Based on scan results, it creates a remediation script, and then simply builds a new image from the scanned image. During the build the fix script is be invoked. We can think of it as adding new layer on the original image. This layer contains all the changes needed to comply with a SCAP profile.

Most of the work is done by openscap-daemon. This pull request requires openscap-daemon-0.1.7 (released last week) in the underlying "openscap" container, it doesn't work with older versions of openscap-daemon.
EDIT: openscap-daemon-0.1.7 is now available in Fedora repositories

Example:
```
atomic scan --harden --scan_type configuration_compliance --scanner_args profile=xccdf_org.ssgproject.content_profile_stig-rhel7-disa registry.access.redhat.com/rhel7
``` 
I have opened this as a work in progress pull request. I assume there needs to be done much more. I have already a couple of thoughts:
* Should it use Docker API instead of a process call?
* I guess it should show the build progress
* maybe extend the JSON results
* it works for images now, what about containers? I think we need make an image from them first

Feedback and suggestions are welcome.

## Related Bugzillas
- https://bugzilla.redhat.com/show_bug.cgi?id=1472379

## Related Issue Numbers
- none

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -548,7 +548,7 @@ class Scan(Atomic):
                     raise RuntimeError("Docker exception: {}\n".format(e))
                 build_output = []
                 for item in build_output_generator:
-                    item_dict = eval(item.decode("utf-8"))
+                    item_dict = json.loads(item.decode("utf-8"))
                     if "error" in item_dict:
                         raise RuntimeError("Error during Docker build {}\n".format(item_dict["error"]))
                     if self.args.debug:
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -1,3 +1,7 @@
+import shutil
+import tempfile
+import subprocess
+
 from . import Atomic
 from . import mount
 from . import util
@@ -21,6 +25,7 @@ def cli(subparser):
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scan type"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
     scanp.add_argument("--scanner_args", default=None, help=_("Specify arguments to be passed to the scanner"))
+    scanp.add_argument("--harden", action='store_true', default=False, help=_("Create hardened image after scan"))
     disp_group = scanp.add_mutually_exclusive_group()
     disp_group.add_argument("--verbose", action='store_true', default=False, help=_("Show more output from scanning container"))
     disp_group.add_argument("--json", action='store_true', default=False, help=_("Output results in JSON format"))
@@ -176,6 +181,10 @@ class Scan(Atomic):
             # output results
             if self.useTTY:
                 self.output_results()
+
+            # build a hardened image
+            if self.args.harden:
+                self.harden()
 
             # record environment
             self.record_environment()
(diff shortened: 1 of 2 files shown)


Atomic/scan.py
```diff
@@ -517,3 +526,33 @@ def write_persistent_data(self):
...
+                    )
+            except subprocess.CalledProcessError as e:
+                util.write_err("Cannot build hardened image from {}. Docker returned {}.\n".format(target.id, e.returncode))
+                util.write_err("Details:\n{}\n".format(e.stderr.decode("utf-8")))
+                continue
+            build_output = build_process.stdout.decode("utf-8")
+            image_id = build_output.splitlines()[-1].split()[-1]
+            util.write_out("Sucessfully built hardened image {} from {}.\n".format(image_id, target.id))
+            shutil.rmtree(temp_dir)
+        return 0
```
mpreisler: this return is never used

Atomic/scan.py
```diff
@@ -517,3 +526,33 @@ def write_persistent_data(self):
...
+                        stderr=subprocess.PIPE,
+                    )
+            except subprocess.CalledProcessError as e:
+                util.write_err("Cannot build hardened image from {}. Docker returned {}.\n".format(target.id, e.returncode))
+                util.write_err("Details:\n{}\n".format(e.stderr.decode("utf-8")))
+                continue
+            build_output = build_process.stdout.decode("utf-8")
+            image_id = build_output.splitlines()[-1].split()[-1]
+            util.write_out("Sucessfully built hardened image {} from {}.\n".format(image_id, target.id))
+            shutil.rmtree(temp_dir)
```
mpreisler: This should absolutely be in `finally` and everything that can throw an exception after we create it via mkdtemp has to be in `try`. We need to make sure this is cleaned up.

Atomic/scan.py
```diff
@@ -517,3 +526,33 @@ def write_persistent_data(self):
...
+
+            shutil.copy(fix_script, temp_dir)
+            dockerfile_path = os.path.join(temp_dir, "Dockerfile")
+            with open(dockerfile_path, "w") as f:
+                f.write("FROM " + target.id + "\n")
+                f.write("COPY fix.sh /\n")
+                f.write("RUN chmod +x /fix.sh; /fix.sh\n")
+            docker_build_cmd = ['docker', 'build', '--no-cache', temp_dir]
+            try:
+                build_process = subprocess.run(
```
mpreisler: This isn't available in python2.

Atomic/scan.py
```diff
@@ -517,3 +526,33 @@ def write_persistent_data(self):
...
+            temp_dir = tempfile.mkdtemp()
+            fix_script = os.path.join(self.results_dir, target.id, "fix.sh")
+
+            shutil.copy(fix_script, temp_dir)
+            dockerfile_path = os.path.join(temp_dir, "Dockerfile")
+            with open(dockerfile_path, "w") as f:
+                f.write("FROM " + target.id + "\n")
+                f.write("COPY fix.sh /\n")
+                f.write("RUN chmod +x /fix.sh; /fix.sh\n")
+            docker_build_cmd = ['docker', 'build', '--no-cache', temp_dir]
```
mpreisler: why `--no-cache`?

baude: Would you consider using the python api for this?

http://docker-py.readthedocs.io/en/latest/images.html

jan-cerny: @baude Yes, I'll try that. Thanks.

jan-cerny: @mpreisler Yes, that was my intention to prevent this situation, although I don't know what will happen without `--no-cache`. I'll check that.

jan-cerny:
@baude I have used the Docker API instead of subprocess call.
@mpreisler I have wrapped the whole block into try..except..finally.

mpreisler:
@jan-cerny Is this still WIP? What's missing here?

jan-cerny:
@mpreisler I think I can remove the WIP label for now and we will see if there will be any other feedback.

Atomic/scan.py
```diff
@@ -1,3 +1,7 @@
+import shutil
+import tempfile
+import docker
```
baude: where is this used?

Atomic/scan.py
```diff
@@ -21,6 +25,7 @@ def cli(subparser):
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scan type"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
     scanp.add_argument("--scanner_args", default=None, help=_("Specify arguments to be passed to the scanner"))
+    scanp.add_argument("--harden", action='store_true', default=False, help=_("Create hardened image after scan"))
```
baude: I would like to talk this out next week when I return from PTO.  My preference would be that atomic scan remains neutral in this because it is all based on plugins.  For example, if i had a custom scanner, the --harden would not make any sense.  I have a couple of ideas on how to work around this and will propose them in IRC when I return.

Atomic/scan.py
jan-cerny: It's used on line 548. Docker `build` can return docker.errors.APIError exception.

Atomic/scan.py
jan-cerny: Maybe we can say that scanners can optionally support `--harden` if they wish. Scanners would declare support of automated hardening in their configuration file `/etc/atomic.d/[scanner]`. Atomic could then check if the scanner supports this option. If not, it will print an error message and exit.

PR #1070, opened 2017-08-16
<WIP> Do not merge - F26 fix failures

yuqi-zhang:
Rebased on https://github.com/projectatomic/atomic/pull/1056 and triggering tests.


yuqi-zhang:
There's one more issue with test_storage. Could you take a look @baude ? I'm not really sure what the cause is.

baude:
@yuqi-zhang i increased the size of the secondary disk in the yaml file to 20G and it got further but observed a new error.  I didn't recognize the new error but was a conflict between overlay and dm.  Maybe you have seen it?

yuqi-zhang:
@rhvgoyal do you know what is causing the storage error?

rhvgoyal:
I am not sure what to look at. What's the error message?

yuqi-zhang:
@rhvgoyal to reproduce, set up a f26-atomic host with a secondary disk. Attempt to set up that device with `atomic storage modify --add-device /dev/vdb1` and then `atomic storage modify --remove-device`. It should complain about not having enough space to move the newly created pv.

baude:
@rhvgoyal Here is the extracted error message.  This error is not observed on F25 .. only showed up on the test suite on F26.  There is some debug in there like fdisk and pvdisplay that isnt normally there.

```
==== ./tests/integration/test_storage.sh ====
+ set -euo pipefail
+ lvname=container-root-lv
+ rootfs=/var/lib/containers
+ setup
+ copy /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage-setup.atomic-tests-backup
+ '[' -f /etc/sysconfig/docker-storage-setup ']'
+ cp /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage-setup.atomic-tests-backup
+ copy /etc/sysconfig/docker-storage /etc/sysconfig/docker-storage.atomic-tests-backup
+ '[' -f /etc/sysconfig/docker-storage ']'
+ cp /etc/sysconfig/docker-storage /etc/sysconfig/docker-storage.atomic-tests-backup
+ TEST_DEV_1=/dev/vdb
+ xargs -r umount
+ uniq
+ grep '^/dev/vdb'
+ findmnt -o SOURCE
+ wipefs -a /dev/vdb
/dev/vdb: 8 bytes were erased at offset 0x00000052 (vfat): 46 41 54 33 32 20 20 20
/dev/vdb: 1 byte was erased at offset 0x00000000 (vfat): eb
/dev/vdb: 2 bytes were erased at offset 0x000001fe (vfat): 55 aa
+ TEST_DEV_1_pvs=/dev/vdb1
++ awk '$2 ~ /^\/$/ && $1 !~ /rootfs/ { print $1 }' /proc/mounts
+ ROOT_DEV=/dev/mapper/atomicos-root
++ lvs --noheadings -o vg_name /dev/mapper/atomicos-root
+ VGROUP='  atomicos'
+ trap teardown EXIT
+ rm -f /etc/sysconfig/docker-storage-setup
+ rm -f /etc/sysconfig/docker-storage
+ '[' -n '  atomicos' ']'
+ cat
++ atomic --debug storage modify --add-device /dev/vdb
+ OUTPUT=
+ grep -q '^DEVS="/dev/vdb"$' /etc/sysconfig/docker-storage-setup
++ pvs --noheadings -o vg_name /dev/vdb1
+ '[' atomicos == atomicos ']'
+ pvdisplay
  --- Physical volume ---
  PV Name               /dev/vda2
  VG Name               atomicos
  PV Size               19.71 GiB / not usable 2.00 MiB
  Allocatable           yes 
  PE Size               4.00 MiB
  Total PE              5044
  Free PE               1281
  Allocated PE          3763
  PV UUID               rUSiOJ-kj9d-OA8p-3o4y-1aDY-gTVP-oXA00b
   
  --- Physical volume ---
  PV Name               /dev/vdb1
  VG Name               atomicos
  PV Size               10.00 GiB / not usable 3.00 MiB
  Allocatable           yes 
  PE Size               4.00 MiB
  Total PE              2559
  Free PE               1033
  Allocated PE          1526
  PV UUID               959VBS-kIcc-o83r-3Zjn-G3jI-oSaD-aZFzAy
   
+ fdisk -l
Disk /dev/vda: 20 GiB, 21474836480 bytes, 41943040 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x8960c03d

Device     Boot  Start      End  Sectors  Size Id Type
/dev/vda1  *      2048   616447   614400  300M 83 Linux
/dev/vda2       616448 41943039 41326592 19.7G 8e Linux LVM


Disk /dev/vdb: 10 GiB, 10737418240 bytes, 20971520 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x2d15d890

Device     Boot Start      End  Sectors Size Id Type
/dev/vdb1        2048 20971519 20969472  10G 8e Linux LVM


Disk /dev/mapper/atomicos-root: 8 GiB, 8514437120 bytes, 16629760 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes


Disk /dev/mapper/atomicos-docker--root--lv: 6.7 GiB, 7201619968 bytes, 14065664 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
+ atomic --debug storage modify --remove-device /dev/vdb
Return Code: 5
atomic storage modify --remove-device failed
  File "/usr/lib/python3.6/site-packages/Atomic/storage.py", line 238, in _remove_devices
    util.check_output([ "pvmove", pv ], stderr=subprocess.STDOUT)
  File "/usr/lib/python3.6/site-packages/Atomic/util.py", line 213, in check_output
    return subprocess.check_output(cmd, env=env, stdin=stdin, stderr=stderr, close_fds=True)
  File "/usr/lib64/python3.6/subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "/usr/lib64/python3.6/subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['pvmove', '/dev/vdb1']' returned non-zero exit status 5.

During handling of the above exception, another exception occurred:

  File "/usr/bin/atomic", line 185, in <module>
    sys.exit(_func())
  File "/usr/lib/python3.6/site-packages/Atomic/storage.py", line 176, in modify
    self._remove_devices(self.args.remove_devices, only_unused=False)
  File "/usr/lib/python3.6/site-packages/Atomic/storage.py", line 255, in _remove_devices
    raise ValueError("atomic storage modify --remove-device failed")
ValueError: atomic storage modify --remove-device failed

During handling of the above exception, another exception occurred:

  File "/usr/bin/atomic", line 198, in <module>
    sys.exit(1)
SystemExit: 1
Failure: b'  Insufficient free space: 1526 extents needed, but only 1281 available\n  Unable to allocate mirror extents for atomicos/pvmove0.\n  Failed to convert pvmove LV to mirrored\n'
+ teardown
+ local mnt
+ set -e
+ wipefs -a /dev/vdb
```

rhvgoyal:
Ok, looking deeper. So /dev/vdb1 has already 1526 extents allocated. And when you are trying to remove device from volume group, it is trying to move those to /dev/vda2 but that pv has only 1281 extents free and that's why pvremove failed.

1526 extents are around 6GB. I suspect that this is consumed by volume `/dev/mapper/atomicos-docker--root--lv`. This is the volume setup by docker-storage-setup for overlay for docker.

So either we need to remove this volume first or have a bigger /dev/vda or create smaller docker-root-lv to begin with for this test to pass.

There are some things which are unexplained though. Why this volume size is around 6.7 GB. We create 40% of free space by default. Is your environment different where you override `CONTAINER_ROOT_LV_SIZE=40%FREE`

Secondly, why there are only 1281 extents free on /dev/vda2. atomicos-root volume is of around 8GB size. And total disk size is around 19GB. So 11 GB should have been available and migration of 6.7 GB should have been successful. So I don't know why /dev/vda2 pv's space usage is more.

rhvgoyal:
Is it possible to get "lsblk" output also after disk has been added  (and before removal)

rhvgoyal:
Ok, I think I know what is happening. 

- So we add a device of 10GB to volume group atomicos. That bumps up size of VG to around 30GB
- Then docker-storage-setup is started (I think atomic must be calling it). And it tries to allocate 40% of free space for docker-root-lv.
- Currently around 8GB is already occupied by atomicos-root lv. That means around 22GB is free and it should have created a volume of size around 8GB. Right now I see 6.7GB. not sure why. That means it will also allocate PE both from /dev/vda2 and /dev/vdb1.

- After this operation, now not enough free PEs are left in /dev/vda2 that we can remove /dev/vdb.

if you try `CONTAINER_ROOT_LV_SIZE=20%FREE`, your test might still pass. Otherwise you will have to modify your tests.

rhvgoyal:
IOW, I don't see this as  a software bug. It is more about how configuration is being setup and tested.

rhvgoyal:
You can't remove a disk which is already being used and other disk in volume group does not have enough free space so that one can move all the PEs to other disk.

yuqi-zhang:
@rhvgoyal Thank you very much for the insight. Neither I nor @baude were very familiar with the storage stuff. So if I understand it correct there are two ways of handling this: 1. play around with the sizes so the pvmove is successful or 2. reset docker storage and do the tests. In case of 2, we need to ensure the test runs last since it will otherwise cause all other tests to fail that interact with the test images.

What do you think @baude

yuqi-zhang:
@rhvgoyal just to make sure I'm understanding this correctly, what is the key change from devicemapper->overlay that causes this behaviour? When we added the extra disk, shouldn't c-s-s have tried to do the same thing with devicemapper that it is doing with overlay, since overlay is still using a separate LV for docker? If it is the same idea, then why doesn't the failure happen on f25-atomic? (or any of the other flavours)

yuqi-zhang:
Made a cleaner version at https://github.com/projectatomic/atomic/pull/1076, will close in favour of that once tests passes.

rhvgoyal:
On fedora atomic, by default we have switched to overlay2 driver and also switched to setting up a separate logical volume which will be used by docker to save images/containers using overlay2 fs on top. That's why you see `docker-root-lv`. Now it is possible to have `docker-root-lv` while using devicemapper graph driver but I am not sure how many people will really use that configuration.

PR #1079, opened 2017-08-23
dockertar-sha256-helper: vendor gzip and flate from golang1.6

giuseppe:
The implementation changed in golang1.8 and the returned checksum is
different.  Vendor the version from golang 1.6 so that we can produce
the same checksum.


ashcrow:
@giuseppe I assume the reason to use 1.6 over the latest release is for backwards compatibility?

giuseppe:
if we change the version (and the checksum) then layers that were pulled in the past need to be pulled again.  The gzip/deflate implementation in golang 1.6 gives the same checksum as Docker (did?)

ashcrow:
@giuseppe I see. Makes sense.

giuseppe:
the tests keep failing for a timeout

ashcrow:
bot, test this please

ashcrow:
I've asked @jlebon if there is any known issues with f26 tester.

ashcrow:
bot, retest this please

.papr.yml
```diff
@@ -51,3 +51,5 @@ packages:
   - atomic
   - python3-coverage
   - rpm-build
+  - findutils
+  - make
```
yuqi-zhang: I didn't know that github complained about missing newline

ashcrow: I think it's just noting that there isn't a newline after.

ashcrow:
FWIW it looks like moby doesn't vendor anything in the ``compress`` package.

ashcrow:
bot, retest this please

yuqi-zhang:
Hm, seems like same failure as last time, where its stuck trying to pull the centos image from docker hub. Trying on a f26 atomic host myself I cannot reproduce this, strange...

ashcrow:
@yuqi-zhang there is a known issue right now. The best we can do is keep on top of it and try until it passes.

ashcrow:
@giuseppe I have two commits in https://github.com/ashcrow/atomic/tree/vendor-dir which move vendored code into ``./vendor`` as a start. Feel free to include it otherwise I'll make a different PR after this merges.

giuseppe:
@vbatts does this change make any sense to you?  Do you know if Docker is going to use the gzip implementation from golang 1.8 and we can switch as well?

vbatts:
oh interesting. I figured this day would come. It would be ripping a bandaid off to move to the newer gzip/flate. Is the go1.8 implementation consistent to itself, just not consistent to prior versions?

giuseppe:
@vbatts yes, go1.8 is consistent to itself but it differs from golang1.6.

vbatts:
it makes sense, but eventually will have conflicts with checksums by other systems if they haven't done the same. This is why many folks (e.g. me) have pushed to only compare checksums of the _uncompressed_ archive [where possible]. I think this is something docker have begun doing more of.

From a quick look at moby/moby master, they are not vendoring/forking this code and only importing from `compress/gzip`, so where they're relying on the checksum of a compressed archive, the hash will change when the daemon is compiled with newer golang.

giuseppe:
Thanks.  If dockertar-sha256-helper was compiled with golang 1.8 then it is probably safe to assume the same version of golang was used for Moby or any other tool that calculates the checksum.  I am going to close this PR since there are no clear benefits and we start carrying a lot more code

PR #1082, opened 2017-08-25
syscontainers: use Skopeo copy to pull images

giuseppe:
New versions of Skopeo support "ostree:" as a destination for copy.
Use directly Skopeo to copy an image to the OSTree repository.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1822,7 +1822,7 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         if not upgrade and current_rev[1]:
             return False
 
-        can_use_skopeo_copy = str(util.check_output([util.SKOPEO_PATH, "copy", "--help"])).find("ostree") >= 0
+        can_use_skopeo_copy = util.check_output([util.SKOPEO_PATH, "copy", "--help"]).decode().find("ostree") >= 0
 
         if can_use_skopeo_copy:
             return self._check_system_oci_image_skopeo_copy(repo, img)
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1821,6 +1821,28 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         current_rev = repo.resolve_rev(imagebranch, True)
         if not upgrade and current_rev[1]:
             return False
+
+        can_use_skopeo_copy = str(util.check_output([util.SKOPEO_PATH, "copy", "--help"])).find("ostree") >= 0
+
+        if can_use_skopeo_copy:
+            return self._check_system_oci_image_skopeo_copy(repo, img)
+        return self._check_system_oci_image_no_skopeo_copy(repo, img, imagebranch)
+
+    def _check_system_oci_image_skopeo_copy(self, repo, img):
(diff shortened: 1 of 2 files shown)


giuseppe:
tests are failing since Skopeo on CentOS doesn't support ostree as a destination yet.

Can we get an updated version there or should we delay this PR?

rhatdan:
7.4 is released, we need to start moving to it, although for atomic use cases, we will probably need to support 7.3 for a while.

giuseppe:
I've changed the implementation to check at runtime if Skopeo supports "copy" to an OSTree repository.  We can drop this check once we are sure it is supported everywhere.

baude:
@giuseppe you ready for this to merge ?

giuseppe:
yes I think it is ready to go, @ashcrow what do you think?

ashcrow:
@giuseppe reviewing now

Atomic/syscontainers.py
```diff
@@ -1821,6 +1821,28 @@ def _check_system_oci_image(self, repo, img, upgrade):
         current_rev = repo.resolve_rev(imagebranch, True)
         if not upgrade and current_rev[1]:
             return False
+
+        can_use_skopeo_copy = str(util.check_output([util.SKOPEO_PATH, "copy", "--help"])).find("ostree") >= 0
```
ashcrow: ``util.check_output()`` forwards to ``subprocess.check_output()``. On Python 2 the result will be a string but on Python 3 it will be bytes. While one can still technically find ``"ostree"`` in the results, it will be looking at a string representation of bytes.

To check the same result on both Python 2 and 3 you can use ``decode()`` like:

```python
can_use_skopeo_copy = util.check_output([util.SKOPEO_PATH, "copy", "--help"]).decode().find("ostree") >= 0
```

In both cases it will be checking a ``unicode`` instance.

giuseppe: thanks, I've pushed a fixup patch to add the missing `.decode()`

Atomic/syscontainers.py
```diff
@@ -1822,7 +1822,7 @@ def _check_system_oci_image(self, repo, img, upgrade):
         if not upgrade and current_rev[1]:
             return False
 
-        can_use_skopeo_copy = str(util.check_output([util.SKOPEO_PATH, "copy", "--help"])).find("ostree") >= 0
+        can_use_skopeo_copy = str(util.check_output([util.SKOPEO_PATH, "copy", "--help"])).decode().find("ostree") >= 0
```
ashcrow: Almost ... the ``str()`` is still going to cause a problem. Remove it and you should be good:

```
$ python3 -c "str(b'test').decode()"
  File "<string>", line 1, in <module>
AttributeError: 'str' object has no attribute 'decode'
```

```
$ python3 -c "print(b'test'.decode())"
test
$ python2 -c "print('test'.decode())"
test
```

giuseppe: oops, dropped it and pushed a new verison

giuseppe:
ready to 🚢!

ashcrow:
@baude ready for merge.

baude:
@rh-atomic-bot r+ 70f9e86

PR #1085, opened 2017-08-30
system containers: prevent files to be left installed if the checkout could not be deleted

giuseppe:
It solves some issues I've seen on repeatedly installing/uninstalling the OpenShift system containers.  In some cases the containers are restarted immediately during the `uninstall` phase and it caused some files (such as the systemd service file) to be left on the file system.  The half installed container prevented it to be reinstalled without manual intervention.

Also, now "atomic images prune" is a bit more aggressive and it frees immediately not referenced objects from the the ostree storage (the extra `ostree prune --refs-only` seemed to be confusing).

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1441,11 +1441,13 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         if not self.display:
             util.check_call(cmd)
 
-    def _systemctl_command(self, command, name=None, quiet=False):
+    def _systemctl_command(self, command, name=None, now=False, quiet=False):
         cmd = ["systemctl"]
         if self.user:
             cmd.append("--user")
         cmd.append(command)
+        if now:
+            cmd.append("--now")
         if name:
             cmd.append(name)
         if not quiet:
@@ -1505,11 +1507,7 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         unitfileout, tmpfilesout = self._get_systemd_destination_files(name)
         if has_container_service:
             try:
-                self._systemctl_command("stop", name)
-            except subprocess.CalledProcessError:
-                pass
-            try:
-                self._systemctl_command("disable", name)
+                self._systemctl_command("disable", name, now=True)
             except subprocess.CalledProcessError:
                 pass
 
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1516,6 +1516,11 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         if os.path.exists(unitfileout):
             os.unlink(unitfileout)
 
+        try:
+            self._systemctl_command("daemon-reload")
+        except subprocess.CalledProcessError:
+            pass
+
         if os.path.exists(tmpfilesout):
             try:
                 self._systemd_tmpfiles("--remove", tmpfilesout)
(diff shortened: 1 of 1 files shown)


Atomic/syscontainers.py
```diff
@@ -521,7 +521,7 @@ def _resolve_remote_path(self, remote_path):
         return real_path
 
     def _checkout(self, repo, name, img, deployment, upgrade, values=None, destination=None, extract_only=False, remote=None, prefix=None, installed_files=None, system_package='no'):
-        destination = destination or "%s/%s.%d" % (self._get_system_checkout_path(), name, deployment)
+        destination = destination or os.path.join(self._get_system_checkout_path(), "{}.{}".format(name, deployment))
```
ashcrow: 👍

Atomic/syscontainers.py
```diff
@@ -1528,18 +1531,22 @@ def uninstall(self, name):
...
-        if os.path.exists(unitfileout):
-            os.unlink(unitfileout)
-
         if rpm_installed:
             RPMHostInstall.uninstall_rpm(rpm_installed.replace(".rpm", ""))
 
+        # Until the symlink and the deployment are in place, we can attempt the uninstall again.
+        # So treat as fatal each failure that happens before we rm the symlink.
+        symlink_path = os.path.join(checkout, name)
+        if os.path.lexists(symlink_path):
```
ashcrow: Is a ``daemon-reload`` needed as well?

Atomic/syscontainers.py
```diff
@@ -1550,6 +1557,32 @@ def prune_ostree_images(self):
...
+
+            if not os.path.isdir(fullpath):
+                continue
+
+            # If we got here, the deployment is not related to an installed
+            # container, so delete it.
+            try:
+                shutil.rmtree(fullpath)
+            except OSError:
+                pass
```
ashcrow: A log line here may be good noting the error.

Atomic/syscontainers.py
giuseppe: that should probably happen just in case of `systemctl start container-just-uninstalled` but sure, it looks like a good idea, going to add a patch

Atomic/syscontainers.py
giuseppe: added a fixup patch

yuqi-zhang:
bot, retest this please

Atomic/syscontainers.py
```diff
@@ -1550,6 +1562,32 @@ def prune_ostree_images(self):
...
+                continue
+
+            container = i[:-2]
+
+            # Check if the container is still installed.
+            if os.path.lexists(os.path.join(checkouts, container)):
+                continue
+
+            fullpath = os.path.join(checkouts, i)
+
```
yuqi-zhang: Small nit, too much extra space here and there

ashcrow: Personally I like the space before and after the line as it breaks up the two checking blocks. However, I think it could be pulled down to be just above the second check block as it's related.

giuseppe: thanks, I've amended the last fixup patch and have now:

```
            fullpath = os.path.join(checkouts, i)
            if not os.path.isdir(fullpath):
                continue
```

Indeed it looks better!  I usually don't leave a space between a declaration and `if ...` using the variable itself

giuseppe:

giuseppe:
@rhatdan can we add @ashcrow to the list of atomic developers?

ashcrow:
@rh-atomic-bot r+ 64a017e7eccabf9d94e6473edefd88a3b6eff72c

giuseppe:
@rh-atomic-bot retry

PR #1087, opened 2017-08-30
storage: On reset, umount docker root dir before removing

mbarnes:
If the storage configuration used
```
  CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/docker
```
then the volume must be unmounted before attempting to remove the directory.

Fixes https://github.com/projectatomic/container-storage-setup/issues/254

### DIFF
--- a/Atomic/storage.py
+++ b/Atomic/storage.py
@@ -160,6 +160,7 @@ class Storage(Atomic):
         util.call(["umount", root + "/devicemapper"], stderr=DEVNULL)
         util.call(["umount", root + "/overlay"], stderr=DEVNULL)
         util.call(["umount", root + "/overlay2"], stderr=DEVNULL)
+        util.call(["umount", root], stderr=DEVNULL)
         shutil.rmtree(root)
         os.mkdir(root)
         try:


rhatdan:
LGTM
@rhvgoyal @baude PTAL

baude:
LGTM

rhvgoyal:
container-storage-setup takes care of unmounting this. This mount was done by container-storage-setup, so I think it should be container-storage-setup cleaning up this mount. 

This patch will work too, but IMO, this is a little hack fix. "atomic storage reset" should not have to deal with parsing for `CONTAINER_ROOT_LV_MOUNT_PATH`.

rhatdan:
@mbarnes DId you see the mount point not being cleaned up?

rhvgoyal:
Is the ordering right? I mean first `container-storage-setup --reset` needs to be called and then removal of /var/lib/docker/ needs to happen?

rhatdan:
Here is atomic storage reset code:

```
    def reset(self):
        root=self.graphdir
        try:
            self.d.info()
            raise ValueError("Docker daemon must be stopped before resetting storage")
        except (NoDockerDaemon, requests.exceptions.ConnectionError):
            pass
        util.check_call(["docker-storage-setup", "--reset"], stdout=DEVNULL)
        util.call(["umount", root + "/devicemapper"], stderr=DEVNULL)
        util.call(["umount", root + "/overlay"], stderr=DEVNULL)
        util.call(["umount", root + "/overlay2"], stderr=DEVNULL)
        shutil.rmtree(root)
        os.mkdir(root)
        try:
            selinux.restorecon(root.encode("utf-8"))
        except (TypeError, OSError):
            selinux.restorecon(root)
```

rhvgoyal:
I just tested upstream container-storage-setup and it is cleaning up the mount point. In fact I tried `atomic storage reset` and that seems to be working fine too on F26. So I am not sure where is the problem?

Following is my config.

VG=docker-vg
CONTAINER_ROOT_LV_NAME=container-root-lv
CONTAINER_ROOT_LV_MOUNT_PATH=/var/lib/docker
STORAGE_DRIVER=overlay2

mbarnes:
> @mbarnes DId you see the mount point not being cleaned up?

At the time, yes, I observed the mount point not being cleaned up.  But now that I'm looking at it again, `atomic storage reset` does seem to be working here starting from this config:
```
# Edit this file to override any configuration options specified in
# /usr/share/container-storage-setup/container-storage-setup.
#
# For more details refer to "man container-storage-setup"
DEVS=/dev/vdb
VG=docker_vg
DATA_SIZE=99%VG

STORAGE_DRIVER="overlay2"

CONTAINER_ROOT_LV_MOUNT_PATH="/var/lib/docker"

CONTAINER_ROOT_LV_NAME="docker-root-lv"
```
I see now where @rhvgoyal is referring to in `container-storage-setup.sh`: if I'm reading this right it looks like `reset_extra_volume_compat` has code to unmount the container root volume on `docker-storage-setup --reset`.  Perhaps a pre-condition was failing such that `umount` wasn't getting called.  Unfortunately I didn't capture my config when I was seeing the failure, or I'd trace it.

If I stumble on this again, I'll investigate the appropriate package.  Seems to me now that `atomic` can be left alone.

rhatdan:
Closing for now, reopen if you get a reproducer.

PR #1090, opened 2017-09-05
Atomic/scan.py: Add ability to remediate

baude:
## Description

In the case of some scanners, they generate a script to remediate the
images|containers they have scanned.  We needed to provide a hook
for this ability. We now read the scanner's configuration file looking
for a "remediation_script" key, whose value should be a fq path to the
remediation script.  The remediation script should be delivered via
the scanning image via atomic install.

As of now, we pass the id of the scanned object and its results directory
as named arguments to the remediation script.

Signed-off-by: baude <bbaude@redhat.com>

## Related Bugzillas
-
-

## Related Issue Numbers
-
-

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -20,6 +20,7 @@ def cli(subparser):
     scanp.add_argument("--scanner", choices=[x['scanner_name'] for x in scanners], default=None, help=_("define the intended scanner"))
     scanp.add_argument("--scan_type", default=None, help=_("define the intended scan type"))
     scanp.add_argument("--list", action='store_true', default=False, help=_("List available scanners"))
+    scanp.add_argument("--remediate", action='store_true', default=False, help=_("Perform remediation of scanned object"))
     scanp.add_argument("--scanner_args", default=None, help=_("Specify arguments to be passed to the scanner"))
     disp_group = scanp.add_mutually_exclusive_group()
     disp_group.add_argument("--verbose", action='store_true', default=False, help=_("Show more output from scanning container"))
@@ -123,6 +124,15 @@ class Scan(Atomic):
 
         self.results_dir = os.path.join(self.results, self.scanner, self.cur_time)
 
+        # Get remediation script
+        if self.args.remediate:
+            for i in self.scanners:
+                remediation_script = i.get("remediation_script", None)
+                break
+            if remediation_script is None:
+                raise ValueError("{} has no remediation script.".format(self.scanner))
+
+
         # Create the input directory
         if not os.path.exists(self.chroot_dir):
             os.makedirs(self.chroot_dir)
@@ -193,6 +203,12 @@ class Scan(Atomic):
             json_results = json.load(open(json_file))
             if json_results['Successful'].upper() != 'TRUE':
                 return 1
+
+        # Perform remediation
+        if self.args.remediate:
+            for i in self.scan_list:
+                self.remediate(remediation_script, i.id, self.results_dir)
+
         return 0
 
     def _get_scan_list(self):
@@ -517,3 +533,6 @@ class Scan(Atomic):
 
         with open(summary_file, 'w') as f:
             json.dump(persistent_data, f, indent=4)
+
+    def remediate(self, script, iid, results_dir):
+        util.check_call([sys.executable, script, '--id', iid, '--results_dir', results_dir])
--- a/bash/atomic
+++ b/bash/atomic
@@ -250,6 +250,7 @@ _atomic_scan() {
 		--containers
 		--rootfs
 		--list
+		--remediate
 		--scanner
 		--scan_type
 		--verbose
--- a/docs/atomic-scan.1.md
+++ b/docs/atomic-scan.1.md
@@ -54,6 +54,12 @@ Select a scan_type other than the default.
   Rootfs path to scan.  Can provide _--rootfs_ multiple times.
   Note: SELinux separation will be disabled for --rootfs scans, but all other container
   separation will still be in place.
+  
+**--remediate**
+  Allows the scanner to run a remediation script when scanning is complete.  The remediation script is provided
+  by the scanner itself.  
+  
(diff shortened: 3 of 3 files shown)


baude:
[Add ability to remediate from atomic scan](https://trello.com/c/nrSZigSY/432-add-ability-to-remediate-from-atomic-scan)

Atomic/scan.py
```diff
@@ -123,6 +125,14 @@ def get_additional_args():
 
         self.results_dir = os.path.join(self.results, self.scanner, self.cur_time)
 
+        # Get remediation script
+        if self.args.remediate:
+            for i in self.scanners:
+                self.remediation_script = i.get("remediation_script", None)
```
mpreisler: shouldn't this break once it finds it? Or shouldn't it append to list of remediation scrpits? This code seems suspect.

baude: agree, nice comment

baude:
bot, retest this please

Atomic/scan.py
```diff
@@ -517,3 +533,7 @@ def write_persistent_data(self):
 
         with open(summary_file, 'w') as f:
             json.dump(persistent_data, f, indent=4)
+
+
+    def remediate(self, script, iid, results_dir):
+        util.check_call([sys.executable, script, 'id={}'.format(iid), 'results_dir={}'.format(results_dir)])
```
jan-cerny: @baude Could it be `--id` and `--results_dir`?

Atomic/scan.py
```diff
@@ -193,6 +204,11 @@ def get_additional_args():
             json_results = json.load(open(json_file))
             if json_results['Successful'].upper() != 'TRUE':
                 return 1
+
+        # Perform remediation
+        for i in self.scan_list:
+            self.remediate(self.remediation_script, i.id, self.results_dir)
```
jan-cerny: This should be run only if `--remediate` was specified.

jan-cerny:
The second part of this effort is proposed to OpenSCAP Daemon upstream in this pull request: https://github.com/OpenSCAP/openscap-daemon/pull/110

Atomic/scan.py
baude: Fixed. Nice catch

baude:
@rhatdan ready for merge

jan-cerny:
@baude @rhatdan I'm OK with this.

baude:
@rhatdan ptal

Atomic/scan.py
```diff
@@ -123,6 +125,15 @@ def get_additional_args():
 
         self.results_dir = os.path.join(self.results, self.scanner, self.cur_time)
 
+        # Get remediation script
+        if self.args.remediate:
+            for i in self.scanners:
+                self.remediation_script = i.get("remediation_script", None)
```
rhatdan: Any reason for this to be in self?  Is the remediate_script only used in this function?

rhatdan:
Bash completions?
Man pages?

Atomic/scan.py
mpreisler: +100
@baude could this please follow standard UNIX command line conventions? Would enable us to use argparse and get a lot of safeties for free in the script.

Atomic/scan.py
baude: Fixed.

Atomic/scan.py
baude: ok

baude:
updated with request change to command line, bash complementations, and docs

Atomic/scan.py
```diff
@@ -517,3 +533,7 @@ def write_persistent_data(self):
 
         with open(summary_file, 'w') as f:
             json.dump(persistent_data, f, indent=4)
+
+
+    def remediate(self, script, iid, results_dir):
+        util.check_call([sys.executable, script, '--id {}'.format(iid), '--results_dir {}'.format(results_dir)])
```
rhatdan: Shouldn't these be all separate.

   util.check_call([sys.executable, script, '--id', '{}'.format(iid), '--results_dir', '{}'.format(results_dir)])

baude: I think I will separate them but not use the format statements, they are no useful here.

baude:
@rh-atomic-bot r+ fc5ae37

giuseppe:
does it modify the image?

baude:
@giuseppe that depends on the implementation of the remediator.  in the case of openscap, yes they do.

mpreisler:
> @giuseppe that depends on the implementation of the remediator. in the case of openscap, yes they do.

This may get a little technical but we actually don't modify the image. We take the top layer of the image and build on top of it. What we end up with is a different image. You can think of it as the remediator building a "hardening layer" on top of the existing image.

giuseppe:
we will need to disable it for system containers as it can corrupt the ostree storage.  I can do that later, as a follow up patch (and before that verify the image is always mounted ro from ostree, so the issue won't really happen).

baude:
@giuseppe sounds good, can you also make sure it gets disabled for containers-storage/runc as well?  Maybe through the use of a backend conditional check or something?

giuseppe:
sure, sounds good to me

baude:
bot, retest this please

baude:
bot, retest this please

giuseppe:
@rh-atomic-bot retry

giuseppe:
bot, retest this please

baude:
bot, retest this please

baude:
@jlebon can you peek at what might be going on here?

giuseppe:
bot, retest this please

jlebon:
There's issues with the backing OpenStack instance. It's being worked on.

TomSweeneyRedHat:
LGTM with happy tests.

baude:
bot, retest this please

baude:
@rh-atomic-bot r+ 1c99dd1

PR #1098, opened 2017-09-11
syscontainers: fix systemd directive tag for Restart=

giuseppe:
Probably nobody noticed as it is used only for containers
that do not have a default config.json.template file.  I've seen
the error while installing busybox for a test.


giuseppe:

A tiny change :-)

ashcrow:
@rh-atomic-bot r+ 88ba8a1

Issue #1101, opened 2017-09-20
storage: Processing --lvsize option still tries to use libdss.sh


mbarnes:
Atomic code is still using `docker-storage-setup`, but there's no `libdss.sh -> libcss.sh` soft link like there is for `docker-storage-setup -> container-storage-setup`.

```
# atomic storage modify --driver=overlay2 --lvname=docker-root-lv --lvsize=100%FREE --rootfs=/var/lib/docker
bash: /usr/lib/docker-storage-setup/libdss.sh: No such file or directory
bash: check_data_size_syntax: command not found
Invalid format for --lvsize
```

mbarnes:
Whoops, nevermind.  This is fixed in master already, was looking at old code.

PR #1102, opened 2017-09-20
Use container-storage-setup / libcss.sh

mbarnes:
Closes: #1101 

The 2nd commit is just cleanup, not required.


mbarnes:
Nevermind, this was addressed already in https://github.com/projectatomic/atomic/pull/1076

PR #1105, opened 2017-09-26
syscontainers: move storage to /var/lib/containers/atomic when not running on AH

giuseppe:
When /var/lib/containers/atomic is not on the same file system as /ostree we weren't able to create hardlinks thus falling back to copying files.  Additionally, when not running on Atomic Host, this avoids to create a system wide /ostree that is used only for system containers.

This problem doesn't exist on Atomic Host since we use the OSTree sysroot to get the destination path for the checkouts.  On AH, we will still continue to use /ostree/repo so that we are able to share the storage with the system.

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1495588

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -964,7 +964,17 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         if self.user:
             return os.path.join(HOME, ".containers/repo")
 
-        return self.get_atomic_config_item(["ostree_repository"]) or "/ostree/repo"
+        if self.get_atomic_config_item(["ostree_repository"]) is not None:
+            return self.get_atomic_config_item(["ostree_repository"])
+
+        # If we are not running on Atomic and not using a system wide ostree
+        # storage create the repository under /var/lib/containers/atomic/.storage/.ostree
+        # and avoid problems with creating hard links when /ostree and the checkout
+        # path are on different file systems.
+        if not os.path.exists("/ostree/repo"):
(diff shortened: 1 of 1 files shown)


rhatdan:
@giuseppe This means that storage is only used under /var/lib/containers/atomic iff you are on a non atomic host machine?  Nothing is stored in that directory otherwise?

rhatdan:
Should there be changes in the man pages for this change?

giuseppe:
@rhatdan no this is to change the location of the OSTree storage itself.  On Atomic Host it is shared with the system under /ostree/repo.  On non-AH since there is no system storage, we can decide where to create it.  For simplicity before it was created in the same location as AH, but this gives us problems when we try to create hard links since /ostree/repo and /var/lib/containers/atomic (where containers are checked out) can be on different file systems.  If we create the OSTree repository under /var/lib/containers/atomic/.storage/ostree itself, then we won't have this problem.
This issue doesn't happen on AH at all since we create the hard links using the OS deployment (so any bind mount on /var/ won't affect it).

There should not be changes in the man pages I think as there are no visible changes to the user.  Also, this won't break systems that are already using /ostree/repo as we check if the directory exists.

giuseppe:
alternative solution: https://github.com/projectatomic/atomic-system-containers/pull/120

ashcrow:
Merged https://github.com/projectatomic/atomic-system-containers/pull/120

giuseppe:
let's close this for now...

PR #1106, opened 2017-10-03
tests/storage: only umount /dev/vdb if needed

jlebon:
Split the gory `findmnt` pipeline into a conditional so that the `grep`
doesn't fail if there's nothing to unmount. Also make sure to let things
settle if we have to unmount.

### DIFF
--- a/tests/integration/test_storage.sh
+++ b/tests/integration/test_storage.sh
@@ -35,7 +35,9 @@ setup () {
     # Perform setup routines here.
     copy /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage-setup.atomic-tests-backup
     TEST_DEV_1=/dev/vdb
-    findmnt -o SOURCE | grep "^$TEST_DEV_1" | uniq | xargs -r umount
+    if findmnt -o SOURCE | grep "^$TEST_DEV_1"; then
+        findmnt -o SOURCE | grep "^$TEST_DEV_1" | uniq | xargs umount
+    fi
     wipefs -a "$TEST_DEV_1"
     TEST_DEV_1_pvs=${TEST_DEV_1}1
 


jlebon:
This is needed now because PAPR nodes are provisioned with attached Cinder volumes, which are no longer automatically mounted since they are true block devices (unlike ephemeral disks, which were pre-formatted with ext4 and mounted at `/var/mnt`).

tests/integration/test_storage.sh
```diff
@@ -35,8 +35,11 @@ setup () {
     # Perform setup routines here.
     copy /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage-setup.atomic-tests-backup
     TEST_DEV_1=/dev/vdb
-    findmnt -o SOURCE | grep "^$TEST_DEV_1" | uniq | xargs -r umount
-    wipefs -a "$TEST_DEV_1"
+    if findmnt ${TEST_DEV_1}; then
+        umount ${TEST_DEV_1};
+        udevadm settle
```
TomSweeneyRedHat: is this call to udevadm needed?  I'm not seeing what it does for you.
Otherwise LGTM

jlebon: Actually, we don't need this whole chunk at all. I had left it to be compatible with whatever setup folks may have had locally, though it's rather unusual to have the block device itself be mounted. I nuked it completely below. ⬇️

tests/integration/test_storage.sh
```diff
@@ -35,8 +35,7 @@ setup () {
     # Perform setup routines here.
     copy /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage-setup.atomic-tests-backup
     TEST_DEV_1=/dev/vdb
-    findmnt -o SOURCE | grep "^$TEST_DEV_1" | uniq | xargs -r umount
-    wipefs -a "$TEST_DEV_1"
+    wipefs -a ${TEST_DEV_1}
```
TomSweeneyRedHat: do you need -f here in case it's mounted?

rhatdan:
@rhvgoyal PTAL

giuseppe:
anything against merging this PR?  Tests on other PRs are still failing

tests/integration/test_storage.sh
rhatdan: @giuseppe Can you address this?

rhatdan: -f, --force
              Force erasure, even if  the  filesystem  is  mounted.   This  is
              required  in  order  to  erase  a partition-table signature on a
              block device.

giuseppe: Opened a PR here: https://github.com/projectatomic/atomic/pull/1111

jlebon: Sorry, was offline today. I updated the PR below ⬇️ to just conditionalize on the `grep`. See the commit message for details.

> do you need -f here in case it's mounted?

Hmm, that sounds dangerous. We definitely want to make sure it's completely unmounted before wiping signatures, right? (So we *want* it to fail in case our `umount` bit didn't kick in somehow). The reason I originally removed the `findmnt | grep | umount` pipeline was because there is nothing to unmount in the first place in the new CI setup. But I restored it to better cover the local dev cases.

giuseppe:
we have accumulated a few PRs that are currently blocked by the CI.  Let's merge this for now and revisit later if needed

giuseppe:
@rh-atomic-bot delegate=TomSweeneyRedHat

ashcrow:
@TomSweeneyRedHat merge away!

TomSweeneyRedHat:
@ashcrow I'm not authorized to merge.  Beyond which @rhatdan asked @rhvgoyal to PTAL.  If I could merge, I'd like to get a head nod from either or both of those folks first.

rhatdan:
LGTM

rhatdan:
@rh-atomic-bot r+

PR #1109, opened 2017-10-09
docs, pull: add documentation for http: prefix

giuseppe:
Signed-off-by: Giuseppe Scrivano <gscrivan@redhat.com>

PR #1114, opened 2017-10-11
syscontainers: Fall back to destination

ashcrow:
## Description

If /sysroot/ostree/deploy/rhel-atomic-host/var does not show the same
content destination then fall back destination.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1500961

## Related Bugzillas
- https://bugzilla.redhat.com/show_bug.cgi?id=1500961

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -991,7 +991,8 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         checkouts = self._get_system_checkout_path()
         if not os.path.exists(checkouts):
             os.makedirs(checkouts)
-        if not SystemContainers._are_on_the_same_filesystem("/ostree/repo", checkouts):
+        rootdir = "/ostree/repo" if os.path.exists("/ostree/repo") else "/"
+        if not SystemContainers._are_on_the_same_filesystem(rootdir, checkouts):
             return os.path.join(self.get_storage_path(skip_canonicalize=True), "ostree")
 
         return "/ostree/repo"
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -659,6 +659,12 @@ class SystemContainers(object):
         values["IMAGE_ID"] = image_id
         return values
 
+    @staticmethod
+    def _are_on_the_same_filesystem(a, b):
+        a_stat = os.stat(a)
+        b_stat = os.stat(b)
+        return a_stat.st_dev == b_stat.st_dev and a_stat.st_ino == b_stat.st_ino
+
     def _canonicalize_location(self, destination):
         # Under Atomic, get the real deployment location if we're using the
         # system repo. It is needed to create the hard links.
@@ -674,10 +680,10 @@ class SystemContainers(object):
 
             # Verify that content under the ostree destination shows up on destination.
             # If it does, we use the ostree destination.
-            os.makedirs(os.path.dirname(destination))
-            dest_stat = os.stat(os.path.dirname(destination))
-            ostree_stat = os.stat(os.path.dirname(ostree_destination))
-            if dest_stat.st_dev == ostree_stat.st_dev and dest_stat.st_ino == ostree_stat.st_ino:
+            if not os.path.exists(os.path.dirname(destination)):
+                os.makedirs(os.path.dirname(destination))
+            if SystemContainers._are_on_the_same_filesystem(os.path.dirname(destination),
+                                                            os.path.dirname(ostree_destination)):
                 destination = ostree_destination
 
         except: #pylint: disable=bare-except
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -981,7 +981,19 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         if self.user:
             return os.path.join(HOME, ".containers/repo")
 
-        return self.get_atomic_config_item(["ostree_repository"]) or "/ostree/repo"
+        if self.get_atomic_config_item(["ostree_repository"]) is not None:
+            return self.get_atomic_config_item(["ostree_repository"])
+
+        # If the checkout directory and the OSTree storage are not on the same
+        # file system, then we cannot use the system repository, as it would not
+        # support hard links checkouts.
+        checkouts = self._get_system_checkout_path()
+        if not os.path.exists(checkouts):
+            os.makedirs(checkouts)
+        if not SystemContainers._are_on_the_same_filesystem("/ostree/repo", checkouts):
+            return os.path.join(self.get_storage_path(skip_canonicalize=True), "ostree")
+
+        return "/ostree/repo"
 
     def _get_ostree_repo(self):
         if not OSTREE_PRESENT:
@@ -2137,7 +2149,7 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         repo.transaction_set_ref(None, get_image_branch(dest), rev)
         repo.commit_transaction(None)
 
-    def get_storage_path(self):
+    def get_storage_path(self, skip_canonicalize=False):
         """
         Returns the path to storage.
 
@@ -2145,6 +2157,8 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         :rtype: str
         """
         storage = os.path.sep.join([self._get_system_checkout_path(), ".storage"])
+        if skip_canonicalize:
+            return storage
         return self._canonicalize_location(storage)
 
     def _ensure_storage_for_image(self, repo, img):
@@ -2187,6 +2201,8 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         if not os.path.exists(storage):
             return
         for i in os.listdir(storage):
+            if i == "ostree":
+                continue
             branch = "{}{}".format(OSTREE_OCIIMAGE_PREFIX, i)
             rev_layer = repo.resolve_rev(branch, True)[1]
             if not rev_layer:
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -260,6 +260,7 @@ class SystemContainers(object):
         return_value = None
         # If we don't have a dockertar file or a reference to a docker engine image
         if not image.startswith('dockertar:/') and not (image.startswith("docker:") and image.count(':') > 1):
+            image = util.remove_skopeo_prefixes(image)
             labels = self.inspect_system_image(image).get('Labels', {})
             # And we have a run-once label
             if labels and labels.get('atomic.run') == 'once':
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -668,7 +668,18 @@ class SystemContainers(object):
             sysroot = OSTree.Sysroot()
             sysroot.load()
             osname = sysroot.get_booted_deployment().get_osname()
-            destination = os.path.realpath(os.path.join("/ostree/deploy/", osname, os.path.relpath(destination, "/")))
+            ostree_destination = os.path.realpath(
+                os.path.join("/ostree/deploy/", osname, os.path.relpath(
+                    destination, "/")))
+
+            # Verify that content under the ostree destination shows up on destination.
+            # If it does, we use the ostree destination.
+            os.makedirs(os.path.dirname(destination))
+            dest_stat = os.stat(os.path.dirname(destination))
+            ostree_stat = os.stat(os.path.dirname(ostree_destination))
+            if dest_stat.st_dev == ostree_stat.st_dev and dest_stat.st_ino == ostree_stat.st_ino:
+                destination = ostree_destination
+
         except: #pylint: disable=bare-except
             pass
         return destination
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1703,7 +1703,7 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         :rtype: tuple(bool, str)
         """
         # Remove prefixes which we use to map to skopeo args
-        real_image = utils.remove_skopeo_prefixes(image)
+        real_image = util.remove_skopeo_prefixes(image)
 
         if full_resolution:
             try:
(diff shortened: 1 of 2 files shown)


Atomic/syscontainers.py
```diff
@@ -669,6 +669,17 @@ def _canonicalize_location(self, destination):
             sysroot.load()
             osname = sysroot.get_booted_deployment().get_osname()
             destination = os.path.realpath(os.path.join("/ostree/deploy/", osname, os.path.relpath(destination, "/")))
```
giuseppe: if we don't overwrite destination here, we can still use it instead of hardcoding "/var" (which is not always correct as the same function is used in user mode as well)

Atomic/syscontainers.py
```diff
@@ -669,6 +669,17 @@ def _canonicalize_location(self, destination):
...
+            # Test that the ostree location and /var show the same file.
+            # If they do not then we must return /var.
+            test_file = str(uuid.uuid4())
+            with open(os.path.join(destination, '.' + test_file), 'w') as f:
```
giuseppe: destination might not exist when this function is called, so it would be safer to create the file as sibling of destination

Atomic/syscontainers.py
```diff
@@ -669,6 +669,17 @@ def _canonicalize_location(self, destination):
...
+                f.write('')
+            if not os.path.isfile('/var/.' + test_file):
```
giuseppe: same here, let's not assume `/var` but whatever 'destination' is pointing to

Atomic/syscontainers.py
```diff
@@ -669,6 +669,17 @@ def _canonicalize_location(self, destination):
...
+                util.write_out(
+                    '{} does not show the same content as /var. Falling back to /var'.format(destination))
```
giuseppe: let's print this only for debugging, as it will be printed each time when a separate `/var` is used.

Atomic/syscontainers.py
ashcrow: If we reuse destination here we will end up checking the same path we wrote out as destination is provided as a parameter for the function, then overridden within the block to be the ostree location.

Atomic/syscontainers.py
ashcrow: It think it should be as it's both a parameter to the function as well as set a few lines prior.

Atomic/syscontainers.py
ashcrow: Do you mean rename this instance of destination to something like ``ostree_destination``?

ashcrow:
@giuseppe updated per notes.

giuseppe:
thinking more of it.  Would something like:
```
a = os.stat(dirname(destination))
b = os.stat(dirname(ostree_destination))
can_use_ostree_destination = a.st_dev == b.st_dev && a.st_ino == b.st_ino
```

be enough?  I've not checked on AH, but I believe it should be enough

ashcrow:
@giuseppe actually, yeah, it should be. I'll update.

ashcrow:
It assumes that ``destination`` is passed in with a trailing ``/`` but I think that's acceptable.

Atomic/syscontainers.py
```diff
@@ -668,7 +668,17 @@ def _canonicalize_location(self, destination):
...
             sysroot.load()
             osname = sysroot.get_booted_deployment().get_osname()
-            destination = os.path.realpath(os.path.join("/ostree/deploy/", osname, os.path.relpath(destination, "/")))
+            ostree_destination = os.path.realpath(
+                os.path.join("/ostree/deploy/", osname, os.path.relpath(
+                    destination, "/")))
+
+            # Verify that content under the ostree destination shows up on destination.
+            # If it does, we use the ostree destination.
+            dest_stat = os.stat(os.path.dirname(destination))
```
giuseppe: the only issue I've observed is that on a fresh installation, `os.path.dirname(destination)` doesn't exist so the entire check fails and we fallback to `/var` in any case.  The fix is simply to ensure the directory exists, I tried with:
```
            os.makedirs(os.path.dirname(destination))
            dest_stat = os.stat(os.path.dirname(destination))
            ....
```
and it solves the issue I've observed.  LGTM from me with this fixup included.

ashcrow:
@giuseppe updated

giuseppe:
@rh-atomic-bot r+ c3fc81c

cgwalters:
We always want hardlinks right?  Otherwise the ostree bit is mostly pointless.  So I think we should move the location to e.g. `/usr/lib/atomic/containers`.

ashcrow:
``/usr/lib/atomic/containers`` sounds fine to me. @cgwalters ``/usr/lib/atomic/containers`` for all cases or the case specific to Atomic Host?

cgwalters:
I'd vote for both cases; system containers need to be tightly integrated with the host, and having them share storage with `/` hence makes sense to me.  I not uncommonly blow away everything in `/var/lib/docker`, but that's only safe because nothing there is required by system bootstrap.

giuseppe:
There is already some code for system containers preinstalled under /usr/lib/atomic/containers but they are treated as read only, so no modifications are allowed there.  Will `/usr/lib/atomic/containers` be writeable?

What about moving the ostree repository for system containers to `/var/lib/containers/atomic` (with a parent repo of /ostree/repo on AH)?

ashcrow:
@giuseppe I remember you noting there was a requirement for ``/var/`` to show the same result. Could we (easily) move to ``/usr/lib/atomic/containers/`` as a single location for both AH and non-AH systems?

cgwalters:
Yeah...a different approach would be to detect that `/var` is a mount point, and put the ostree repo there too.  (Though this gives up sharing storage with the host...though we need to fix [the selinux issue](https://github.com/flatpak/flatpak/issues/927) to really make that work).  As far as having `/usr` be writable, that's no problem, right now `/sysroot` is writable, and that's the same method that `rpm-ostree livefs` uses.  I'd like to make this a more formal API...see https://github.com/ostreedev/ostree/issues/1265 but for now one can rely on it.

giuseppe:
I still feel like we should support both configurations:
1) a system container is completely integrated into the system so it is deployed on `/usr/lib` and the systemd service file goes as well in `/usr/lib/systemd/system` (this is currently supported and atomic refuses to upgrade/manage them) and
2) system containers that are installed on top of the immutable system and are deployed under `/var`.

If we install the system containers under `/usr/lib` wouldn't we need to reinstall them all on an upgrade?

cgwalters:
Yeah duh...for AH/ostree based systems we'd indeed need to integrate `atomic` and `rpm-ostree` to teach the latter how to basically move the dir across updates.  (Well and really this opens up the can of worms about being able to do *both* simultaneously).

So I think my short term vote is - detect if `/var` is a mount, and if so, make the syscontainers ostree repo live in `/var` too.

giuseppe:
we can re-use this closed PR to achieve the migration of the repository:

https://github.com/projectatomic/atomic/pull/1105

The difference will be in https://github.com/projectatomic/atomic/pull/1105/commits/14e273c8937d035800095ab960d7c0b607716bbd#diff-47fc29e44fac9e9de875734500ac6759R974

Instead of checking for the existence of "/ostree/repo" we will have to check if `/var` is on a different file system (the stat comparison already done in this PR)

ashcrow:
@giuseppe should we close this PR and move to+reopen #1105?

giuseppe:
I think it is a good starting point, we do still need what is done here.  We can cherry-pick the patch here on top of yours.

What I'd do:

1) Simplify the check we are doing to just `ATOMIC_VAR` and `/ostree/repo`.
2 ) Reuse it instead of https://github.com/projectatomic/atomic/pull/1105/commits/14e273c8937d035800095ab960d7c0b607716bbd#diff-47fc29e44fac9e9de875734500ac6759R974

giuseppe:
@ashcrow could you cherry-pick the two commits from https://github.com/giuseppe/atomic/tree/1500961 ?

ashcrow:
@giuseppe done

giuseppe:
I've tested it on RHELAH 7.4.2 and it fixes the issue for me.

@ashcrow if it is fine for you too, please feel free to r+ it.

ashcrow:
@giuseppe Works on AH, not on non AH ...

```
$ sudo python atomic install --system --system-package=no -n rrr --storage ostree repo:8888/openshift3/cri-o:latest
FATA[0000] Invalid destination name ostree:repo:8888/openshift3/cri-o:latest@/var/lib/containers/atomic/.storage/ostree: invalid tag format 
```

ashcrow:
The error is coming from ``skopeo``:

```
$ skopeo --version
skopeo version 0.1.23 commit: 24510500d48f15b52ddfed6972a2a56452ef16b6
```

edsantiago:
Looks like https://bugzilla.redhat.com/show_bug.cgi?id=1496267

ashcrow:
Also fails with:

```
$ skopeo --version 
skopeo version 0.1.24-dev commit: 6ac96b7874815c8fb0ae7eff563942334db5d45d
```

ashcrow:
@edsantiago thanks! It does look like the same issue.

ashcrow:
Related https://github.com/containers/image/pull/342

miabbott:
> @giuseppe Works on AH, not on non AH ...

@ashcrow Sounds like we need an updated test case for this

ashcrow:
> @ashcrow Sounds like we need an updated test case for this

Agreed. I'm looking at the integration tests now to see where/how to throw one in.

ashcrow:
I did some manual testing and it is related to not having https://github.com/containers/image/pull/342. The way we'd test this would be to check against a registry that has a port. Do we have a public, non TLS registry to use? 

Another possible option would be to use ``ncat`` to create a forwarding port but that would be a bit nasty for testing.

giuseppe:
@ashcrow @edsantiago is it a duplicate of https://bugzilla.redhat.com/show_bug.cgi?id=1499274?  It should work with an updated Skopeo

miabbott:
>public, 

> non TLS

Pick one.  :)

Could a local registry be setup for the purposes of the test?

```
# docker pull docker.io/registry:2
# docker run -d -p 5000:5000 --name registry docker.io/registry:2
# docker pull docker.io/busybox
# docker tag docker.io/busybox localhost:5000/busybox:test
# docker push localhost:5000/busybox:test
```

You could tag whatever image to the registry, then teardown the registry after pulling/installing?

giuseppe:
while having more tests doesn't surely hurt, isn't 1499274 a Skopeo bug?  Skopeo can now be used to pull an image directly to the ostree storage without requiring atomic:

skopeo copy docker://busybox ostree:busybox:latest@/ostree/repo

We have already added some tests upstream: https://github.com/containers/image/pull/342/files#diff-79363df29db8db13844d50bb5b7f5e39R47 that should prevent this issue to happen again.

ashcrow:
@giuseppe it is, however I'm noticing with the fixed version ``atomic install --system --system-package=no -n NAME repo:8888/path/image:tag`` causes a failure. I just realized it's a *different* error that's causing the failure ...

```
sudo python atomic --debug install --system --system-package=no -n rrr --storage ostree repo:8888/openshift3/container-engine:latest

FATA[0000] Error initializing image from source docker://repo:8888/openshift3/container-engine:latest: pinging docker registry returned: Get https://repo:8888/v2/: http: server gave HTTP response to HTTPS client        
```
Adding ``http:`` in front causes an atomic failure. I believe this is an ``atomic`` issue that is caused from the new version of ``skopeo``.

ashcrow:
I'm checking out ``util.Decompose`` as it seems like by the time ``install`` the registry doesn't get set.

ashcrow:
I have a hacky patch I'm trying to make a little less hacky, but it works 😄

giuseppe:
@ashcrow it looks like: https://github.com/projectatomic/atomic/pull/1109

Does make any difference when you add the `http:` prefix?

ashcrow:
@giuseppe yeah it would fail. I just added another patch which strips the ``http:`` or ``https:`` from the image name when installing via a new function in util. This ended up working for me. Without the patch adding ``http:`` would pass ``skopeo`` but fail in the actual installation.

giuseppe:
the `http:` prefix is handled by atomic, but anyway it is just a detail in the comment.

I added a fixup patch (a3e22777b707aa1634ba55a471dbe93d4d49a951) to my branch, could you please cherry-pick it?

ashcrow:
@giuseppe done

ashcrow:
@giuseppe I did notice it was handled on pull, but on install it would pass through a different code path that wouldn't handle it properly. I looked at how it was being handled in ``syscontainers.py`` and made a simplified ``util`` version which is accessible outside of the ``syscontainers`` module.

Atomic/util.py
```diff
@@ -1096,3 +1099,18 @@ def kpod(cmd, storage=None, debug=None):
     if debug:
         write_out(" ".join(cmd))
     return check_output(_kpod, env=os.environ)
+
+
+def remove_skopeo_prefixes(image):
+    """
+    Remove prefixes used by skopeo but not expected in other image uses.
```
giuseppe: The prefix are not really used by Skopeo itself, but we map them to the Skopeo options.  I'd just s|by|with|

I am still not sure this should be in util.py as it is specific to system containers, could we move it to syscontainers.py?  On the other hand, I don't want to hold this PR any longer, so we can address this later

Atomic/util.py
```diff
@@ -1096,3 +1099,18 @@ def kpod(cmd, storage=None, debug=None):
...
+
+    :param image: The full image string
+    :type image: str
+    :returns: The image string without skopeo prefixes
+    :rtype: str
+    """
+    for remove in ('http:', 'https:'):
```
giuseppe: this version looks better than what we do in `_get_skopeo_args` (_syscontainers.py). Could you please add the `oci:` prefix as well to the list and replace the code in `_get_skopeo_args` with a call to this function?

ashcrow: Sure!

Atomic/util.py
ashcrow: I'll update the docstring.

ashcrow:
@giuseppe updated

giuseppe:
@rh-atomic-bot r+ 17c226a

PR #1119, opened 2017-10-24
syscontainers: introduce ALL_PROCESS_CAPABILITIES

giuseppe:
## Description
it can be used by privileged containers to add all the capabilities known to the host.  It prevents hard coding the list in the config.json.template file so that the same image can be used on different kernel versions where there can be different capabilities supported.

## Pull Request Checklist:

- [x] Unittests
- [x] Integration Tests

### DIFF
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -1130,7 +1130,7 @@ def get_all_known_process_capabilities():
 
     mask = hex((1 << (last_cap + 1)) - 1)
 
-    out = subprocess.check_output([CAPSH_PATH, '--decode=%s' % mask], stderr=DEVNULL)
+    out = subprocess.check_output([CAPSH_PATH, '--decode={}'.format(mask)], stderr=DEVNULL)
 
     # The output looks like 0x0000003fffffffff=cap_chown,cap_dac_override,...
     # so take only the part after the '='
--- a/Atomic/util.py
+++ b/Atomic/util.py
@@ -40,6 +40,7 @@ BWRAP_OCI_PATH = os.environ.get("BWRAP_OCI", "/usr/bin/bwrap-oci")
 RUNC_PATH = os.environ.get("RUNC", "/bin/runc")
 SKOPEO_PATH = os.environ.get("SKOPEO_PATH", "/usr/bin/skopeo")
 KPOD_PATH = os.environ.get("KPOD_PATH", "/usr/bin/kpod")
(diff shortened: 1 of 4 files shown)


Atomic/util.py
```diff
@@ -1115,3 +1116,26 @@ def remove_skopeo_prefixes(image):
         if image.startswith(remove):
             image = image.replace(remove, '')
     return image
+
+def get_all_known_process_capabilities():
+    """
+    Get all the known process capabilities
+
+    :returns: The list of known capabilities
```
ashcrow: 👍 Thanks for using fully documented doc strings!

Atomic/util.py
```diff
@@ -1115,3 +1116,26 @@ def remove_skopeo_prefixes(image):
...
+    :rtype: list
+    """
+
+    with open("/proc/sys/kernel/cap_last_cap", 'r') as f:
+        last_cap = int(f.read())
+
+    mask = hex((1 << (last_cap + 1)) - 1)
+
+    out = subprocess.check_output([CAPSH_PATH, '--decode=%s' % mask], stderr=DEVNULL)
```
ashcrow: nit: ``"--decode={}".format(mask)``

tests/test-images/system-container-files/config.json.template
```diff
@@ -114,11 +114,9 @@
 	],
 	"hooks": {},
 	"linux": {
-		"capabilities": [
-			"CAP_AUDIT_WRITE",
-			"CAP_KILL",
-			"CAP_NET_BIND_SERVICE"
-		],
+	         "capabilities": {
+			"bounding": [${ALL_PROCESS_CAPABILITIES}]
```
ashcrow: Is ``bounding`` supported for all the current versions of runc we are supporting?

ashcrow: Does this mean we will be allowing all capabilities supported by a system from now on?

giuseppe: we already dropped the support for the older OCI spec file support where there was no differentiation among the different kind of capabilities (i.e. `capabilities` was a list not a dictionary). 
runc honors this setting.

We don't allow all capabilities, this is meant only for privileged containers like CRI-O/Docker that anyway are already listing all the capabilities in their config.json.template file so that we don't need to carry two different lists on Fedora and RHEL since the capabilities sets are different.

/cc @rhatdan

giuseppe:

ashcrow:
@rh-atomic-bot r+ 6df8f07

PR #1120, opened 2017-10-25
pull: add support for credentials for accessing the source registry

giuseppe:
## Description

Allow to use a registry that needs credentials

## Related Bugzillas
https://bugzilla.redhat.com/show_bug.cgi?id=1505744

### DIFF
--- a/Atomic/backends/_docker.py
+++ b/Atomic/backends/_docker.py
@@ -338,12 +338,13 @@ class DockerBackend(Backend):
         if '@sha256:' in image:
             image = image.replace("@sha256:", ":")
 
+        src_creds = kwargs.get('src_creds')
         insecure = True if util.is_insecure_registry(self.d.info()['RegistryConfig'], registry) else False
         trust = Trust()
         trust.discover_sigstore(fq_name)
         util.write_out("Pulling {} ...".format(fq_name))
         util.skopeo_copy("docker://{}".format(fq_name), image, debug=debug, insecure=insecure,
-                         policy_filename=trust.policy_filename)
+                         policy_filename=trust.policy_filename, src_creds=src_creds)
         return 0
 
     def delete_container(self, container, force=False):
--- a/docs/atomic-pull.1.md
+++ b/docs/atomic-pull.1.md
@@ -78,6 +78,9 @@ pulling from an atomic registry could be:
 **-h** **--help**
 Print usage statement
 
+**--src-creds=USERNAME[:PASSWORD]**
+Define the credentials to use with the source registry.
+
 **--storage=[ostree|docker]**
 Define the destination storage for the pulled image.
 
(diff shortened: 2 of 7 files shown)


rhatdan:
LGTM

Atomic/backends/_docker.py
```diff
@@ -338,12 +338,13 @@ def pull_image(self, image, remote_image_obj, **kwargs):
...
             image = image.replace("@sha256:", ":")
 
+        src_creds = kwargs.get('src_creds')
         insecure = True if util.is_insecure_registry(self.d.info()['RegistryConfig'], registry) else False
         trust = Trust()
         trust.discover_sigstore(fq_name)
         util.write_out("Pulling {} ...".format(fq_name))
         util.skopeo_copy("docker://{}".format(fq_name), image, debug=debug, insecure=insecure,
-                         policy_filename=trust.policy_filename)
+                         policy_filename=trust.policy_filename, src_creds=src_creds)
```
TomSweeneyRedHat: Should you (or are you planning to later) add the destination creds too?

docs/atomic-pull.1.md
```diff
@@ -78,6 +78,9 @@ pulling from an atomic registry could be:
 **-h** **--help**
 Print usage statement
 
+**--src-creds=USERNAME[:PASSWORD]**
+Define the credentials to use with the source registry.
+
```
TomSweeneyRedHat: It would be nice to add a pull example with the creds in use.

TomSweeneyRedHat:
Changes LGTM, a few questions.   Also any plans to add a pointer to the cert-dir in order to use certificates?

Atomic/backends/_docker.py
giuseppe: with the storage types we support in atomic we don't write to any destination registry, so I think that is not necessary yet.  Or is there any case where skopeo_copy is used with a destination registry?

giuseppe:
@TomSweeneyRedHat thanks for the info, I'll take a look at that separately and submit a new PR.

Atomic/backends/_docker.py
TomSweeneyRedHat: @giuseppe It's possible for skopeo to copy to a destination registry, how much it's used in practice I don't know.  I'd guess it's not a common occurrence and probably less likely in the Atomic arena.  @mtrmac any thoughts?

giuseppe: yes, Skopeo copies to a destination registry but at the moment, AFAICS, this possibility is not plugged into `atomic pull`.  Until we have that supported, it makes no sense to expose dest-creds.

mtrmac: something→registry is called `push` in `atomic`; in this case, the destination (`image`) is always `docker-daemon:`, which does not need credentials (of this kind at least; it requires access to the socket).

giuseppe:

ashcrow:
@rh-atomic-bot r+ 9985518

PR #1121, opened 2017-10-26
docs: document system container variables

giuseppe:
not sure if it is the best location, but we should document this stuff


rhatdan:
LGTM

giuseppe:
added another patch to document the files used by system containers

baude:
@rh-atomic-bot r+ 5955252

PR #1124, opened 2017-11-06
syscontainers: add a warning if pulling a not fully qualified image

giuseppe:
Not fully qualified images are ambiguous in the ostree reference as
internally e.g. docker.io/busybox and busybox are stored as two
separate references.  Suggest the user to use a fully qualified name
to avoid the ambiguosity.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1954,6 +1954,15 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             if not missing_layers:
                 return False
 
+        try:
+            d = util.Decompose(img)
+            if d.registry == "":
+                util.write_err("The image `{}` is not fully qualified.  The default registry configured for Skopeo will be used.".format(img))
+        except:  # pylint: disable=bare-except
+            # This is used only to raise a warning to the user, so catch any exception
+            # and never let it fail.
+            pass
+
         can_use_skopeo_copy = util.check_output([util.SKOPEO_PATH, "copy", "--help"]).decode().find("ostree") >= 0
 
         if can_use_skopeo_copy:
--- a/docs/atomic-pull.1.md
+++ b/docs/atomic-pull.1.md
@@ -59,9 +59,9 @@ It tells Skopeo to not do TLS verification on the specified registry.
 
 `atomic pull --storage ostree http:REGISTRY/IMAGE:TAG`
 
-Not fully qualified images are supported when pulling to 'ostree',
-although it is suggested to use a fully qualified name to refer
-unambiguously to an image.
+Images where the registry is not specified are supported
+when pulling to 'ostree'.  However, we recommend that you use a
+fully qualified name to refer unambiguously to the image.
 
 If your /etc/containers/policy.json requires signature verification, the 
 pulled image is verified prior to being made available to the local docker
--- a/docs/atomic-pull.1.md
+++ b/docs/atomic-pull.1.md
@@ -59,6 +59,10 @@ It tells Skopeo to not do TLS verification on the specified registry.
 
 `atomic pull --storage ostree http:REGISTRY/IMAGE:TAG`
 
+Not fully qualified images are supported when pulling to 'ostree',
+although it is suggested to use a fully qualified name to refer
+unambiguously to an image.
+
 If your /etc/containers/policy.json requires signature verification, the 
 pulled image is verified prior to being made available to the local docker
 daemon. When interacting with a docker registry, Atomic uses the policy 


giuseppe:
bot, please retry

Atomic/syscontainers.py
```diff
@@ -1954,6 +1954,15 @@ def _check_system_oci_image(self, repo, img, upgrade, src_creds=None):
             if not missing_layers:
                 return False
 
+        try:
+            d = util.Decompose(img)
+            if d.registry == "":
+                util.write_err("The image `{}` is not fully qualified.  The default registry configured for Skopeo will be used.".format(img))
+        except:  # pylint: disable=bare-except
```
peterbaouoft: Hey, not a review, but a question here as I am relatively new to the code base. Hopefully you won't mind :p. I noticed that some of the expect code part, e.g: https://github.com/giuseppe/atomic/blob/4c45bee25c3bf23cce0d61143daf59bf2b58333f/Atomic/syscontainers.py#L865
does not have '#pylint: disable=bare-except' as a comment, so just wondering what will be the reason for this comment to be there? :).

Atomic/syscontainers.py
```diff
@@ -1954,6 +1954,15 @@ def _check_system_oci_image(self, repo, img, upgrade, src_creds=None):
             if not missing_layers:
                 return False
 
+        try:
+            d = util.Decompose(img)
+            if d.registry == "":
+                util.write_err("The image `{}` is not fully qualified.  The default registry configured for Skopeo will be used.".format(img))
```
peterbaouoft: A nit, hopefully you won't mind :). It seems a bit weird to me that in this function, it is also possible to perform non-skopeo copy right? However, this code section seems to me that it only belongs to skopeo part? Do we possibly want to move it to the code block into the if statement later:

```
if can_use_skopeo_copy:

       try:
            xxxx
       except:
            pass
       
       return self._check_system_oci_image_skopeo_copy(repo, img, src_creds=src_creds)
```

Or may be we could reword the error message here to cover non-skopeo case?

docs/atomic-pull.1.md
```diff
@@ -59,6 +59,10 @@ It tells Skopeo to not do TLS verification on the specified registry.
 
 `atomic pull --storage ostree http:REGISTRY/IMAGE:TAG`
 
+Not fully qualified images are supported when pulling to 'ostree',
```
peterbaouoft: Just to confirm, here the qualified image will be in the form of REGISTRY/IMAGE:TAG right? The reason why I aksed that was because ... https://github.com/projectatomic/atomic/pull/1124/files#diff-21fc15fd5e971e0a8aad60171d003206L26, the description there kinda confuses me.

Atomic/syscontainers.py
giuseppe: both code paths will end up in Skopeo, but in different ways.

The new way is to use "skopeo copy" that does basically everything: pulls the image, writes the references to the OSTree repository, check signatures.

The old way was to use "skopeo inspect", then check what layers we need and fetch them with "skopeo layers", that as you can see, it is even deprecated in newer Skopeo versions.

Atomic/syscontainers.py
giuseppe: we don't specify an exception type here, so pylint would complain:

http://pylint.blogspot.com/2014/08/no-exception-types-specified-pylint.html

But here I really want to be sure no exceptions are raised ever, it is only a warning, in the worse case we will just miss it.

docs/atomic-pull.1.md
giuseppe: yes exactly, what we really want is that the registry is specified as well

Atomic/syscontainers.py
peterbaouoft: ah I see, makes sense :-)

Atomic/syscontainers.py
peterbaouoft: ah, I see. Ignore my last comment then, did not notice the skopeo call in the codebase. Thanks for letting me know. :-)

docs/atomic-pull.1.md
peterbaouoft: May be we want to be a slightly more specific here about **qualified name** then? It is really my personal opinion tho, I feel we can tell the user what fully qualified name is as it is not mentioned explicitly anywhere in this doc? Tho, both way feels ok to me. Other than this nit, LGTM :-).

giuseppe: I have added a fixup patch, does it look better?

peterbaouoft: yup, LGTM =)

giuseppe:
@rh-atomic-bot delegate=peterbaouoft

Atomic/syscontainers.py
rhatdan: Do we need to respect the --quiet option?

docs/atomic-pull.1.md
```diff
@@ -59,9 +59,9 @@ It tells Skopeo to not do TLS verification on the specified registry.
 
 `atomic pull --storage ostree http:REGISTRY/IMAGE:TAG`
 
-Not fully qualified images are supported when pulling to 'ostree',
-although it is suggested to use a fully qualified name to refer
-unambiguously to an image.
+Not fully specified images, where the registry is not specified, are
```
rhatdan: This does not read well.
```
Images where the registry is not specified are supported 
when pulling to 'ostree'.  However, we recommend that you use a
fully qualified name to refer unambiguously to the image.```

Comment (giuseppe) [reply docs/atomic-pull.1.md]: thanks, I have update the patch

File: Atomic/syscontainers.py (hunk 2 above)
Comment (giuseppe) [reply Atomic/syscontainers.py]: it seems we have `quiet` only for `atomic run` and `atomic images|containers list`, not other command is supporting it AFAICS:

```
# atomic pull --quiet foo
atomic: unrecognized arguments: --quiet

# atomic install --quiet foo
atomic: unrecognized arguments: --quiet
```

Or as a global option:

```
# atomic --quiet install foo
atomic: unrecognized arguments: --quiet
```

rhatdan: We probably should add support for them.  ::)

peterbaouoft:
LGTM, tho, I am still relatively new to the repo/codebase, so I will let other reviewers to take a look as well. =)

giuseppe:
@peterbaouoft it is a quite small and simple change as it adds only a warning, feel free to r+ it if you think it is fine

peterbaouoft:
sure =) 
@rh-atomic-bot  r+ 843ffaa

PR #1131, opened 2017-11-17
syscontainers: do not delete modified files with --system-package=no

giuseppe:
Store the checksum of files copied to the host, so that on an update or
an uninstall we can skip the files that have a mismatch in the
checksum.

This reflects in the --system-package=no case what the rpm backend
already does.

## Pull Request Checklist:

- [ ] Unittests
- [x] Integration Tests

### DIFF
--- a/Atomic/rpm_host_install.py
+++ b/Atomic/rpm_host_install.py
@@ -90,7 +90,8 @@ class RPMHostInstall(object):
                         dest_path = os.path.join(prefix or "/", os.path.relpath(rel_dest_path, "/"))
 
                     if os.path.exists(dest_path):
-                        util.write_out("File %s already exists." % dest_path, lf="\n")
+                        if os.path.isfile(dest_path):
+                            util.write_out("File %s already exists" % os.path.normpath(dest_path), lf="\n")
                         continue
 
                     if not os.path.exists(os.path.dirname(dest_path)):
--- a/Atomic/rpm_host_install.py
+++ b/Atomic/rpm_host_install.py
@@ -59,7 +59,7 @@ class RPMHostInstall(object):
                 new_checksum = RPMHostInstall.file_checksum(path)
                 if new_checksum != checksum:
                     # Do not delete the file if it was modified.
-                    util.write_out("Do not delete %s as it was manually modified." % path, lf="\n")
+                    util.write_out("Will not delete %s as it was manually modified." % path, lf="\n")
                     continue
                 try:
                     os.remove(path)
--- a/Atomic/rpm_host_install.py
+++ b/Atomic/rpm_host_install.py
@@ -43,12 +43,12 @@ class RPMHostInstall(object):
             return "0"
 
         h = hashlib.sha256()
-        with open(path, "r") as f:
+        with open(path, "rb") as f:
             while True:
                 chunk = f.read(blocksize)
                 if chunk == None or len(chunk) == 0:
                     break
-                h.update(chunk.encode())
+                h.update(chunk)
         return h.hexdigest()
 
     @staticmethod
--- a/Atomic/rpm_host_install.py
+++ b/Atomic/rpm_host_install.py
@@ -4,6 +4,7 @@ from . import util
 from . import rpmwriter
 import tempfile
 import shutil
+import hashlib
 
 RPM_NAME_PREFIX = "atomic-container"
 
@@ -12,16 +13,20 @@ class RPMHostInstall(object):
     @staticmethod
     def copyfile(src, dest):
         if os.path.isdir(src):
-            # add the directory only if it is empty, so we don't create directories that
+            # add the directory only if it is empty, so we don't delete directories that
             # weren't added by us.  Anyway a non empty directory would be created by
             # its children.
             if len(os.listdir(src)) == 0:
                 os.mkdir(dest)
+                return True
         elif os.path.islink(src):
             linkto = os.readlink(src)
             os.symlink(linkto, dest)
+            return True
         else:
             shutil.copy2(src, dest)
+            return True
+        return False
 
     @staticmethod
     def _do_rename_path(path, rename_files):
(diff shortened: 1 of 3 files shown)


lsm5:
umm so is this similar to `%config(noreplace)` in rpms?

giuseppe:
@lsm5 yes, correct.  We don't throw away changes done by the user when we upgrade a container

Atomic/syscontainers.py
```diff
@@ -1597,9 +1606,12 @@ def uninstall(self, name):
         installed_files = None
         with open(os.path.join(checkout, name,  "info"), 'r') as info_file:
             info = json.loads(info_file.read())
-            installed_files = info["installed-files"] if "installed-files" in info else None
-        if installed_files:
-            RPMHostInstall.rm_add_files_to_host(installed_files, None)
+            installed_files_checksum = info["installed-files-checksum"] if "installed-files-checksum" in info else None
+            installed_files = info["installed-files"] if "installed-files" in info else []
```
peterbaouoft: Minor: seems like the assignment of checksum here is a bit inconsistent with other implementations, 
```
 installed_files_checksum = info["installed-files-checksum"] if "installed-files-checksum" in info else None
        if installed_files_checksum is None:
            installed_files = info["installed-files"] if "installed-files" in info else None
            installed_files_checksum = {k : "" for k in installed_files}

```
any reasons for that?

Atomic/rpm_host_install.py
```diff
@@ -12,16 +13,20 @@ class RPMHostInstall(object):
     @staticmethod
     def copyfile(src, dest):
         if os.path.isdir(src):
-            # add the directory only if it is empty, so we don't create directories that
+            # add the directory only if it is empty, so we don't delete directories that
```
peterbaouoft: **Q**, hmm, I don't quite get ... how "delete" got involved in this function, can you explain it a bit :).

Atomic/rpm_host_install.py
```diff
@@ -33,12 +38,31 @@ def _do_rename_path(path, rename_files):
...
+        if not os.path.exists(path) or os.path.isdir(path):
+            return "0"
+
+        h = hashlib.sha256()
+        with open(path, "r") as f:
+            while True:
+                chunk = f.read(blocksize)
+                if chunk == None or len(chunk) == 0:
+                    break
+                h.update(chunk.encode())
```
peterbaouoft: Q: why do we have an encode here? would it make a difference if we update with chunk directly?

tests/integration/test_system_containers_rpm.sh
```diff
@@ -149,3 +150,11 @@ for i in /usr/local/lib/renamed-atomic-test-system-hostfs /usr/local/lib/secret-
 do
     test -e $i
 done
+
+echo "This message will not be deleted" > /usr/local/lib/secret-message
+
+ATOMIC_OSTREE_TEST_FORCE_IMAGE_ID=NEW-ID ${ATOMIC} containers update atomic-test-system-hostfs
+
```
peterbaouoft: Minor: Do you think it is better to also include a check to see if the error message is there to notify the user? 
I.e: We check if `Do not delete xxx as it was manually modified`  is outputted?

Atomic/rpm_host_install.py
giuseppe: we keep track of what files/directories we are adding.  We add the directory only if it is not already present.
In this way when we uninstall the container we will not delete directories that were already present when we first installed the container

Atomic/rpm_host_install.py
giuseppe: good point, it is probably not needed.  But I spotted anothe issue, the file must be opened with 'rb'.  I will fix it

Atomic/syscontainers.py
giuseppe: yes, it is done a bit differently so if installed_files_checksum is not present (for containers installed before this patch), we assume it is empty:

`installed_files_checksum = {k: "" for k in installed_files}`

tests/integration/test_system_containers_rpm.sh
giuseppe: I do that no?

There is:

```
+assert_matches "This message will not be deleted" /usr/local/lib/secret-message
```

peterbaouoft:
Thanks for answering the questions. LGTM =). Tho, I am still relatively new to the code base, so I will wait for other reviewers' opinion as well. :p

Atomic/rpm_host_install.py
```diff
@@ -33,12 +38,31 @@ def _do_rename_path(path, rename_files):
...
+    def rm_add_files_to_host(old_installed_files_checksum, exports, prefix="/", files_template=None, values=None, rename_files=None):
         # if any file was installed on the host delete it
-        if old_installed_files:
-            for i in old_installed_files:
+        if old_installed_files_checksum:
+            for path, checksum in old_installed_files_checksum.items():
+                new_checksum = RPMHostInstall.file_checksum(path)
+                if new_checksum != checksum:
+                    # Do not delete the file if it was modified.
+                    util.write_out("Do not delete %s as it was manually modified." % path, lf="\n")
```
ashcrow: nit: It seems this should read "Will not delete....".

Atomic/syscontainers.py
```diff
@@ -526,7 +526,7 @@ def _resolve_remote_path(self, remote_path):
             raise ValueError("The container's rootfs is set to remote, but the remote rootfs does not exist")
         return real_path
 
-    def _checkout(self, repo, name, img, deployment, upgrade, values=None, destination=None, extract_only=False, remote=None, prefix=None, installed_files=None, system_package='no'):
+    def _checkout(self, repo, name, img, deployment, upgrade, values=None, destination=None, extract_only=False, remote=None, prefix=None, installed_files_checksum=None, system_package='no'):
```
ashcrow: Nothing to do in this PR but this method is getting a bit out of hand with it's arguments!

giuseppe:
I believe (and hope) so :-)

giuseppe:
@ashcrow pushed a fixup patch here ⬆️

ashcrow:
@rh-atomic-bot r+ 73d9191

PR #1133, opened 2017-11-24
syscontainers: fix repo location detection

giuseppe:
Improve the detection of the repo being on the same file system as the checkout path by trying to link a file.  This is necessary since on a bind mount the st_dev for the two directories is the same, but hardlinks will still fail with EXDEV.

Also, change the test in _canonicalize_location to check that on an Atomic Host /sysroot/ostree/deploy/*/var/* and /var/* have the same st_dev and st_ino.  It is necessary so that files created under /sysroot are visible in /var, and in this case it is fine the two locations are on different file systems.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -119,6 +119,7 @@ class SystemContainers(object):
         options.overwrite_mode = OSTree.RepoCheckoutOverwriteMode.UNION_FILES
         options.process_whiteouts = True
         options.disable_fsync = True
+        options.no_copy_fallback = True
         if self.user:
             options.mode = OSTree.RepoCheckoutMode.USER
         repo.checkout_at(options, rootfs_fd, rootfs, rev)
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -10,6 +10,7 @@ import shutil
 import stat # pylint: disable=bad-python3-import
 import subprocess
 import time
+import errno
 from .client import AtomicDocker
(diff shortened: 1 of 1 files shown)


Atomic/syscontainers.py
```diff
@@ -675,7 +678,45 @@ def _amend_values(self, values, manifest, name, image, image_id, destination, pr
...
+        :type destdir: str
+        """
+        repo_stat = os.stat(repo)
+        destdir_stat = os.stat(destdir)
+        # Simple case: st_dev is different
+        if repo_stat.st_dev != destdir_stat.st_dev:
+            return False
+
+        try:
+            os.makedirs(destdir)
```
ashcrow: Won't this leave a created directory behind if everything succeeds? Not the worst thing in the world if it does, but since the ``dest_file`` is cleaned up later I'd expect this to be cleaned up. If it's meant to stick around maybe change the name of the function as it sounds like it shouldn't have a side effect.

giuseppe: I didn't add the rmdir as in any case more containers will still use the same directory (which is hidden in the .storage path).   It is more of a tmp area where we can write temporary files.  In any case, `atomic images prune` takes care of all the temporary files left in the storage (in case the unlink for the dest_file fails)

ashcrow: Ok, fair enough.

ashcrow:
@rh-atomic-bot r+ 7dcf774

PR #1136, opened 2017-11-29
syscontainers: implement "atomic containers|image update --all"

giuseppe:
## Description

allow to upgrade all the system containers witha single command.  Differently than a normal "atomic containers update", this operation also attempts to pull the latest image version.

## Pull Request Checklist:

- [x] Unittests
- [x] Integration Tests

### DIFF
--- a/Atomic/containers.py
+++ b/Atomic/containers.py
@@ -70,10 +70,11 @@ def cli(subparser):
                                               help=_("update a container"),
                                               epilog="Update the container to use a newer image.")
     updatep.set_defaults(_class=Containers, func='update')
-    updatep.add_argument("container",
+    updatep.add_argument("container", nargs='?',
                          help=_("Specify one or more containers. Must be final arguments."))
     updatep.add_argument("--rebase", dest="rebase", default=None,
                          help=_("Rebase to a different image (useful for upgrading to a different tag)"))
+    updatep.add_argument("-a", "--all", action='store_true', dest="all", default=False, help=_("Update all containers"))
     if OSTREE_PRESENT:
         updatep.add_argument("--set", dest="setvalues",
                              action='append',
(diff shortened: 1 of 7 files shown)


Atomic/containers.py
```diff
@@ -289,7 +290,70 @@ def _mark_vulnerable(self, containers):
...
+    def update_all_containers(self):
+        could_not_pull = {}
+        preupdate_containers = self.syscontainers.get_containers()
+        for i in preupdate_containers:
+            img = i['Image']
+            pull_location = i['PullLocation']
+            name = i['Names'][0]
+            d = util.Decompose(img)
+            if d.registry == "":
+                util.write_out("Image {} not fully qualified: skipping {}".format(img, name))
```
rhatdan: Shouldn't this be write_err?

Atomic/containers.py
```diff
@@ -289,7 +290,70 @@ def _mark_vulnerable(self, containers):
...
+                continue
+            try:
+                self.syscontainers.pull_image(pull_location)
+            except:  # pylint: disable=bare-except
+                could_not_pull[img] = True
+                util.write_out("Image {} could not be pulled.  You may need to manually update {}".format(img, name))
```
rhatdan: write_err?

Atomic/containers.py
```diff
@@ -289,7 +290,70 @@ def _mark_vulnerable(self, containers):
...
+        for cnt in preupdate_containers_by_name.keys():
+            pre_version = get_image_id(preupdate_containers_by_name[cnt])
+            post_version = get_image_id(postupdate_containers_by_name[cnt])
+            img_name = img_id = None
+            try:
+                img_name = preupdate_containers_by_name[cnt]['Image']
+                img_id = get_image_id(images_by_name[img_name])
+            except KeyError:
+                pass
+            status = get_status(img_name, pre_version, post_version, img_id)
```
rhatdan: Does thid work with img_name and/or img_id being None?

Atomic/containers.py
```diff
@@ -289,7 +290,70 @@ def _mark_vulnerable(self, containers):
...
+            util.write_out(cols.format(cnt[:15], pre_version[:32], post_version[:32], (img_id or "<NOT FOUND>")[:32], status[:15]))
+
     def update(self):
+        if self.args.all:
```
rhatdan: Should you throw an error if self.args.all is set and self.args.container is not None?

Atomic/syscontainers.py
```diff
@@ -999,10 +1004,22 @@ def _do_checkout(self, repo, name, img, upgrade, deployment, values, destination
...
             elif was_service_active:
-                self._systemctl_command("start", name)
+                if mode == SystemContainers.CHECKOUT_MODE_UPGRADE_CONTROLLED:
+                    must_rollback = False
+                    try:
+                        self._systemctl_command("start", name)
+                    except subprocess.CalledProcessError:
+                        must_rollback = True
+                    if must_rollback:
+                        util.write_out("Could not restart {}.  Attempt automatic rollback".format(name))
```
rhatdan: write_err?

rhatdan:
@baude PTAL

giuseppe:
Thanks, I've fixed the reported issues in the two additional fixup commits

rhatdan:
LGTM

Atomic/containers.py
```diff
@@ -289,7 +290,72 @@ def _mark_vulnerable(self, containers):
...
+                continue
+
+            self.syscontainers.update_container(name, controlled=True)
+        postupdate_containers = self.syscontainers.get_containers()
+
+        images_by_name = {i['RepoTags'][0] : i for i in self.syscontainers.get_system_images()}
+        preupdate_containers_by_name = {i['Names'][0] : i for i in preupdate_containers}
+        postupdate_containers_by_name = {i['Names'][0] : i for i in postupdate_containers}
+
+        def get_image_id(c): return c['ImageId'] if 'ImageId' in c else c['ImageID']
```
peterbaouoft: hmm is `ImageId` included in container part? I don't think I saw it in the code base. 
https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L1308,
seems like only ImageID was there?

Atomic/syscontainers.py
```diff
@@ -999,10 +1004,22 @@ def _do_checkout(self, repo, name, img, upgrade, deployment, values, destination
             if (tmpfiles_template):
                 self._systemd_tmpfiles("--create", tmpfilesout)
 
-            if not upgrade:
+            if mode == SystemContainers.CHECKOUT_MODE_INSTALL:
                 self._systemctl_command("enable", name)
             elif was_service_active:
-                self._systemctl_command("start", name)
+                if mode == SystemContainers.CHECKOUT_MODE_UPGRADE_CONTROLLED:
```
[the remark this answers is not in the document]
giuseppe: if the user is using `atomic containers update` then even if the service cannot be started, we should leave the version the user asked for.  It leaves the opportunity to the user to fix what is wrong and let the service starts.

The new `UPGRADE_CONTROLLED` mode is for the bulk update of all the containers, where we don't want to break if a service cannot be re-started automatically but simply report it.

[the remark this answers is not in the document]
peterbaouoft: makes sense =), thanks.

giuseppe:
I am still undecided how this should look, thinking more of it, I'd probably like to split the pull and the update.

Instead of having `atomic containers update --all` which also tries to pull the images, we could split it in:

`atomic images update --all && atomic containers update --all`

This will make `atomic containers update --all` not behave different from `atomic containers update` which doesn't do any pull of the images.

One thing I don't like of having all the logic in `atomic images containers update` is that stored images that have not a container installed will not be updated.

And the split in general will give more flexibility.

What do you think?

giuseppe:
so I changed my mind about this, I think we really need to split the two operations, it will also make easier to detect when a container can be updated.  I am going to push a new version soon.

giuseppe:
is this good to be merged?

giuseppe:
ping

baude:
@rh-atomic-bot r+

PR #1141, opened 2017-12-04
syscontainers: add two new variables RUNTIME and ATOMIC

giuseppe:
## Description

They allow the creation of wrapper scripts in a much easier way:

For example a script `ls` exported to the host, but that is executed in the container could look like:

```
#!/bin/sh
exec $ATOMIC run $NAME ls $$@
```

the advantage of using ATOMIC compared to RUNTIME is that "atomic run" has more features compared to directly using the container runtime exec command, like the possibility of running commands when the container is not already running.

## Pull Request Checklist:

- [ ] Unittests
- [x] Integration Tests

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -468,26 +468,28 @@ class SystemContainers(object):
                         missing_source_paths.append(source)
         return missing_source_paths
 
+
     def _generate_default_oci_configuration(self, destination):
         conf_path = os.path.join(destination, "config.json")
-
-        # If runc is not installed we are not able to generate the default configuration,
-        # write an empty JSON file
-        if not util.runc_available():
+        try:
+            # Try to use $RUNTIME spec to generate a default configuration file.
+            args = [self._get_oci_runtime(), 'spec']
+            util.subp(args, cwd=destination)
+            with open(conf_path, 'r') as conf:
+                configuration = json.loads(conf.read())
+            configuration['root']['readonly'] = True
+            configuration['root']['path'] = 'rootfs'
+            configuration['process']['terminal'] = False
+            configuration['process']['args'] = ['run.sh']
             with open(conf_path, 'w') as conf:
-                conf.write('{}')
+                conf.write(json.dumps(configuration, indent=4))
             return
-
-        args = [self._get_oci_runtime(), 'spec']
-        util.subp(args, cwd=destination)
-        with open(conf_path, 'r') as conf:
-            configuration = json.loads(conf.read())
-        configuration['root']['readonly'] = True
-        configuration['root']['path'] = 'rootfs'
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM
@baude PTAL

Atomic/syscontainers.py
```diff
@@ -468,26 +468,28 @@ def _check_oci_configuration_file(self, conf_path, remote=None, include_all=Fals
...
+
     def _generate_default_oci_configuration(self, destination):
         conf_path = os.path.join(destination, "config.json")
-
-        # If runc is not installed we are not able to generate the default configuration,
-        # write an empty JSON file
-        if not util.runc_available():
+        try:
+            # Try to use $RUNTIME spec to generate a default configuration file.
+            args = [self._get_oci_runtime(), 'spec']
```
peterbaouoft: Hmm, for the run time, could it be bwrap as a runtime here ( e.g: when self.user = true)?  Also, for function `_get_oci_runtime()`, are the only runtimes we have are bwrap and runc?

Atomic/syscontainers.py
```diff
@@ -2415,9 +2415,6 @@ def container_exec(self, name, detach, args):
 
         tty = os.isatty(0)
 
-        if util.is_user_mode():
-            raise ValueError("Command not supported in user mode")
```
peterbaouoft: Hi, I am slightly confused with the `_get_oci_runtime` function. This question is a bit similar to the previous question. Could the run time that we get from `_get_oci_runtime` still be bwrap? If so, the command will fail later because bwrap does not handle exec? (guessing).

Atomic/syscontainers.py
giuseppe: yes it could be bwrap-oci, that is something `_get_oci_runtime()` could return

Atomic/syscontainers.py
giuseppe: yes, but more in general we allow to override it with anything else.  The exec could still fail or not be implemented as in the `bwrap-oci` case. 

In any case, since we don't know if the runtime supports exec or not, we should not check for it here.

Atomic/syscontainers.py
peterbaouoft: ah, just making sure I get it right, so the idea is that we do not handle error handling /runtime checking in our end anymore. Instead, if the run time does not support spec/exec (i.e bwrap spec), we let bwrap/subprocess call fail and let the subprocess report error?

giuseppe: yes correct.  Since we can't know in advance whether the runtime supports that operation or not.

giuseppe:
Fine to merge?

baude:
@rh-atomic-bot r+

Issue #1146, opened 2017-12-08
"atomic run openshift/origin-ansible" result into 'NoneType' object has no attribute 'get_label'


mscherer:
Same as #1144 , I tried the following command:
```
# atomic run openshift/origin-ansible
```

After downloading everything, it show this:
```
Need to pull openshift/origin-ansible
Pulling docker.io/openshift/origin-ansible:latest ...
Copying blob sha256:85432449fd0fabb6bef4772246d9b7948723275a4c7e4dee18728f28d79f1c2c
 65.93 MB / 70.03 MB [=====================================================>---]
Copying blob sha256:d0cf7642e3d8ee1a5bb9ed967641bf46888582f646b72ec4ce38378bd470a07b
 0 B / 1.13 KB [---------------------------------------------------------------]
Copying blob sha256:eee11c7fbb7ab2101dd113b9fa9a2080dcc845c502eb06f25ff2a2bfee3af099
 138.54 MB / 140.28 MB [=======================================================]
Copying blob sha256:879d7900e49063e3547d2a34dc20b340d683ecf6383dd625c3b83f0798915fb4
 0 B / 7.02 KB [---------------------------------------------------------------]
Copying blob sha256:6ebced23ab9dfca7fcd57e827299b8bd50bdcd2e485a5a4e45a1c1ba6bc92de2
 2.72 MB / 3.01 MB [=====================================================>-----]
Copying blob sha256:4550e37cf10e7d9446e916a298ba0ea8fc05b34054cbf230092c54bb3db845f5
 0 B / 583 B [-----------------------------------------------------------------]
Copying config sha256:15a8a3d3ec573c6ad97e2752fa93da9b27ae637056c7bb5a84c8baf704ec640d
 0 B / 7.32 KB [---------------------------------------------------------------]
Writing manifest to image destination
Storing signatures
 7.32 KB / 7.32 KB [===========================================================]'NoneType' object has no attribute 'get_label'
```

The traceback is the following (using a modified atomic to display it, due to #1144):
```
  File "/tmp/atomic", line 203, in <module>
    sys.exit(_func())
  File "/usr/lib/python3.6/site-packages/Atomic/run.py", line 123, in run
    return be.run(img_object, atomic=self, args=self.args)
  File "/usr/lib/python3.6/site-packages/Atomic/backends/_docker.py", line 584, in run
    if iobject.get_label('INSTALL') and not args.ignore and not util.InstallData.image_installed(iobject):
[...]

PR #1147, opened 2017-12-09
Do not assume that util.skopeo_copy can't fail

mscherer:
## Description
If the disk is full, atomic will continue and show errors later, which is puzzling and hard to debug.

## Related Bugzillas
- 
-

## Related Issue Numbers
- https://github.com/projectatomic/atomic/issues/1146
-

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests


Atomic/run.py
```diff
@@ -108,7 +108,9 @@ def run(self):
             if self.args.display:
                 return 0
             try:
-                db.pull_image(self.image, remote_image_obj)
+                err = db.pull_image(self.image, remote_image_obj)
+                if err:
+                    raise ValueError("Unable to pull image {}"..format(self.image))
```
giuseppe: there is one extra '.' here

Atomic/backends/_docker.py
```diff
@@ -343,9 +343,8 @@ def pull_image(self, image, remote_image_obj, **kwargs):
         trust = Trust()
         trust.discover_sigstore(fq_name)
         util.write_out("Pulling {} ...".format(fq_name))
-        util.skopeo_copy("docker://{}".format(fq_name), image, debug=debug, insecure=insecure,
+        return util.skopeo_copy("docker://{}".format(fq_name), image, debug=debug, insecure=insecure,
```
giuseppe: doesn't it raise a `CalledProcessError` exception when `skopeo_copy` fails?

Atomic/run.py
giuseppe: doesn't it fail immediately once `pull_image` raises `CalledProcessError`?

mscherer: Nice catch, I will fix. And no, it doesn't seems to raise anything. I sent the PR first to see the tests running, but I didn't had time to test or validate it (as I reinstalled my system with enough disk).

Issue #1149, opened 2017-12-14
Start refactoring for syscontainers.py


peterbaouoft:
We may want to start the refactoring process for syscontainers.py.(I am gonna do it =) if no objections). First I think we can start refactoring by removing potential duplicated code sections, and make those duplicated parts become a function. WDYT? Also, here are some duplicated code sections that I think we could start moving them into a function:

1: in function run_once and _do_checkout, the code share a bit of similarity
https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L330-L365
https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L863-L867 and https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L875-L885

2: Another example of duplication I found is installed files in 'rollback', 'update_container' and 'uninstall':
https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L1657-L1664
https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L1189-L1195
https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L1128-L1137

P.S: Since I am relatively new to the code base, I might make some poor assumptions/mistakes while I was doing the refactoring. If you happen to see these errors/mistakes, feel free to correct me =).

peterbaouoft:
Another part of the refactoring would be to break `do_checkout` function logic down into multiple readable pieces. Then the users who read the code will benefit from the separation of the logic.

peterbaouoft:
It would also be nice that we can start to break the long parameter list down into dictionaries or classes. For instance, in function _do_checkout, we can break the parameter list into the following:

```
Before: 
def _do_checkout(self, repo, name, img, upgrade, deployment, values, destination, unitfileout,
                     tmpfilesout, extract_only, remote, prefix=None, installed_files_checksum=None, system_package='no'):
```

```
After:
img_container_info= {name, img, repo, deployment}
user_input_file_locations = {destination, unitfileout, tmpfilesout}
user_input_options = {upgrade, extract_only, remote, prefix, system_package}
user_input_values = {installed_files_checksum, values}
def _do_checkout (self, img_container_info, user_input_file_locations, user_input_options, user_input_values)
```

peterbaouoft:
[...]

PR #1152, opened 2017-12-28
syscontainers: refactor part of remote logic into a function

peterbaouoft:
The code logic for getting remote path was "refactored" into a new
function to shorten the length of 'do_checkout' method.

Remote_rootfs is also discarded as it appears that when 'remote_path' is
specified, the rootfs location will represent the remote_rootfs location

There shouldn't be any functional impact from this commit

Feedback is always welcome =).
~**Note: the unit test is in progress, first submit without it here for comments/testing purposes**~
~( Although I am not too sure if unit tests is needed here, it will be great if someone could clarify the~
~situations where a unit test would best fit :p)~

## Related Issue Numbers
-

## Pull Request Checklist:

- [x] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -293,15 +293,22 @@ class SystemContainers(object):
         :param remote_input: input path from user
         :returns: the parsed remote location
         """
-        remote_path = self._resolve_remote_path(remote_input)
-        if remote_path:
-            remote_rootfs = os.path.sep.join([remote_path, "rootfs"])
-            if os.path.exists(remote_rootfs):
-                util.write_out("The remote rootfs for this container is set to be {}".format(remote_rootfs))
-            elif os.path.exists(os.path.sep.join([remote_path, "usr"])):  # Assume that the user directly gave the location of the rootfs
-                remote_path = os.path.dirname(remote_path)  # Use the parent directory as the "container location"
-            else:
-                raise ValueError("--remote was specified but the given location does not contain a rootfs")
+
+        if not remote_input:
+            return None
+
+        remote_path = os.path.realpath(remote_input)
+        if not os.path.exists(remote_path):
+            raise ValueError("The container's rootfs is set to remote, but the remote rootfs does not exist")
+
+        # Here we know remote path does exist on the fs
+        remote_rootfs = os.path.sep.join([remote_path, "rootfs"])
+        if os.path.exists(remote_rootfs):
+            util.write_out("The remote rootfs for this container is set to be {}".format(remote_rootfs))
+        elif os.path.exists(os.path.sep.join([remote_path, "usr"])):  # Assume that the user directly gave the location of the rootfs
+            remote_path = os.path.dirname(remote_path)  # Use the parent directory as the "container location"
+        else:
+            raise ValueError("--remote was specified but the given location does not contain a rootfs")
         return remote_path
 
     def install(self, image, name):
@@ -567,15 +574,6 @@ class SystemContainers(object):
                 unitfileout = os.path.join(SYSTEMD_UNIT_FILES_DEST, "%s.service" % name)
                 tmpfilesout = os.path.join(SYSTEMD_TMPFILES_DEST, "%s.conf" % name)
         return unitfileout, tmpfilesout
-
-    def _resolve_remote_path(self, remote_path):
-        if not remote_path:
-            return None
-
-        real_path = os.path.realpath(remote_path)
-        if not os.path.exists(real_path):
-            raise ValueError("The container's rootfs is set to remote, but the remote rootfs does not exist")
-        return real_path
 
     def _checkout(self, repo, name, img, deployment, upgrade, values=None, destination=None, extract_only=False, remote=None, prefix=None, installed_files_checksum=None, system_package='no'):
         destination = destination or os.path.join(self._get_system_checkout_path(), "{}.{}".format(name, deployment))
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -311,6 +311,27 @@ class SystemContainers(object):
             raise ValueError("--remote was specified but the given location does not contain a rootfs")
         return remote_path
 
+    @staticmethod
+    def _rewrite_rootfs(destination, remote_rootfs):
+        """
+        When remote rootfs is specified, we rewrite the ['root']['path']
+
+        :param destination: the destination of the container
+        :param remote_rootfs: remote rootfs location
+        """
+        destination_path = os.path.join(destination, "config.json")
+        with open(destination_path, 'r') as config_file:
+            try:
+                config = json.loads(config_file.read())
+            except ValueError:
+                raise ValueError("Invalid config.json file in given remote location: {}.".format(destination_path))
+            config['root']['path'] = remote_rootfs
+        with open(destination_path, 'w') as config_file:
+            config_file.write(json.dumps(config, indent=4))
+        # create a symlink to the real rootfs, so that it is possible
+        # to access the rootfs in the same way as in the not --remote case.
+        os.symlink(remote_rootfs, os.path.join(destination, "rootfs"))
+
     def install(self, image, name):
         """
         External container install logic.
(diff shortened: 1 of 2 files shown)


peterbaouoft:
Added unit test and a change to the original logic, reasoning will be put as a review format for ease of understanding

Atomic/syscontainers.py
```diff
@@ -296,9 +296,9 @@ def _get_remote_location(self, remote_input):
         remote_path = self._resolve_remote_path(remote_input)
         if remote_path:
             remote_rootfs = os.path.sep.join([remote_path, "rootfs"])
-            if os.path.exists(remote_path):
+            if os.path.exists(remote_rootfs):
```
peterbaouoft: This is changed because when remote is `xx/rootfs` and `/xx/rootfs/usr` exist, because remote itself exist, it does not mean `xx/rootfs/rootfs`(remote rootfs) exist. 

As a result before case will fail and return a wrong remote path location: i.e (/xx/rootfs).(which I think is wrong).  It is because before the `if os.path.exists(remote_path)` will be true as `/xx/rootfs`(remote_path) exist

Atomic/syscontainers.py
```diff
@@ -296,9 +296,9 @@ def _get_remote_location(self, remote_input):
...
                 util.write_out("The remote rootfs for this container is set to be {}".format(remote_rootfs))
-            elif os.path.exists(os.path.sep.join([remote_input, "usr"])):  # Assume that the user directly gave the location of the rootfs
+            elif os.path.exists(os.path.sep.join([remote_path, "usr"])):  # Assume that the user directly gave the location of the rootfs
```
peterbaouoft: This is to handle the input with an extra `/` in it. For instance, we can have a input value as `/xx/rootfs/` as remote input, and because we have `os.path.sep.join` here, we will have `/xx/rootfs//usr` in the statement. 

If we change to remote_path(which is real path without the extra /), the above case mentioned will be fine.

Atomic/syscontainers.py
```diff
@@ -286,6 +286,24 @@ def _gather_installed_files_info (checkout, name):
 
         return info, rpm_installed, installed_files_checksum
 
+    def _get_remote_location(self, remote_input):
```
giuseppe: This function should handle two cases.  Either the user provided the path to the directory containing the `rootfs` directory (e.g. `/var/lib/atomic/containers/foo.0`) or the rootfs directory itself (e.g. `/var/lib/atomic/containers/foo.0/rootfs`.

In the previous version, `remote_rootfs` pointed to `/var/lib/atomic/containers/foo.0/rootfs` while `remote_path` pointed to `/var/lib/atomic/containers/foo.0`.

I think to minimize changes in the code, we could make this function return both values.

What do you think?

Also, the `_resolve_remote_path` function should probably be merged in this new function.

tests/unit/test_syscontainers.py
```diff
@@ -35,6 +35,38 @@ def wrapper(*args, **kwargs):
...
+    Unit tests for refactored function from SystemContainers.do_checkout method.
+    """
+    def test_get_remote_location(self):
+        """
+        We create temp directories in tmp folder, then remove it after the test completes
+
+        """
+        # Create directories for 3 different cases '_get_remote_location' is checking
+        try:
+            os.makedirs('/tmp/test-remote/rootfs/usr')
```
giuseppe: instead of hardcoding `/tmp`, please use `tempfile.mkdtemp`.  Also all the code using it must be in a try/finally block, where the finally does the `shutil.rmtree` of the temporary directory created

Atomic/syscontainers.py
```diff
@@ -926,12 +935,12 @@ def _do_checkout(self, repo, name, img, upgrade, deployment, values, destination
                     config = json.loads(config_file.read())
                 except ValueError:
                     raise ValueError("Invalid config.json file in given remote location: {}.".format(destination_path))
-                config['root']['path'] = remote_rootfs
+                config['root']['path'] = rootfs
```
giuseppe: `rootfs` will probably have the wrong value.  I suggest to return remote_rootfs from the `_get_remote_location` function and keep using it here and on line 943

tests/unit/test_syscontainers.py
peterbaouoft: ah, sounds good, thanks for the suggestion, will start wrapping entire code section under a try/finally block.

Atomic/syscontainers.py
peterbaouoft: > rootfs will probably have the wrong value

hmm, AFAICS, it seems like when remote_path is specified, rootfs will default to remote root fs anyway right? And this code section also lies under the condition of ( remote_path = true)? 

The following code seems to be the only place that does the change to the rootfs? In this code section, the rootfs will be assigned to remote rootfs location. i.e: remote/rootfs (I could be wrong tho, so feel free to correct me if I happened to understand it wrong)
https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L835-L848
```
if extract_only:
            rootfs = destination
        elif remote_path:
            rootfs = os.path.join(remote_path, "rootfs")
        else:
            destination = self._canonicalize_location(destination)
            rootfs = os.path.join(destination, "rootfs")

        if remote_path:
            if not os.path.exists(destination):
                os.makedirs(destination)
        else:
            if not os.path.exists(rootfs):
os.makedirs(rootfs)
```

Atomic/syscontainers.py
peterbaouoft:
```
This function should handle two cases. Either the user provided the path to the directory containing the rootfs directory (e.g. /var/lib/atomic/containers/foo.0) or the rootfs directory itself (e.g. /var/lib/atomic/containers/foo.0/rootfs.

In the previous version, remote_rootfs pointed to /var/lib/atomic/containers/foo.0/rootfs while remote_path pointed to /var/lib/atomic/containers/foo.0.
```
I agree, but for both cases remote path will point to `/var/lib/atomic/containers/foo.0` right?  That is why I kinda move toward the direction of making remote rootfs outside out of the function. Due to the reasoning explained in the other review comment. 

Tho that being said, I can definitely also go with the return route as well, let me know your thoughts on this, and I can change it =).

 >Also, the _resolve_remote_path function should probably be merged in this new function.

Yea, good point, I will do that in a fixup

giuseppe: I was thinking more of keeping the possibility of having the `rootfs` somewhere else, we could analyze the `config.json` for example.  This is not done now but we could decide to do it at some point.

Anyway, since it is not done now, it doesn't matter for the current patch to think of this case.  I am fine either way.

Atomic/syscontainers.py
giuseppe: sure, I am fine in the `remote_path` case to consider `rootfs` always under `remote_path` itself

peterbaouoft:
Thanks for the feedback and pushed a fixup and a new commit to address the comments ⬆️. 
- The temp directory in unit test is now generated from `tempfile.mkdtemp` and the testing code was all in a try/finally block
- The `_resolve_remote_path` is now moved into `_get_remote_location`

Atomic/syscontainers.py
```diff
@@ -926,12 +933,12 @@ def _do_checkout(self, repo, name, img, upgrade, deployment, values, destination
...
             with open(destination_path, 'w') as config_file:
                 config_file.write(json.dumps(config, indent=4))
             # create a symlink to the real rootfs, so that it is possible
             # to access the rootfs in the same way as in the not --remote case.
-            os.symlink(remote_rootfs, os.path.join(destination, "rootfs"))
+            os.symlink(rootfs, os.path.join(destination, "rootfs"))
```
giuseppe: what do you think about moving the entire `if remote_path:` block to a new function `_rewrite_rootfs(destination_path, rootfs)`?

giuseppe:
Thanks, the code looks good.

I've added an inline comment since I think the change is related to this PR.  What do you think about it?

peterbaouoft:
>what do you think about moving the entire if remote_path: block to a new function _rewrite_rootfs(destination_path, rootfs)?

Sounds good, originally I was planning to do the "sequential refactoring" (refactoring the do_checkout function sequentially into functions), But I am fine with having `_rewrite_rootfs(destination_path, rootfs)` in this patch. 

Will be working on it =).

Atomic/syscontainers.py
```diff
@@ -286,6 +286,52 @@ def _gather_installed_files_info (checkout, name):
...
+
+    @staticmethod
+    def _rewrite_rootfs(destination, remote_rootfs):
+        """
+        When remote rootfs is specified, we rewrite the ['root']['path']
+
+        :param destination: the destination of the container
+        :param remote_rootfs: remote rootfs location
+        """
+        destination_path = os.path.join(destination, "config.json")
```
peterbaouoft: Note: destination path is "generated" here instead of being a parameter passed in. This is because, in the future refactoring I plan to remove "destination_path" from local variable, to reach "separation of logic" and reduce the number of parameters/local variables. I am curious for your opinions, WDYT?  (I am ok with passing in as a parameter and save as a local variable as well)

By removing destination path, we can also save/remove two duplicated pieces, 
1: https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L380
2: https://github.com/projectatomic/atomic/blob/master/Atomic/syscontainers.py#L914

giuseppe:
@rh-atomic-bot r+ 5cf2a1e

peterbaouoft:
Thanks for the review =)

PR #1154, opened 2018-01-03
syscontainers: refactor handling rootfs into a function

peterbaouoft:
## Description
 syscontainers: refactor handling rootfs into a function

In _do_checkout, we refactor the part for getting rootfs location
into a function.

A todo comment( refactoring 'missing image layer check') is also added
in this commit based on discussions on irc

A unit test is added for future regression

## Related Issue Numbers

- [x] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -286,7 +286,8 @@ class SystemContainers(object):
 
         return info, rpm_installed, installed_files_checksum
 
-    def _get_remote_location(self, remote_input):
+    @staticmethod
+    def _get_remote_location(remote_input):
         """
         Parse the remote input and return actual remote path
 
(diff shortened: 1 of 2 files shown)


Atomic/syscontainers.py
```diff
@@ -286,7 +286,8 @@ def _gather_installed_files_info (checkout, name):
 
         return info, rpm_installed, installed_files_checksum
 
-    def _get_remote_location(self, remote_input):
+    @staticmethod
+    def _get_remote_location(remote_input):
```
peterbaouoft: I think it should be a static method as there is no `self` involved in the function, but  I normally made a decision of "static vs non-static" according to my instinct. So, is there a common way to identify when to make a function as static and when not to? (/me also searches for this online). Any suggestions are welcome =).

peterbaouoft:
>W:893, 0: TODO: The "missing layer" check logic here can be refactored once we migrate to use skopeo only (fixme)

ouch, #facepalm, let's try change to small letter case for todo

giuseppe:
@rh-atomic-bot r+ 20eb366

PR #1156, opened 2018-01-04
add support for mounting overlay objects live

### LINKED ISSUE #1153: add support for `atomic mount --live` for overlayfs backend
I understand this is just about writing the code for atomic, right? Or is there a technical reason why containers stored on overlayfs backend can't be mounted RW?

```
$ atomic mount --live something ./asd
The OverlayFS backend does not support live mounts.
```

TomasTomecek:
Fixes #1153


giuseppe:
LGTM

rhatdan:
@rhvgoyal PTAL

TomasTomecek:
ping

rhvgoyal:
This is all racy. It will keep mount point busy and lead to all kinds of issues with container removal etc. We have plenty of issues customers report with docker. This kind of functionality will just make it worse and we will not even have a way to debug it.

So who needs it and why?

rhatdan:
Vivek and I discussed this in email.  While @rhvgoyal Does not like this because it further extends a known race condition.  It keeps us on par with what we have in Devicemapper so I am going to merge, I believe this is necessary for atomic scan.

rhatdan:
@rh-atomic-bot r+

TomasTomecek:
Thanks for merging!

The reason I proposed this patch is the exact reason Dan mentioned: this functionality already exists for lvm, so why not for overlay too, especially when overlay2 is now the default for docker: https://github.com/projectatomic/container-storage-setup/commit/14012df5cf29398a31248c1e97eb163abee2abd2

We are using this functionality over here: https://github.com/fedora-modularity/conu/blob/master/conu/backend/docker/container.py#L61

PR #1163, opened 2018-01-15
ostree: try hard links before copying files to the host

giuseppe:
## Description

We had several discussions about copying client (like docker, or crioctl) to the host to make easier using system containers.  In such a perspective, take advantage of the new SELinux labelling for system containers and try to use hard links instead of copying files when possible.

This to work requires: https://github.com/containers/image/pull/401

### DIFF
--- a/Atomic/rpm_host_install.py
+++ b/Atomic/rpm_host_install.py
@@ -81,7 +81,11 @@ class RPMHostInstall(object):
         return h.hexdigest()
 
     @staticmethod
-    def rm_add_files_to_host(old_installed_files_checksum, exports, prefix="/", files_template=None, values=None, rename_files=None):
+    def _should_use_hard_link(dest_path):
+        return dest_path.startswith("/usr/")
+
+    @staticmethod
+    def rm_add_files_to_host(old_installed_files_checksum, exports, prefix="/", files_template=None, values=None, rename_files=None, use_links=True):
         # if any file was installed on the host delete it
         if old_installed_files_checksum:
             for path, checksum in old_installed_files_checksum.items():
@@ -146,7 +150,8 @@ class RPMHostInstall(object):
                         shutil.copystat(src_file, dest_path)
                         created = True
                     else:
-                        created = RPMHostInstall._copyfile(selinux_hnd, src_file, dest_path)
+                        try_hardlink = use_links and RPMHostInstall._should_use_hard_link(dest_path)
+                        created = RPMHostInstall._copyfile(selinux_hnd, src_file, dest_path, try_hardlink=try_hardlink)
 
                     if created:
                         new_installed_files_checksum[rel_dest_path] = RPMHostInstall.file_checksum(dest_path)
--- a/Atomic/rpm_host_install.py
+++ b/Atomic/rpm_host_install.py
@@ -12,7 +12,7 @@ RPM_NAME_PREFIX = "atomic-container"
 class RPMHostInstall(object):
 
     @staticmethod
-    def _copyfile(selinux_hnd, src, dest):
+    def _copyfile(selinux_hnd, src, dest, try_hardlink=False):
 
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM

giuseppe:
I've added a patch to make it configurable for each system container

giuseppe:
fine to merge?

Atomic/rpm_host_install.py
```diff
@@ -36,10 +36,24 @@ def _copyfile(selinux_hnd, src, dest):
             os.symlink(linkto, dest)
             return True
         else:
-            # we cannot use shutil.copy2() or shutil.copystat() here as it would override the
-            # security.selinux xattr.
-            shutil.copy(src, dest)
-            shutil.copymode(src, dest)
+            file_copied = False
+            # newer version of Skopeo/SELinux use the same SELinux context for f
```
peterbaouoft: Incomplete comment? maybe f refer to file here?

Atomic/rpm_host_install.py
```diff
@@ -132,7 +150,8 @@ def rm_add_files_to_host(old_installed_files_checksum, exports, prefix="/", file
                         shutil.copystat(src_file, dest_path)
```
peterbaouoft: Question: will the comment in `_copyfile` apply here as well? /me does not know very well  about copystat tho. Seems like the selinux handle is very similar to `_copyfile` part right?
```
# we cannot use shutil.copy2() or shutil.copystat() here as it would override the
# security.selinux xattr.
```

Atomic/rpm_host_install.py
```diff
@@ -36,10 +36,24 @@ def _copyfile(selinux_hnd, src, dest):
...
             return True
         else:
-            # we cannot use shutil.copy2() or shutil.copystat() here as it would override the
-            # security.selinux xattr.
-            shutil.copy(src, dest)
-            shutil.copymode(src, dest)
+            file_copied = False
+            # newer version of Skopeo/SELinux use the same SELinux context for f
+            if try_hardlink and selinux_hnd is not None:
+                src_label = selinux.getfilecon(src)
```
peterbaouoft: Below is from https://selinuxproject.org/page/LibselinuxAPISummary

getfilecon getfilecon_raw | Wrapper for the xattr API - Get file context, and set *con to refer to it. Caller must free via freecon(3).
-- | --

Do we need to worry about freeing context here then?

Atomic/syscontainers.py
```diff
@@ -1297,6 +1301,7 @@ def rollback(self, name):
         installed_files_template = info["installed-files-template"] if "installed-files-template" in info and rpm_installed is None else None
         has_container_service = info["has-container-service"] if "has-container-service" in info else True
         rename_files = info["rename-installed-files"] if "rename-installed-files" in info else None
+        use_links = info["use-links"] if "use-links" in info else False
```
peterbaouoft: > Specify if files copied to the host under */usr* should use
+hard links when possible.  By default it is True.

Shouldn't we default to True here?

Atomic/rpm_host_install.py
rhatdan: The selinux bindings should be handling that, so no.

Atomic/rpm_host_install.py
giuseppe: great catch.  I pushed a fixup patch to change it to `copymode`

Atomic/rpm_host_install.py
giuseppe: fixed in a fixup patch

giuseppe:
@rh-atomic-bot delegate= peterbaouoft

peterbaouoft:
LGTM =) , and since @rhatdan also approves it earlier, @rh-atomic-bot r+ f9b1c08

PR #1160, opened 2018-01-09
Refactor syscontainer config dup

peterbaouoft:
## Description
This pull request removes the unused function introduced earlier and contains refactoring for write config in '_do_checkout' function,
this refactoring also removes the duplication of writing config in 'run_once'

There should not be functional impact with this commit.

A unit test is added for future regression

## Related Issue Numbers
- https://github.com/projectatomic/atomic/issues/1149

If your Pull request contains new features or functions, tests are required. If the PR is a bug fix and no tests exist, please consider adding some to prevent regressions.
- [x] Unittests
- [ ] Integration Tests

PR #1164, opened 2018-01-15
syscontainer: refactor parameter list

peterbaouoft:
## Description
According to discussion, the '_checkout' function is now being replaced
temporarily with '_checkout_wrapper'. Now, the parameters other than
repo is being packed into a dictionary and the parameter list for
'_do_checkout' and '_checkout' is greatly shortened

This commit also contains a small  "if fix", by refactoring the same
if statements into one.

This commit enables further refactoring on '_do_checkout' and should
not introduce any functional change. With the help with other future
refactoring that builds on top of this commit, '_do_checkout' function
should look nicer in the future =).

**Note**: there are many improvements can be built on top of this commit, but I felt it might be nice to have this as a separate working commit on its own. 

## Related Issue Numbers
-https://github.com/projectatomic/atomic/issues/1149

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -249,7 +249,7 @@ class SystemContainers(object):
         rootfs = os.path.join(rpm_content, "usr/lib/containers/atomic", name)
         os.makedirs(rootfs)
         try:
-            self._checkout(repo, name, image, 0, SystemContainers.CHECKOUT_MODE_INSTALL, values=values, destination=rootfs, prefix=rpm_content)
+            self._checkout_wrapper(repo, name, image, 0, SystemContainers.CHECKOUT_MODE_INSTALL, values=values, destination=rootfs, prefix=rpm_content)
             if self.display:
                 return None
             img = self.inspect_system_image(image)
@@ -532,7 +532,7 @@ class SystemContainers(object):
                 util.write_out("Generated rpm %s" % destination)
             return False
 
-        self._checkout(repo, name, image, 0, SystemContainers.CHECKOUT_MODE_INSTALL, values=values, remote=self.args.remote, system_package=self.args.system_package)
+        self._checkout_wrapper(repo, name, image, 0, SystemContainers.CHECKOUT_MODE_INSTALL, values=values, remote=self.args.remote, system_package=self.args.system_package)
 
     def _check_oci_configuration_file(self, conf_path, remote=None, include_all=False):
         with open(conf_path, 'r') as conf:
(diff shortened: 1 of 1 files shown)


Atomic/syscontainers.py
```diff
@@ -634,34 +634,54 @@ def _resolve_remote_path(self, remote_path):
...
-                pass
-            try:
-                if not extract_only and mode == SystemContainers.CHECKOUT_MODE_INSTALL:
+                except OSError:
+                    pass
+                try:
                     os.unlink(tmpfilesout)
-            except OSError:
-                pass
+                except OSError:
```
giuseppe: tabs here

Atomic/syscontainers.py
```diff
@@ -634,34 +634,54 @@ def _resolve_remote_path(self, remote_path):
...
+        options = {"name" : name,
+                   "img" : img,
+                   "deployment" : deployment,
+                   "UpgradeMode" : mode,
+                   "values" : values,
+                   "destination" : destination,
+                   "extract_only" : extract_only,
+                   "remote" : remote,
+                   "prefix" : prefix,
+                   "InstalledFilesChecksum" : installed_files_checksum,
```
giuseppe: some options names use camel case, other ones use underscores.  Could we make them all in the underscore notation (i.e. replace UpgradeMode and InstalledFilesChecksum)?

Atomic/syscontainers.py
peterbaouoft: Thanks for reviewing =). 
> tabs here

hmm, can you explain a bit more on this? Are you referring to the "unnecessarily added tabs" here? Do note I have changed the expression here a bit, "I refactored the three duplicated if statements outside of the try except blocks", so each `try-except` block is now under one `if statement` umbrella.

Atomic/syscontainers.py
peterbaouoft: sounds good, sure, will do that

Atomic/syscontainers.py
giuseppe: github showed there is a tab to indent the block.  If I misread it, please just ignore this comment :-)

giuseppe:
LGTM!

@rh-atomic-bot r+ 2999e28

Atomic/syscontainers.py
```diff
@@ -963,29 +982,29 @@ def _do_checkout(self, repo, name, img, mode, deployment, values, destination, u
         else:
             systemd_template = SYSTEMD_UNIT_FILE_DEFAULT_TEMPLATE
 
-        values = self._amend_values(values, manifest, name, img, image_id, destination, prefix, unit_file_support_pidfile=SystemContainers._template_support_pidfile(systemd_template))
+        values = self._amend_values(options["values"], manifest, options["name"], options["img"], image_id, options["destination"], options["prefix"], unit_file_support_pidfile=SystemContainers._template_support_pidfile(systemd_template))
```
peterbaouoft: Ouch, just noticed in when I was doing other rebasing, it is interesting that none of our tests actually caught it. Sorry it was my fault for not noticing it.  Will do the fix with options["values"] in pull request: https://github.com/projectatomic/atomic/pull/1160

Atomic/syscontainers.py
```diff
@@ -963,29 +982,29 @@ def _do_checkout(self, repo, name, img, mode, deployment, values, destination, u
...
 
         src = os.path.join(exports, "config.json")
-        destination_path = os.path.join(destination, "config.json")
+        destination_path = os.path.join(options["destination"], "config.json")
         if os.path.exists(src):
             shutil.copyfile(src, destination_path)
         elif os.path.exists(src + ".template"):
             with open(src + ".template", 'r') as infile:
                 util.write_template(src + ".template", infile.read(), values, destination_path)
```
peterbaouoft: And ofc... there is another duplicated values here, that makes the compiler not show warnings #facepalms

Atomic/syscontainers.py
peterbaouoft: I think we should add some test to verify the expected values for "values" or something like that in the future(either via integration or unit test)  will update this in the issue https://github.com/projectatomic/atomic/issues/1149.

And will assign this to myself once "major refactoring" (I think) is done

PR #1166, opened 2018-01-25
syscontainers: allow the container name to end in .[01]

giuseppe:
we skipped file names just based on their ending without checking if
it is a symlink.  This prevented container with a name ending in .0 or
.1 to be uninstalled.

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1389,7 +1389,12 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             if not os.path.exists(fullpath):
                 continue
             if fullpath.endswith(".0") or fullpath.endswith(".1"):
-                continue
+                # Skip the deployments if we have a symlink named in
+                # the same name except the ".[01]" ending.
+                # We avoid duplicates as we are already including the container
+                # in the results from the symlink and not from the deployments.
+                if os.path.islink(fullpath[:-2]):
+                    continue
 
             with open(os.path.join(fullpath, "info"), "r") as info_file:
                 info = json.load(info_file)
--- a/tests/integration/test_system_containers_install.sh
+++ b/tests/integration/test_system_containers_install.sh
@@ -13,6 +13,7 @@ IFS=$'\n\t'
 # 6. install from a dockertar
 # 7. install from local docker
 # 8. install a run-once system container
+# 9. install a container with .0 in the name
 
 setup () {
     docker save atomic-test-system > ${WORK_DIR}/atomic-test-system.tar
@@ -108,7 +109,7 @@ assert_matches ${SECRET}-new ${ATOMIC_OSTREE_CHECKOUT_PATH}/${NAME}.0/config.jso
 teardown
 
 
-# 5. installing from a dockertar
+# 6. installing from a dockertar
 export NAME="test-dockertar-system-container-$$"
 ${ATOMIC} install --name=${NAME} --set=RECEIVER=${SECRET} --system dockertar:/${WORK_DIR}/atomic-test-system.tar
 test -e /etc/tmpfiles.d/${NAME}.conf
@@ -119,7 +120,7 @@ test -e ${ATOMIC_OSTREE_CHECKOUT_PATH}/${NAME}.0/tmpfiles-${NAME}.conf
 teardown
 
 
-# 6. install from local docker
+# 7. install from local docker
 export NAME="test-docker-system-container-$$"
 ${ATOMIC} install --name=${NAME} --set=RECEIVER=${SECRET} --system docker:atomic-test-system:latest
 test -e /etc/tmpfiles.d/${NAME}.conf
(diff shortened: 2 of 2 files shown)


peterbaouoft:
.. the test output/bot seems exploded, I am suspecting it might be a flake. But to confirm, do you mind running the tester again, so I can see if the error remains and debug for it? =P

peterbaouoft:
bot, retest this please

peterbaouoft:
```
Unable to find atomic-test-system:latest
  File "atomic", line 185, in <module>
    sys.exit(_func())
  File "/var/tmp/checkout/Atomic/install.py", line 118, in install
    remote_image_obj = be.make_remote_image(self.args.image)
  File "/var/tmp/checkout/Atomic/backends/_ostree.py", line 156, in make_remote_image
    img_obj = self._make_remote_image(image)
  File "/var/tmp/checkout/Atomic/backends/_ostree.py", line 161, in _make_remote_image
    return self._make_image(image, None, remote=True)
  File "/var/tmp/checkout/Atomic/backends/_ostree.py", line 50, in _make_image
    img_obj = Image(image, backend=self, remote=remote)
  File "/var/tmp/checkout/Atomic/objects/image.py", line 54, in __init__
    self._instantiate()
  File "/var/tmp/checkout/Atomic/objects/image.py", line 81, in _instantiate
    self._setup_common()
  File "/var/tmp/checkout/Atomic/objects/image.py", line 89, in _setup_common
    self.registry, self.repo, self.image, self.tag, self.digest = Decompose(self.fq_name).all
  File "/var/tmp/checkout/Atomic/objects/image.py", line 127, in fq_name
    self._fq_name = ri.find_image_on_registry(quiet=True)
  File "/var/tmp/checkout/Atomic/discovery.py", line 96, in find_image_on_registry
    raise RegistryInspectError("Unable to find {}".format(self.orig_input))
Atomic.discovery.RegistryInspectError: Unable to find atomic-test-system:latest
```
hmm, looks like registry is down?

Atomic/syscontainers.py
```diff
@@ -1388,7 +1388,7 @@ def _get_containers_at(self, checkouts, are_preinstalled, containers=None):
             fullpath = os.path.join(checkouts, x)
             if not os.path.exists(fullpath):
                 continue
-            if fullpath.endswith(".0") or fullpath.endswith(".1"):
+            if not os.path.islink(fullpath) and (fullpath.endswith(".0") or fullpath.endswith(".1")):
```
peterbaouoft: Hmm, this might not be too relevant, but I am confused about the expression here. It would be great if you can clarify it a bit =).  I thought the actual folder for the container itself usually ends with .0 and .1 right? (and .0 or .1 is the deployment version).

Then If I understand this correctly, this check basically says, 1: if the path is not a symlink and 2: if the path is a container folder, we will skip the path. (but why do we want information only from a symlink path? shouldn't we care more about the actual existent path?)

The reason why I think folder name ends with .0 or .1 will be "actual container folder" (not symlink) is the following two lines: 
```
options["destination"] = options["destination"] or os.path.join(self._get_system_checkout_path(), "{}.{}".format(options["name"], options["deployment"]))
```
and 
```
        sym = os.path.join(self._get_system_checkout_path(), options["name"])
        if os.path.exists(sym):
            os.unlink(sym)
        os.symlink(options["destination"], sym)
```

which suggests the symlink is the one without the version number? Thanks in advance.

giuseppe: we use only the name in the symlink as it doesn't have the additional ".0" and ".1" that are in the checkout directories.

I've modified the patch a bit and pushed a new version that includes also an inline comment.

peterbaouoft: ah I think it starts to make sense now. Thanks for clarifying this. So to confirm I understand this, the reason why we only care about symlink is because it represents the current container, and it also eliminates the concern about the container with older versions (i.e the folder with both container versions in it. For instance, firewalld.0 and firewalld.1 can both appear in a same container directory, and we only want the current version of container, which is linked by the symlink) right?

peterbaouoft:
LGTM, and tested this patch 

Before:
```
[root@localhost atomic]# atomic install --system --name=container.0 --system-package=no hello-world                   
Extracting to /var/lib/containers/atomic/container.0.0
Warning: /exports directory not found.  Default config files will be generated.
Warning: You may want to modify `/var/lib/containers/atomic/container.0.0/config.json` before starting the service
systemctl daemon-reload
systemctl enable container.0
[root@localhost atomic]# atomic containers list -a
   CONTAINER ID IMAGE                NAME       COMMAND    CREATED          STATE      BACKEND    RUNTIME   
   51dbd0649b9a atomic-test-6        thirsty_pi sleep 100  2018-01-25 22:53 exited     docker     docker    
   3b0ad1936c68 atomic-test-6        agitated_l sleep 100  2018-01-18 18:47 exited     docker     docker    
   d1ef850fae38 atomic-test-6        quizzical_ sleep 100  2018-01-15 22:05 exited     docker     docker    
   733d4dda2c32 atomic-test-6        dreamy_tho sleep 100  2017-12-21 20:26 exited     docker     docker    
   eaef7d920261 atomic-test-6        wonderful_ sleep 100  2017-12-13 19:56 exited     docker     docker    
   f9d9fb28632e atomic-test-6        zealous_fe sleep 100  2017-12-12 22:52 exited     docker     docker    
   firewalld    firewalld:latest     firewalld  /usr/bin/r 2017-12-12 22:25 running    ostree     /bin/runc 
[root@localhost atomic]# atomic uninstall container.0
[root@localhost atomic]# atomic uninstall container.0
Unable to find 'container.0' in the following backends: docker, ostree
```

Applied this patch:
```
[root@localhost atomic]# atomic containers list -a
   CONTAINER ID IMAGE                NAME       COMMAND    CREATED          STATE      BACKEND    RUNTIME   
   51dbd0649b9a atomic-test-6        thirsty_pi sleep 100  2018-01-25 22:53 exited     docker     docker    
   3b0ad1936c68 atomic-test-6        agitated_l sleep 100  2018-01-18 18:47 exited     docker     docker    
   d1ef850fae38 atomic-test-6        quizzical_ sleep 100  2018-01-15 22:05 exited     docker     docker    
   733d4dda2c32 atomic-test-6        dreamy_tho sleep 100  2017-12-21 20:26 exited     docker     docker    
   eaef7d920261 atomic-test-6        wonderful_ sleep 100  2017-12-13 19:56 exited     docker     docker    
   f9d9fb28632e atomic-test-6        zealous_fe sleep 100  2017-12-12 22:52 exited     docker     docker    
   container.0  hello-world:latest   container. run.sh     2018-01-31 22:22 inactive   ostree     /bin/runc 
   firewalld    firewalld:latest     firewalld  /usr/bin/r 2017-12-12 22:25 running    ostree     /bin/runc 
[root@localhost atomic]# atomic uninstall container.0
systemctl disable --now container.0
systemctl daemon-reload
```

giuseppe:
@rh-atomic-bot r=peterbaouoft

PR #1170, opened 2018-02-01
syscontainers: read labels from v1Compatibility

giuseppe:
Images coming from registry.access.redhat.com through "skopeo copy"
have this information in the v1Compatibility block.  Use this
information so we can correctly catch "atomic.run".

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1646,7 +1646,13 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             virtual_size = self._get_virtual_size(repo, manifest)
             if 'Labels' in manifest:
                 labels = manifest['Labels']
-            else:
+            elif 'history' in manifest:
+                for i in manifest['history']:
+                    if 'v1Compatibility' in i:
+                        config = json.loads(i['v1Compatibility'])
+                        if 'container_config' in config and 'Labels' in config['container_config']:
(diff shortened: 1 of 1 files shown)


Atomic/syscontainers.py
```diff
@@ -1646,7 +1646,13 @@ def _inspect_system_branch(self, repo, imagebranch):
             virtual_size = self._get_virtual_size(repo, manifest)
             if 'Labels' in manifest:
                 labels = manifest['Labels']
-            else:
+            elif 'history' in manifest:
```
TomSweeneyRedHat: Might be time to educate Tom, but should this be an if rather than an elif?

giuseppe: no, I'd like to take this branch only if the first 'if' was not satisfied.

If the labels are found in `manifest['Labels']` we don't need to look into 'history'.

TomSweeneyRedHat: @giuseppe Thanks for the info.  FYI some of the tests aren't looking happy.

giuseppe: it is because of storage :( it is failing since long time but it should not be related to this PR

baude:
@rh-atomic-bot r+

giuseppe:
@rh-atomic-bot retry

baude:
bot, retest this please

giuseppe:
@rh-atomic-bot r=baude 12edeeb

Issue #1177, opened 2018-02-05
Remove ostree fallback in syscontainers.py


peterbaouoft:
>At this point I think it is safe to assume every Skopeo released has
support for "ostree", so we can drop the fallback version in syscontainers.py.  
We will need to check though that it works in the CI
for the CentOS build

Open an issue here to track the progress of the above task and it can also be a place for possible discussion. (including implementation details, migration details, design details, and even regression details if possible)

Moreover, this can be a place for me to ask questions =P.

peterbaouoft:
**FTR**: After discussion with Giuseppe, we think it is better to remove "docker->ostree" fallback as well. That means, when user do `atomic pull --storage ostree docker:firewalld:latest` or `atomic pull --storage ostree dockertar:/path/to/tar`, only skopeo will be used from now on.  Further information will be updated.

peterbaouoft:
PR merged, closing

PR #1180, opened 2018-02-06
syscontainer: remove of fallback version of skopeo

peterbaouoft:
Based on discussion, skopeo is assumed to include support for
ostree copy from now on.

The users will now be noted to install the newest version
if their skopeo version does not support copy to ostree.

## Related Issue Numbers
- https://github.com/projectatomic/atomic/issues/1177
-

## Pull Request Checklist:

- [ ] Unittests
- [x] Integration Tests

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -2062,10 +2062,6 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         args, img = self._convert_to_skopeo(image)
         return util.skopeo_inspect(img, args)
 
-    def _skopeo_get_layers(self, image, layers):
-        _, img = self._convert_to_skopeo(image)
-        return util.skopeo_layers(img, [], layers)
-
     def _image_manifest(self, repo, rev):
         return SystemContainers._get_commit_metadata(repo, rev, "docker.manifest")
 
@@ -2274,7 +2270,8 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
 
         if can_use_skopeo_copy:
             return self._check_system_oci_image_skopeo_copy(repo, img, src_creds=src_creds)
-        return self._check_system_oci_image_no_skopeo_copy(repo, img, imagebranch)
+        else:
+            raise ValueError("Skopeo version too old, please install the newest version")
 
     def _check_system_oci_image_skopeo_copy(self, repo, img, src_creds=None):
         repo = self.get_ostree_repo_location()
@@ -2291,37 +2288,6 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             util.skopeo_copy("docker://" + img, destination, dest_ostree_tmp_dir=temp_dir, insecure=insecure, src_creds=src_creds)
         finally:
             shutil.rmtree(temp_dir, ignore_errors=True)
-        return True
-
-    def _check_system_oci_image_no_skopeo_copy(self, repo, img, imagebranch):
-        try:
-            manifest = self._skopeo_get_manifest(img)
-        except ValueError:
-            raise ValueError("Unable to find {}".format(img))
-        layers = SystemContainers.get_layers_from_manifest(manifest)
-        missing_layers = []
-        for i in layers:
-            layer = i.replace("sha256:", "")
-            has_layer = repo.resolve_rev("%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer), True)[1]
-            if not has_layer:
-                missing_layers.append(layer)
-                util.write_out("Pulling layer %s" % layer)
-        layers_dir = None
-        try:
-            layers_to_import = {}
-            if len(missing_layers):
-                layers_dir = self._skopeo_get_layers(img, missing_layers)
-                for root, _, files in os.walk(layers_dir):
-                    for f in files:
-                        if f.endswith(".tar"):
-                            layer_file = os.path.join(root, f)
-                            layer = f.replace(".tar", "")
-                            if not repo.resolve_rev("%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer), True)[1]:
-                                layers_to_import[layer] = layer_file
-            self._import_layers_into_ostree(repo, imagebranch, manifest, layers_to_import)
-        finally:
-            if layers_dir:
-                shutil.rmtree(layers_dir)
         return True
 
     @staticmethod
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -2165,13 +2165,18 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
 
     def _skopeo_copy_img_to_ostree(self, img, skopeo_img_source, src_creds=None, insecure=None):
         """
-        Pefrom skopeo copy operation to copy images from img_source to ostree location based on image
+        Perform skopeo copy operation to copy images from img_source to ostree location based on image
         name
         :param img: image name
+        :type img: str
         :param skopeo_img_source: the source for skopeo copying
-        :src_creds: source credientials if pulled from online registries
-        :insecure: tell skopeo whether the registry is secure or not
+        :type skopeo_img_source: str
+        :param src_creds: source credentials if pulled from online registries
+        :type src_creds: str in the form of USERNAME[:PASSWORD]
+        :param insecure: tell skopeo whether the registry is secure or not
+        :type insecure: bool
         :returns: True if successfully performed copy operation, a ValueError can be raised upon failures
+        :rtype: bool
         """
         repo = self.get_ostree_repo_location()
 
@@ -2191,7 +2196,9 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         """
         Retrieve the image name from the dockertar file
         :param tarpath: the path of the docker tar file
+        :type tarpath: str
         :returns: the true image name got from the dockertar
+        :rtype: str
         """
         temp_dir = tempfile.mkdtemp()
         try:
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -179,6 +179,12 @@ class SystemContainers(object):
     def _pull_image_to_ostree(self, repo, image, upgrade, src_creds=None):
         if not repo:
             raise ValueError("Cannot find a configured OSTree repo")
+
+        # Since we are migrating to use skopeo only, we will move the check for ostree upper level too
+        can_use_skopeo_copy = util.check_output([util.SKOPEO_PATH, "copy", "--help"]).decode().find("ostree") >= 0
+        if not can_use_skopeo_copy:
+            raise ValueError("Skopeo version too old, Please use skopeo version after 1.21")
+
         if image.startswith("docker:") and image.count(':') > 1:
             image = self._pull_docker_image(image.replace("docker:", "", 1))
         elif image.startswith("dockertar:/"):
@@ -2125,9 +2131,19 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         return layers
 
     def _pull_docker_image(self, image):
-        with tempfile.NamedTemporaryFile(mode="w") as temptar:
-            util.check_call(["docker", "save", "-o", temptar.name, image])
-            return self._pull_docker_tar(temptar.name, image)
+        """
+        Use skopeo to copy docker image from docker daemon to ostree
+
+        :param image: The full name of image to pull. E.g: docker:image:latest
+        :type image: str
+        :returns image name
+        :rtype: str
+        """
+        skopeo_source = "docker-daemon:" + image
+        self._skopeo_copy_img_to_ostree(image, skopeo_source)
+
+        # Return back the image name for future usages
+        return image
 
     def _pull_docker_tar(self, tarpath, image):
         """
@@ -2140,10 +2156,6 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         :rtype: str
         """
         image_name = SystemContainers._get_image_name_from_dockertar(tarpath) or image
-        can_use_skopeo_copy = util.check_output([util.SKOPEO_PATH, "copy", "--help"]).decode().find("ostree") >= 0
-
-        if not can_use_skopeo_copy:
-            raise ValueError("Skopeo version too old, Please use version after 1.21")
 
         skopeo_source = "docker-archive:" + tarpath
         self._skopeo_copy_img_to_ostree(image_name, skopeo_source)
@@ -2229,12 +2241,7 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             # and never let it fail.
             pass
 
-        can_use_skopeo_copy = util.check_output([util.SKOPEO_PATH, "copy", "--help"]).decode().find("ostree") >= 0
-
-        if can_use_skopeo_copy:
-            return self._check_system_oci_image_skopeo_copy(img, src_creds=src_creds)
-        else:
-            raise ValueError("Skopeo version too old, please install the newest version")
+        return self._check_system_oci_image_skopeo_copy(img, src_creds=src_creds)
 
     def _check_system_oci_image_skopeo_copy(self, img, src_creds=None):
         # Pass the original name to "skopeo copy" so we don't resolve it in atomic
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -181,10 +181,10 @@ class SystemContainers(object):
         if not repo:
             raise ValueError("Cannot find a configured OSTree repo")
         if image.startswith("docker:") and image.count(':') > 1:
-            image = self._pull_docker_image(repo, image.replace("docker:", "", 1))
+            image = self._pull_docker_image(image.replace("docker:", "", 1))
         elif image.startswith("dockertar:/"):
             tarpath = image.replace("dockertar:/", "", 1)
-            image = self._pull_docker_tar(repo, tarpath, os.path.basename(tarpath).replace(".tar", ""))
+            image = self._pull_docker_tar(tarpath, os.path.basename(tarpath).replace(".tar", ""))
         else: # Assume "oci:"
             self._check_system_oci_image(repo, image, upgrade, src_creds=src_creds)
         return image
@@ -2197,12 +2197,63 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
 
         repo.commit_transaction(None)
 
-    def _pull_docker_image(self, repo, image):
+    def _pull_docker_image(self, image):
         with tempfile.NamedTemporaryFile(mode="w") as temptar:
             util.check_call(["docker", "save", "-o", temptar.name, image])
-            return self._pull_docker_tar(repo, temptar.name, image)
+            return self._pull_docker_tar(temptar.name, image)
 
-    def _pull_docker_tar(self, repo, tarpath, image):
+    def _pull_docker_tar(self, tarpath, image):
+        """
+        Resolves a docker tar and pull its content into ostree using skopeo
+        :param tarpath: the path for the docker tar file
+        :type tarpath: str
+        :param image: the default image name extracted from tar file name
+        :type image: str
+        :returns: the resolved image name that will be pulled into ostree
+        :rtype: str
+        """
+        image_name = SystemContainers._get_image_name_from_dockertar(tarpath) or image
+        can_use_skopeo_copy = util.check_output([util.SKOPEO_PATH, "copy", "--help"]).decode().find("ostree") >= 0
+
+        if not can_use_skopeo_copy:
+            raise ValueError("Skopeo version too old, Please use version after 1.21")
+
+        skopeo_source = "docker-archive:" + tarpath
+        self._skopeo_copy_img_to_ostree(image_name, skopeo_source)
+
+        # Return the image name for future usages
+        return image_name
+
+    def _skopeo_copy_img_to_ostree(self, img, skopeo_img_source, src_creds=None, insecure=None):
+        """
+        Pefrom skopeo copy operation to copy images from img_source to ostree location based on image
+        name
+        :param img: image name
+        :param skopeo_img_source: the source for skopeo copying
+        :src_creds: source credientials if pulled from online registries
+        :insecure: tell skopeo whether the registry is secure or not
+        :returns: True if successfully performed copy operation, a ValueError can be raised upon failures
+        """
+        repo = self.get_ostree_repo_location()
+
+        checkout = self._get_system_checkout_path()
+        destdir = checkout if os.path.exists(checkout) else None
+        temp_dir = tempfile.mkdtemp(prefix=".", dir=destdir)
+
+        destination = "ostree:{}@{}".format(img, repo)
+        try:
+            util.skopeo_copy(skopeo_img_source, destination, dest_ostree_tmp_dir=temp_dir, insecure=insecure, src_creds=src_creds)
+        finally:
+            shutil.rmtree(temp_dir, ignore_errors=True)
+        return True
+
+    @staticmethod
+    def _get_image_name_from_dockertar(tarpath):
+        """
+        Retrieve the image name from the dockertar file
+        :param tarpath: the path of the docker tar file
+        :returns: the true image name got from the dockertar
+        """
         temp_dir = tempfile.mkdtemp()
         try:
             with tarfile.open(tarpath, 'r') as t:
@@ -2213,28 +2264,13 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
                     with open(manifest_file, 'r') as mfile:
                         manifest = mfile.read()
                     for m in json.loads(manifest):
-                        if "Config" in m:
-                            config_file = os.path.join(temp_dir, m["Config"])
-                            with open(config_file, 'r') as config:
-                                config = json.loads(config.read())
-                                labels = config['config']['Labels']
-                        imagename = m["RepoTags"][0] if m.get("RepoTags") else image
-                        imagebranch = "%s%s" % (OSTREE_OCIIMAGE_PREFIX, SystemContainers._encode_to_ostree_ref(imagename))
-                        input_layers = m["Layers"]
-                        self._pull_dockertar_layers(repo, imagebranch, temp_dir, input_layers, labels=labels)
+                        imagename = m["RepoTags"][0] if m.get("RepoTags") else None
                 else:
                     repositories = ""
                     repositories_file = os.path.join(temp_dir, "repositories")
                     with open(repositories_file, 'r') as rfile:
                         repositories = rfile.read()
                     imagename = list(json.loads(repositories).keys())[0]
-                    imagebranch = "%s%s" % (OSTREE_OCIIMAGE_PREFIX, SystemContainers._encode_to_ostree_ref(imagename))
-                    input_layers = []
-                    for name in os.listdir(temp_dir):
-                        if name == "repositories":
-                            continue
-                        input_layers.append(name + "/layer.tar")
-                    self._pull_dockertar_layers(repo, imagebranch, temp_dir, input_layers)
             return imagename
         finally:
             shutil.rmtree(temp_dir)
@@ -2269,26 +2305,15 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         can_use_skopeo_copy = util.check_output([util.SKOPEO_PATH, "copy", "--help"]).decode().find("ostree") >= 0
 
         if can_use_skopeo_copy:
-            return self._check_system_oci_image_skopeo_copy(repo, img, src_creds=src_creds)
+            return self._check_system_oci_image_skopeo_copy(img, src_creds=src_creds)
         else:
             raise ValueError("Skopeo version too old, please install the newest version")
 
-    def _check_system_oci_image_skopeo_copy(self, repo, img, src_creds=None):
-        repo = self.get_ostree_repo_location()
-
-        checkout = self._get_system_checkout_path()
-        destdir = checkout if os.path.exists(checkout) else None
-        temp_dir = tempfile.mkdtemp(prefix=".", dir=destdir)
-
+    def _check_system_oci_image_skopeo_copy(self, img, src_creds=None):
         # Pass the original name to "skopeo copy" so we don't resolve it in atomic
         insecure, img = self._get_skopeo_args(img, full_resolution=False)
-
-        destination = "ostree:{}@{}".format(img, repo)
-        try:
-            util.skopeo_copy("docker://" + img, destination, dest_ostree_tmp_dir=temp_dir, insecure=insecure, src_creds=src_creds)
-        finally:
-            shutil.rmtree(temp_dir, ignore_errors=True)
-        return True
+        skopeo_img_src = "docker://" + img
+        return self._skopeo_copy_img_to_ostree(img, skopeo_img_src, src_creds=src_creds, insecure=insecure)
 
     @staticmethod
     def _generate_tmpfiles_data(missing_bind_paths):
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -7,7 +7,6 @@ import tarfile
 from string import Template
 import calendar
 import shutil
-import stat # pylint: disable=bad-python3-import
 import subprocess
 import time
 import errno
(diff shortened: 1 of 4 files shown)


Atomic/syscontainers.py
```diff
@@ -2239,7 +2235,8 @@ def _check_system_oci_image(self, repo, img, upgrade, src_creds=None):
 
         if can_use_skopeo_copy:
             return self._check_system_oci_image_skopeo_copy(repo, img, src_creds=src_creds)
-        return self._check_system_oci_image_no_skopeo_copy(repo, img, imagebranch)
+        else:
+            raise ValueError("Skopeo version too old, please install the newest version")
```
peterbaouoft: **Question**: Do we want to have a hard dependency on skopeo version here? We can also use Makefile to enforce the version match (>= a specific version of skopeo).  

We can also choose to give user a warning instead of raising errors, i am open to suggestions


Another doubt here is that, how do we retrieve the version of the "working" skopeo, so users know which version at least they need to install? The dilemma here is to decide between whether we should hard code version here, or we retrieve it dynamically (dunno how to do it tho)

peterbaouoft:
Looks like centos ci still have an old skopeo version.(The error message is what I just added to the PR). Is there a way to fix this atm? If not, let's wait a bit on this (until centos ci have the newer skopeo)  :-)? 
```
+ atomic --debug pull --storage ostree busybox
The image `busybox` is not fully qualified.  The default registry configured for Skopeo will be used.
Failed: Skopeo version too old, please install the newest version
  File "/usr/bin/atomic", line 185, in <module>
    sys.exit(_func())
  File "/usr/lib/python2.7/site-packages/Atomic/pull.py", line 64, in pull_image
    raise ValueError("Failed: {}".format(e))
ValueError: Failed: Skopeo version too old, please install the newest version
Namespace(_class=<class 'Atomic.pull.Pull'>, assumeyes=False, debug=True, func='pull_image', ignore=False, image='busybox', profile=False, reg_type=None, src_creds=None, storage='ostree')
+ teardown
+ set +o pipefail
+ atomic --debug -y containers delete busybox
+ true
+ ostree --repo=/var/tmp/checkout/.tmp.x2KxQiRV1l/repo refs --delete ociimage
```

ashcrow:
Is there an upstream ticket for updating ``skopeo``?

peterbaouoft:
> Is there an upstream ticket for updating skopeo

Hi, no not yet, I had a question about it on irc, guess you might missed it. going to repaste here :-). 
> <rubao|laptop> ashcrow, hmm for upstream issue, where should I file against? cuz, skopeo itself already support the ostree copy feature, but centos ci itself did not "upgrade" to the newest skopeo version yet

ashcrow:
I believe https://bugs.centos.org/my_view_page.php is the location to add issues for centos ci.

peterbaouoft:
> I believe https://bugs.centos.org/my_view_page.php is the location to add issues for centos ci. 

thanks for letting me know, will register account and file an issue today. :-)

peterbaouoft:
bot, retest this please

ashcrow:
bot, retest this please

peterbaouoft:
From centos ci, looks like skopeo version is up to date(contain ostree copy), will therefore mark this unblocked, and mark this to be WIP :-). Thanks for the help!
```
Running test_containers_list.sh....             PASS
Running test_diff.sh....                        PASS
Running test_display.sh....                     PASS
Running test_dockertar_sha256_helper.sh....     PASS
Running test_help.sh....                        PASS
Running test_images_list.sh....                 PASS
Running test_info.sh....                        PASS
Running test_migrate.sh....                     PASS
Running test_mount.sh....                       PASS
Running test_params.sh....                      PASS
Running test_pass.sh....                        PASS
Running test_printenv.sh....                    PASS
Running test_run.sh....                         PASS
Running test_run_opts.sh....                    PASS
Running test_storage.sh....                     FAIL
Running test_system_containers_images.sh....    PASS
Running test_system_containers_install.sh....   PASS
Running test_system_containers_mount.sh....     PASS
Running test_system_containers_rpm.sh....       SKIP
Running test_system_containers_runtime.sh....   PASS
Running test_tag.sh....                         PASS
Running test_top.sh....                         PASS
Running test_verify.sh....                      PASS
```

peterbaouoft:
Let me know your thoughts about the second commit. I did a bit of reordering of logic there( verify first, then checkout layers).  And, for code that checks for missing layer, but do not error out, I am not sure of a neat way to refactor those yet. Any suggestions will be appreciated :). 

For the  check missing layer but "do not error out" case that I have mentioned: an example is like the following: in `_check_system_oci_image` function
```
 for layer in layers:
                rev_layer = repo.resolve_rev("%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer.replace("sha256:", "")), True)[1]
                if not rev_layer:
                    missing_layers = True
                    break
            if not missing_layers:
                return False
```

Atomic/syscontainers.py
ashcrow: I don't think we can use Makefile to ensure the version of skopeo as where a build happens and where a tool is installed will likely be different hosts. We can add a dependency check in the RPM spec file. If we want to make a hard need on version a util function that checks the skopeo version on first usage could be made but that's a different PR.

ashcrow:
In follow the logic switch you did @peterbaouoft. There is a corner case drawback I can think of which is if the underlying data is changed between the verification and the checkout the results will either be incorrect or, more likely, an error will be raised. It's probably safer to check for error while checking out, but it wouldn't hurt to do both if you want to give a nice error upfront.

@giuseppe WDYT?

Atomic/syscontainers.py
peterbaouoft: Thanks for the comment :). 

>If we want to make a hard need on version a util function that checks the skopeo version on first usage could be made but that's a different PR.

I think this approach sounds nicer. I also don't think it would be too much work for doing this check as well. Will write up a PR to check for skopeo version after this PR.

peterbaouoft:
> here is a corner case drawback I can think of which is if the underlying data is changed between the verification and the checkout the results will either be incorrect or, more likely, an error will be raised

Ah, yes, thanks for pointing it out! Yea, I think it might be worth it to do both checking(early checking and in between checking). Will rework on the patch a bit to cover both cases you mentioned. Might take some extra time as it seems a bit more complicated than I thought.

Atomic/syscontainers.py
giuseppe: yes I agree with @ashcrow, this problem must be solved in the RPM .spec file.

If you'd still like to raise a nicer error message, you could parse the `skopeo --version` output.

giuseppe:
can you please rebase this PR?  I am not sure you inherited this change: https://github.com/projectatomic/atomic/pull/1184

About checking for errors:

The first check is to ensure all needed layers are present if for any reason they are missing.  If there is any missing layer, then the image is re-pulled.

We need to check for errors also during the checkout as there are other kind of errors that cannot be found earlier, or the repo could have been changed in the meanwhile.

Once https://github.com/ostreedev/ostree/pull/1292 lands in OSTree, we definitely should start using it for system containers, as the pruning is not really safe while we are pulling other contents.

Atomic/syscontainers.py
peterbaouoft: Sounds good. /me will spend time learning how to include files in RPM .spec, and push out a PR once figured out how.

Atomic/syscontainers.py
```diff
@@ -2079,6 +2096,7 @@ def get_manifest(self, image, remote=False):
 
         imagebranch = SystemContainers._get_ostree_image_branch(image)
         commit_rev = repo.resolve_rev(imagebranch, True)
+        # todo: ask question about here
         if not commit_rev:
             raise ValueError("Layer not found: %s.  Please pull the image again" % imagebranch.replace("sha256:", ""))
```
peterbaouoft: This part feels a bit weird here.  `commit_rev` itself is a tuple right? and for other cases when commit_rev, we all output an internal error: E.g: in `_check_system_ostree_image`: Is that intended behavior? 
```
 current_rev = repo.resolve_rev(imagebranch, True)
        if not current_rev:
            raise ValueError("Internal error for image: %s.  Please pull the image again" % imagebranch)
```

Atomic/syscontainers.py
```diff
@@ -2522,9 +2540,11 @@ def _ensure_storage_for_image(self, repo, img):
         if manifest is None:
             raise ValueError("Image `%s` not present" % img)
         layers = SystemContainers.get_layers_from_manifest(json.loads(manifest))
+        SystemContainers._check_for_valid_layers(repo, layers)
```
peterbaouoft: This check can also be dropped, if the perf impact for checking is large, but it is good to have :).

Atomic/syscontainers.py
ashcrow: Deferring to @giuseppe

ashcrow:
Another rebase is needed.

peterbaouoft:
@ashcrow. Yea, srry  I got sick in the afternoon, forgot to do a rebase. Was still investigating weather I could replace some of the code logic for importing docker images to ostree. For images such as docker:firewalld:latest, we have our own logic for importing it to ostree there, but I am still investigating if it is possible to replace that with skopeo instead. 

 Should of have some progress today(if I am not sick again)  and will do a rebase afterwards :-P

peterbaouoft:
Rebased and pushed newer commits above. There are a few changes compared to the previous commits. 
1: the refactor for missing layer will be temporarily removed, decided to open another pull request for that
2: Now, when user do `atomic pull --storage ostree docker:xxx:latest`, the image will default to be pulled using skopeo, and the old importing_layers_to_ostree will no longer be used. The old functionality of those are also removed.
3: Tests related to image size check is also commented out, based on discussion with giuseppe. It seems like that skopeo has a different way of size changing. Therefore, comment that out for now as well.

peterbaouoft:
```
ValueError: Image with name centos:latest already exists in the Ostree Backend
```

Hmm, I  am a bit confused about the error here, any clarification will be greatly appreciated :P. In my thoughts, I think when we do a `pull_docker_tar`, we should first check if the image with the same name exists in Ostree or not. If it is, we should avoid copying images, otherwise the original image with same name will get overwritten.

However, in `pull.py`. I am not sure if `docker:centos` image got pulled twice or not.. I am confused about the implementation in test_pull.py. Hmm looks like centos image never existed in the local or docker daemon, I am not sure why it will appear inside the ostree backend all the sudden.. Unless part of my logic was wrong... 
``` 
    class Args():
        def __init__(self):
            self.image = "docker:centos:latest"
            self.user = False
    def test_pull_as_privileged_user(self):
        args = self.Args()
        testobj = SystemContainers()
        testobj.set_args(args)
        testobj.pull_image()
```

Any ideas on why above error happened @giuseppe ? ( The value error was raised by my code, but I am not sure why, it should not if the image does not exist in the ostree backend before)

Atomic/syscontainers.py
```diff
@@ -2131,84 +2127,56 @@ def get_layers_from_manifest(manifest):
...
+        image_name = SystemContainers._get_image_name_from_dockertar(tarpath) or image
+        imagebranch = "%s%s" % (OSTREE_OCIIMAGE_PREFIX, SystemContainers._encode_to_ostree_ref(image_name))
+        image_rev = repo.resolve_rev(imagebranch, True)
 
-            repo.transaction_set_ref(None, "%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer), csum)
+        # Before copying into ostree, we want to first verify if image with same name exists or not
+        # to avoid accidental image overrides
+        if not image_rev:
+            raise ValueError("Internal error for image: %s.  Please pull the image again" % imagebranch)
+        if image_rev[1]:
```
peterbaouoft: This line checks if the image rev exists in ostree backend. But it also causes the unit test failure in `test_pull.py`. Not sure why that happened yet...

ashcrow:
> In my thoughts, I think when we do a pull_docker_tar, we should first check if the image with the same name exists in Ostree or not. 

Keep in mind ```centos:latest``` may not be the same as ```centos:latest``` :smile:. Since tags can point to a specific hash it would likely be better to check for the hash(es) rather than the name.

peterbaouoft:
>Keep in mind centos:latest may not be the same as centos:latest 

@ashcrow, that's true, but the way how the image is stored as ostree refs will be the same right? For instance: image with name centos:latest will always be encoded as "centos_3Alatest" right?

Then, in that case the image with same name will be overwritten as shown by the following example, and I think we don't want that no? 
Before: 
```
[root@localhost atomic]# atomic images list
   REPOSITORY                          TAG      IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
>  atomic-test-secret                  latest   6c9f1d7ad457   2018-02-16 21:03   68.0 B         docker    
   firewalld                           latest   52d78869b513   2018-02-16 14:44   273.73 MB      docker    
   atomic-test-system-update           latest   5ee637aaef58   2018-02-12 20:13   228.09 MB      docker    
   atomic-test-system-hostfs           latest   c0933affae2c   2018-02-12 20:13   228.09 MB      docker    
   atomic-test-system                  latest   67c11abc7d79   2018-02-12 20:13   228.09 MB      docker    
   atomic-test-runonce                 latest   d3ced6a01e9a   2018-02-12 20:12   207.2 MB       docker    
>  atomic-test-6                       latest   65516bdc753f   2018-02-12 20:12   207.2 MB       docker    
   atomic-test-5                       latest   807bfdef358c   2018-02-12 20:12   207.2 MB       docker    
   atomic-test-4                       latest   3854bdd56484   2018-02-12 20:12   207.2 MB       docker    
   atomic-test-3                       latest   8d84031818f0   2018-02-12 20:12   207.2 MB       docker    
   atomic-test-1                       latest   0d243d3875f3   2018-02-12 20:12   207.2 MB       docker    
>  devpet                              latest   b60d72c299c6   2018-01-22 19:13   1.02 GB        docker    
   registry.fedoraproject.org/fedora   27       9881e4229c95   2018-01-18 07:48   252.33 MB      docker    
   registry.fedoraproject.org/fedora   latest   9881e4229c95   2018-01-18 07:48   252.33 MB      docker    
   docker.io/centos                    latest   ff426288ea90   2018-01-08 19:58   207.2 MB       docker    
>  docker.io/hello-world               latest   f2a91732366c   2017-11-21 00:23   1.85 KB        docker    
   docker.io/centos                    latest   7e94d6055269   2018-02-26 14:39   73.67 MB       ostree
```
Then we do the pull 
```
[root@localhost atomic]# ./atomic pull --storage ostree docker:docker.io/centos:latest
Getting image source signatures
Copying blob sha256:e15afa4858b655f8a5da4c4a41e05b908229f6fab8543434db79207478511ff7
 205.76 MB / 205.76 MB [====================================================] 7s
Copying config sha256:ff426288ea903fcf8d91aca97460c613348f7a27195606b45f19ae91776ca23d
 1.82 KB / 1.82 KB [========================================================] 0s
Writing manifest to image destination
Storing signatures
[root@localhost atomic]# ostree refs --repo=/var/lib/containers/atomic/.storage/ostree
ociimage/docker.io_2Fcentos_3Alatest
ociimage/e15afa4858b655f8a5da4c4a41e05b908229f6fab8543434db79207478511ff7
ociimage/af4b0a2388c69010cf675c050e51cb1fabbdf2303f955c31805b280324fd4523
ociimage/ff426288ea903fcf8d91aca97460c613348f7a27195606b45f19ae91776ca23d
```
Now, the image seems get overwritten :(
```[root@localhost atomic]# atomic images list 
   REPOSITORY                          TAG      IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
>  atomic-test-secret                  latest   6c9f1d7ad457   2018-02-16 21:03   68.0 B         docker    
   firewalld                           latest   52d78869b513   2018-02-16 14:44   273.73 MB      docker    
   atomic-test-system-update           latest   5ee637aaef58   2018-02-12 20:13   228.09 MB      docker    
   atomic-test-system-hostfs           latest   c0933affae2c   2018-02-12 20:13   228.09 MB      docker    
   atomic-test-system                  latest   67c11abc7d79   2018-02-12 20:13   228.09 MB      docker    
   atomic-test-runonce                 latest   d3ced6a01e9a   2018-02-12 20:12   207.2 MB       docker    
>  atomic-test-6                       latest   65516bdc753f   2018-02-12 20:12   207.2 MB       docker    
   atomic-test-5                       latest   807bfdef358c   2018-02-12 20:12   207.2 MB       docker    
   atomic-test-4                       latest   3854bdd56484   2018-02-12 20:12   207.2 MB       docker    
   atomic-test-3                       latest   8d84031818f0   2018-02-12 20:12   207.2 MB       docker    
   atomic-test-1                       latest   0d243d3875f3   2018-02-12 20:12   207.2 MB       docker    
>  devpet                              latest   b60d72c299c6   2018-01-22 19:13   1.02 GB        docker    
   registry.fedoraproject.org/fedora   27       9881e4229c95   2018-01-18 07:48   252.33 MB      docker    
   registry.fedoraproject.org/fedora   latest   9881e4229c95   2018-01-18 07:48   252.33 MB      docker    
   docker.io/centos                    latest   ff426288ea90   2018-01-08 19:58   207.2 MB       docker    
>  docker.io/hello-world               latest   f2a91732366c   2017-11-21 00:23   1.85 KB        docker    
   docker.io/centos                    latest   dab49f73e179   2018-02-26 14:53   215.75 MB      ostree
```

ashcrow:
> @ashcrow, that's true, but the way how the image is stored as ostree refs will be the same right?

TBH I'm not sure. I'll defer to @giuseppe 

> Now, the image seems get overwritten :(

"Overwriting" makes sense I think as a tag that has moved means the underlying image has changed. It might make sense to rename the previous version in case one has to roll back ... but that would likely be out of scope for this work.

Atomic/syscontainers.py
```diff
@@ -2131,84 +2127,56 @@ def get_layers_from_manifest(manifest):
...
+            raise ValueError("Image with name {} already exists in the Ostree Backend".format(image_name))
```
giuseppe: please drop this error, the image must be overwritten

peterbaouoft: Sounds good, and one more minor question, do you want the missing layer check( the one you had in `_check_system_oci_image` to be added here as well? Or, we can choose to drop the ostree resolve rev checking part here completely.

ashcrow: @peterbaouoft choose one and let's see what happens :smile_cat:

ashcrow:
Ready for full review?

giuseppe:
> "Overwriting" makes sense I think as a tag that has moved means the underlying image has changed. It might make sense to rename the previous version in case one has to roll back ... but that would likely be out of scope for this work.

the rollback as implemented by `atomic containers rollback` to the previous version installed does not require the old image to be present in the OSTree repository.  IMO if the user for any reason wants to keep the old version around a tag could be used (`atomic images tag`).

peterbaouoft:
@ashcrow yup, i think it should be ok to review :)

ashcrow:
@peterbaouoft great! I'll take a look shortly.

Atomic/syscontainers.py
```diff
@@ -2129,84 +2125,47 @@ def get_layers_from_manifest(manifest):
...
-                root = repo.write_mtree(mtree)[1]
-
-                metav = GLib.Variant("a{sv}", {'docker.layer': GLib.Variant('s', layer),
-                                               'docker.size': get_directory_size(temp_dir)})
-                csum = repo.write_commit(None, "", None, metav, root)[1]
-            finally:
-                shutil.rmtree(temp_dir, ignore_errors=True)
+    def _pull_docker_image(self, image):
+        with tempfile.NamedTemporaryFile(mode="w") as temptar:
+            util.check_call(["docker", "save", "-o", temptar.name, image])
```
giuseppe: could we use skopeo directly here as well?

"skopeo copy docker-daemon:.."

Atomic/syscontainers.py
```diff
@@ -7,7 +7,7 @@
 from string import Template
 import calendar
 import shutil
-import stat # pylint: disable=bad-python3-import
+# pylint: disable=bad-python3-import
```
ashcrow: I think this can be dropped.

Atomic/syscontainers.py
peterbaouoft: ah, did not realize that can also be done via docker-daemon. Thanks for pointing it out, will do so.

Atomic/syscontainers.py
```diff
@@ -2129,84 +2125,47 @@ def get_layers_from_manifest(manifest):
...
-
-                repo.write_directory_to_mtree(Gio.File.new_for_path(temp_dir), mtree, modifier)
-                root = repo.write_mtree(mtree)[1]
-
-                metav = GLib.Variant("a{sv}", {'docker.layer': GLib.Variant('s', layer),
-                                               'docker.size': get_directory_size(temp_dir)})
-                csum = repo.write_commit(None, "", None, metav, root)[1]
-            finally:
-                shutil.rmtree(temp_dir, ignore_errors=True)
+    def _pull_docker_image(self, image):
```
ashcrow: Please add docstring. Example:

```python
def _pull_docker_image(self, image):
    """
    pulls a docker image down to a local tar archive.

    :param image: The full name of the image to pull down.
    :type image: str
    :returns: What does this return?
    :rtype: What's the type?
    """
    ...
```

Atomic/syscontainers.py
```diff
@@ -2129,84 +2125,47 @@ def get_layers_from_manifest(manifest):
...
+        with tempfile.NamedTemporaryFile(mode="w") as temptar:
+            util.check_call(["docker", "save", "-o", temptar.name, image])
+            return self._pull_docker_tar(temptar.name, image)
 
-            repo.transaction_set_ref(None, "%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer), csum)
+    def _pull_docker_tar(self, tarpath, image):
```
ashcrow: docstring :smile:

Atomic/syscontainers.py
```diff
@@ -2129,84 +2125,47 @@ def get_layers_from_manifest(manifest):
...
-        repo.commit_transaction(None)
+        checkout = self._get_system_checkout_path()
+        destdir = checkout if os.path.exists(checkout) else None
+        temp_dir = tempfile.mkdtemp(prefix=".", dir=destdir)
 
-    def _pull_docker_image(self, repo, image):
-        with tempfile.NamedTemporaryFile(mode="w") as temptar:
-            util.check_call(["docker", "save", "-o", temptar.name, image])
-            return self._pull_docker_tar(repo, temptar.name, image)
+        destination = "ostree:{}@{}".format(img, repo)
```
ashcrow: :+1: for ``format()``

tests/integration/test_system_containers_images.sh
```diff
@@ -106,29 +106,30 @@ ${ATOMIC} images list -f type=ostree > ${WORK_DIR}/images.out
 assert_matches "atomic-test-system" ${WORK_DIR}/images.out
 
 
+# This part got commented out because of the new skopeo copy way
```
ashcrow: If these are no longer needed then go ahead and remove them. If they are still legit tests please reenable or replace them.

peterbaouoft: sounds good, will do that when forcing push new commits. :)

Atomic/syscontainers.py
peterbaouoft: yup, sure, I kinda got lazy when the method itself is short. :P

Atomic/syscontainers.py
peterbaouoft: Thanks for the review! and ah, yea.. i was thinking about removing it earlier, but then. there are pylint errors after I remove those. Hmm, weirdo, the pylint error no longer showed up this time, #facepalm. Will drop it then :P.

Atomic/syscontainers.py
ashcrow: Ha! I understand. One of the things I'm trying to push in ``atomic`` is to get more code docs so even documenting the small ones helps others who jump in later understand what's going on.

peterbaouoft:
Force pushed a new version addressing the comments. Things changed since the last commit:
1: docstring added 
2: docker-daemon skopeo copy added 
3: fixup got squashed and the commented out section is removed (as suggested :-) )
4: noticed formatting issues (extra spaces, misspelled words). 

PTAL, thanks in advance :)

tests/integration/setup-scripts/system_containers_setup.sh
```diff
@@ -43,3 +43,7 @@ export PYTHON=${PYTHON:-/usr/bin/python}
 export ATOMIC_OSTREE_REPO=${WORK_DIR}/repo
 export ATOMIC_OSTREE_CHECKOUT_PATH=${WORK_DIR}/checkout
 export NAME="test-system-container-$$"
+
+# This is to prevent the case where the ostree checkout path
+# can be non-existent when no container installation happens prior
+mkdir -p $ATOMIC_OSTREE_CHECKOUT_PATH
```
giuseppe: why do we need to do this?

tests/integration/test_system_containers_images.sh
```diff
@@ -105,31 +105,6 @@ ${ATOMIC} pull --storage ostree docker:atomic-test-system:latest
...
-
-cat > test_images_size.py <<EOF
-#!/usr/bin/env python
-import json
-import sys
-def make_number(x):
-    foo = x.replace("MB", "").replace("KB", "").replace(" ","")
-    return float(foo)
-
-def is_different(x, y):
```
giuseppe: how much is different when copying with Skopeo?  If it helps we can adjust the threshold

tests/integration/setup-scripts/system_containers_setup.sh
peterbaouoft: ah, I noticed that the checkout path was never created until someone does a container installation. 
However, we will fail if the checkout path is not there for some actions( let's say we do `atomic images prune` ). 
```
# Check for dangling checkouts that couldn't be deleted on uninstall.
        checkouts = self._get_system_checkout_path()
        for i in os.listdir(checkouts):
```

It will error out for the above code when the checkout path is not existent. Therefore, to avoid such cases, we first make sure the checkout path exist. I caught the error when writing an integration test, and added it.

tests/integration/test_system_containers_images.sh
peterbaouoft: Hmm, I am not entirely certain yet. I have only investigated a bit for images like `hello-world` or `docker-fedora`, and the difference seems to be varying. For hello-world, it is 2KB extra, and for docker-fedora, it was like 15 Mb extra. I would assume if we want to check for threshold, we might need to calculate a certain percentage difference. 

Let me investigate that a bit.

tests/integration/setup-scripts/system_containers_setup.sh
giuseppe: but in our tests we always install a container before we try a prune right?  Does the issue happen in our tests suite without this change?

peterbaouoft: Hmm, I added a new test, and in that test I tried to clean up the image each time I pull the image to ostree backend. Then the error happens if I did a `atomic images prune` without the checkout folder. To avoid it, i added the above line. Is that the correct way to do so?

giuseppe: I think in general `atomic images prune` should not give an error if the checkout directory doesn't exist, but this is out of the scope of this PR.

If the tests pass without creating the checkout manually, let's drop it

peterbaouoft: > If the tests pass without creating the checkout manually, let's drop it

ah, sorry for explaining it earlier, the test I add( system_containers_pull.sh) will fail if the checkout was not created. It will fail when the test is run by itself via (make test TEST_INTEGRATION=system_containers_pull). So, let's keep it in this PR for now, and I can address the above issue you mentioned in a separate PR?  WDYT

tests/integration/test_system_containers_images.sh
peterbaouoft: yea.. sorry, investigated that a bit, seems like there wasn't an obvious relationship between image size. I have pulled 4 different images from docker to ostree. Here are the results of those:
```
[root@localhost atomic]# atomic images list | grep atomic-test-4
   atomic-test-4               latest   39f9d7398cc3   2018-02-21 22:47   207.2 MB       docker    
   atomic-test-4               latest   99cab1cd1273   2018-02-28 19:59   215.75 MB      ostree

[root@localhost atomic]# atomic images list | grep docker.io/hello-world
   docker.io/hello-world       latest   f2a91732366c   2017-11-21 00:23   1.85 KB        docker    
   docker.io/hello-world       latest   8072a54ebb3b   2018-02-28 19:48   985.0 B        ostree

[root@localhost atomic]# atomic images list | grep atomic-test-system
   atomic-test-system          latest   bb3c8d007810   2018-02-28 19:53   228.09 MB      docker    
   atomic-test-system          latest   56db3c2331ac   2018-02-28 19:54   236.74 MB      ostree

[root@localhost atomic]# atomic images list | grep atomic-test-secret
   atomic-test-secret          latest   6a5b9db82a6f   2018-02-28 19:53   68.0 B         docker    
   atomic-test-secret          latest   1b6853fe7ec4   2018-02-28 19:47   68.0 B         ostree 
```

One thing I do notice tho, we are copying blob of sha256 checksum, and config each time via skopeo, may be we have a different way of calculate it between skopeo and docker?

peterbaouoft: yea, I believe so.. cuz we are only testing it against one image right? For now, to keep the code there, we can hard code the threshold, and I am going to have it to be 4%( that is the difference for atomic-test-system). Will push a fixup soon :)

peterbaouoft: Force pushed a new version below with size threshold relaxed. PTAL :)

Atomic/syscontainers.py
```diff
@@ -2129,84 +2130,69 @@ def get_layers_from_manifest(manifest):
...
-        mtree.set_metadata_checksum(OSTree.checksum_from_bytes(csum_dirmeta))
+        # Return the image name for future usages
+        return image_name
 
-        root = repo.write_mtree(mtree)[1]
-        csum = repo.write_commit(None, "", None, metadata, root)[1]
-        repo.transaction_set_ref(None, imagebranch, csum)
+    def _skopeo_copy_img_to_ostree(self, img, skopeo_img_source, src_creds=None, insecure=None):
+        """
+        Pefrom skopeo copy operation to copy images from img_source to ostree location based on image
```
giuseppe: s|Pefrom|Perform|

Atomic/syscontainers.py
```diff
@@ -2129,84 +2130,69 @@ def get_layers_from_manifest(manifest):
...
+        name
+        :param img: image name
+        :param skopeo_img_source: the source for skopeo copying
+        :src_creds: source credientials if pulled from online registries
```
giuseppe: s|credientials|credentials|

giuseppe:
just two minor things and I think it is ready to go!

Atomic/syscontainers.py
ashcrow: nit:
```
:param src_creds: ...
:type src_creds: ...
```

Atomic/syscontainers.py
```diff
@@ -2129,84 +2130,69 @@ def get_layers_from_manifest(manifest):
...
+        :insecure: tell skopeo whether the registry is secure or not
```
ashcrow: nit:
```
:param insecure: ...
:type insecure: ...
```

Atomic/syscontainers.py
```diff
@@ -2129,84 +2130,69 @@ def get_layers_from_manifest(manifest):
...
+        :returns: True if successfully performed copy operation, a ValueError can be raised upon failures
```
ashcrow: nit:

```
:rtype: bool
```

Atomic/syscontainers.py
```diff
@@ -2129,84 +2130,69 @@ def get_layers_from_manifest(manifest):
...
+        finally:
+            shutil.rmtree(temp_dir, ignore_errors=True)
+        return True
 
-    def _pull_docker_tar(self, repo, tarpath, image):
+    @staticmethod
+    def _get_image_name_from_dockertar(tarpath):
+        """
+        Retrieve the image name from the dockertar file
+        :param tarpath: the path of the docker tar file
```
ashcrow: nit: Missing ``:type tarpath:``

Atomic/syscontainers.py
```diff
@@ -2129,84 +2130,69 @@ def get_layers_from_manifest(manifest):
...
+        :returns: the true image name got from the dockertar
```
ashcrow: nit: Missing ``:rtype:``

peterbaouoft:
Thanks for the review :), pushed a fixup to address the comments above. There is a slight concern for me now tho which is, we have to enforce the version of skopeo after adopting this patch right, and do we need to inform other folks(e.g community) about this migration?
  
That said, I strongly agree with @giuseppe that we should move towards "skopeo only" path for copying images to ostree, for both maintainability and performance reasons.  Tho, I am not too familiar with handling changes that may have "larger impact"(I think, may be I overthought about it tho :p). Any suggestions will be appreciated :D.

ashcrow:
> Thanks for the review :), pushed a fixup to address the comments above. There is a slight concern for me now tho which is, we have to enforce the version of skopeo after adopting this patch right, and do we need to inform other folks(e.g community) about this migration?

It should be part of the ``README.md`` and the rpm spec should require ``>=`` the version for ``skopeo``

Atomic/syscontainers.py
```diff
@@ -2165,13 +2165,18 @@ def _pull_docker_tar(self, tarpath, image):
 
     def _skopeo_copy_img_to_ostree(self, img, skopeo_img_source, src_creds=None, insecure=None):
         """
-        Pefrom skopeo copy operation to copy images from img_source to ostree location based on image
+        Peform skopeo copy operation to copy images from img_source to ostree location based on image
```
giuseppe: Perform :-)

giuseppe:
we don't have to worry about that, we already require Skopeo for stuff like SELinux labelling and that is enforced in the .rpm spec file AFAIK.  Also, getting an error when using an older version is an improvement :-)

Atomic/syscontainers.py
peterbaouoft: oops, ah, missed an r , thanks for pointing it out :)

giuseppe:
Thanks!  LGTM, @ashcrow any objections to the last version?

ashcrow:
Nope! :shipit:

giuseppe:
so I pull the trigger!

@rh-atomic-bot r+ 476b47a

Issue #1179, opened 2018-02-06
'atomic images list' is missing some system containers


miabbott:
On Fedora 27 AH, using `atomic-1.20.1-9.fc27.x86_64`, the output from `atomic images list` is missing some system containers that were pulled to `ostree` storage.  However, if I pass in the `--all` option, the missing images show up.

```
# atomic images list
   REPOSITORY                                                     TAG      IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
   registry.fedoraproject.org/fedora                              latest   9881e4229c95   2018-01-18 07:48   252.33 MB      docker    
>  docker.io/registry                                             2        d1fd7d86a825   2018-01-10 01:22   33.28 MB       docker    
   registry.fedoraproject.org/f27/kubernetes-proxy                latest   4660f3d3b9a3   2018-02-06 16:43   262.53 MB      ostree    
   registry.fedoraproject.org/f27/kubernetes-kubelet              latest   3c419c4956d8   2018-02-06 16:42   267.37 MB      ostree    
   registry.fedoraproject.org/f27/kubernetes-controller-manager   latest   cc1f95c6fe77   2018-02-06 16:40   231.12 MB      ostree    
   registry.fedoraproject.org/f27/etcd                            latest   4501c21370c0   2018-02-06 16:43   111.26 MB      ostree    

# atomic images list --all
   REPOSITORY                                                     TAG      IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
   registry.fedoraproject.org/fedora                              latest   9881e4229c95   2018-01-18 07:48   252.33 MB      docker    
>  docker.io/registry                                             2        d1fd7d86a825   2018-01-10 01:22   33.28 MB       docker    
   <none>                                                         <none>   8d884f5d20c0   2018-02-06 16:43                  ostree    
   <none>                                                         <none>   ff8eae79cd9d   2018-02-06 16:43                  ostree    
   <none>                                                         <none>   a0d3b3cff988   2018-02-06 16:37                  ostree    
   registry.fedoraproject.org/f27/kubernetes-proxy                latest   4660f3d3b9a3   2018-02-06 16:43   262.53 MB      ostree    
   <none>                                                         <none>   9d695cfffacf   2018-02-06 16:40                  ostree    
   <none>                                                         <none>   e8d4ef1c74c2   2018-02-06 16:42                  ostree    
   <none>                                                         <none>   1eb9b88722dd   2018-02-06 16:37                  ostree    
[...]

PR #1181, opened 2018-02-07
syscontainer: fix for missing images display in ostree backend

peterbaouoft:
This fixes https://github.com/projectatomic/atomic/issues/1179.

Before, we assumed the image ref with 64 characters is default to
image layers(images with repository <none>), but as issue 1179
pointed out, there are some exceptions.

We now migrated to check if the ref is hex. If the ref is hex, we
can conclude it should be images with repository <none>

## Related Bugzillas
-
-

## Related Issue Numbers
- https://github.com/projectatomic/atomic/issues/1179.
-

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1702,7 +1702,7 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             if repo is None:
                 return []
         revs = [x for x in repo.list_refs()[1] if x.startswith(OSTREE_OCIIMAGE_PREFIX) \
-                and (get_all or SystemContainers.is_hex(get_img_ref_str(x)))]
+                and (get_all or not SystemContainers.is_hex(get_img_ref_str(x)))]
 
         return [self._inspect_system_branch(repo, x) for x in revs]
 
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1623,12 +1623,13 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         commit = repo.load_commit(commit_rev)[1]
 
         timestamp = OSTree.commit_get_timestamp(commit)
-        branch_id = SystemContainers._decode_from_ostree_ref(imagebranch.replace(OSTREE_OCIIMAGE_PREFIX, ""))
+        img_ref_str = SystemContainers._get_img_ref_str(imagebranch)
+        branch_id = SystemContainers._decode_from_ostree_ref(img_ref_str)
 
         image_id = commit_rev
         id_ = None
 
-        if SystemContainers.is_hex(branch_id):
+        if SystemContainers._is_hex(branch_id):
             image_id = branch_id
             tag = "<none>"
         elif '@sha256:' in branch_id:
@@ -1694,20 +1695,21 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         :returns: A list of image data structures.
         :rtype: list
         """
-        def get_img_ref_str(ostree_ref_str):
-            return ostree_ref_str.replace(OSTREE_OCIIMAGE_PREFIX, "")
-
         if repo is None:
             repo = self._get_ostree_repo()
             if repo is None:
                 return []
         revs = [x for x in repo.list_refs()[1] if x.startswith(OSTREE_OCIIMAGE_PREFIX) \
-                and (get_all or not SystemContainers.is_hex(get_img_ref_str(x)))]
+                and (get_all or not SystemContainers._is_hex(SystemContainers._get_img_ref_str(x)))]
 
         return [self._inspect_system_branch(repo, x) for x in revs]
 
     @staticmethod
-    def is_hex(s):
+    def _get_img_ref_str(ostree_ref_str):
+        return ostree_ref_str.replace(OSTREE_OCIIMAGE_PREFIX, "")
+
+    @staticmethod
+    def _is_hex(s):
         try:
             int(s, 16)
             return True
@@ -1923,8 +1925,8 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
 
         for i in repo.list_refs()[1]:
             if i.startswith(OSTREE_OCIIMAGE_PREFIX):
-                img_ref_str = i.replace(OSTREE_OCIIMAGE_PREFIX, "")
-                if SystemContainers.is_hex(img_ref_str):
+                img_ref_str = SystemContainers._get_img_ref_str(i)
+                if SystemContainers._is_hex(img_ref_str):
                     refs[i] = False
                 else:
                     invalid_encoding = False
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1628,7 +1628,8 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         image_id = commit_rev
         id_ = None
 
-        if SystemContainers.is_hex(branch_id):
+        is_layer_ref = SystemContainers.is_hex(branch_id) and len(branch_id) == 64
+        if is_layer_ref:
             image_id = branch_id
             tag = "<none>"
         elif '@sha256:' in branch_id:
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1628,7 +1628,7 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         image_id = commit_rev
         id_ = None
 
-        if len(branch_id) == 64:
+        if SystemContainers.is_hex(branch_id):
             image_id = branch_id
             tag = "<none>"
         elif '@sha256:' in branch_id:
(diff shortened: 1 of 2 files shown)


peterbaouoft:
```
[root@localhost atomic]# atomic images list
   REPOSITORY                                            TAG      IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
>  registry.fedoraproject.org/f27/kubernetes-scheduler   latest   cb17034be1ed   2018-02-07 17:48   231.12 MB      ostree    
>  hello-world                                           latest   8072a54ebb3b   2018-02-07 18:11   985.0 B        ostree    
[root@localhost atomic]# ostree refs --repo=/var/lib/containers/atomic/.storage/ostree
ociimage/ca4f61b1923c10e9eb81228bd46bee1dfba02b9c7dac1844527a734752688ede
ociimage/ff8eae79cd9d3027cc810b5d3ab0510a859ca7aab77910d7addcc28de7c00db9
ociimage/hello-world_3Alatest
ociimage/1eb9b88722dda2113c3e55a6436d733e4feea21e772b854dbe7f92010b46eafb
ociimage/f2a91732366c0332ccd7afd2a5c4ff2b9af81f549370f7a19acd460f87686bc7
ociimage/2de259afc9c4d080d3a8101a5909db2a5603cf837ac65c9e61896dc28b81fb05
ociimage/04331e646521ddb577d113f3c103aef620cc4451641452c347864298669f8572
ociimage/registry.fedoraproject.org_2Ff27_2Fkubernetes-scheduler_3Alatest
```

Seems the image can now be shown properly, @giuseppe  PTAL

ashcrow:
CI issue opened by @peterbaouoft: https://bugs.centos.org/view.php?id=14466

peterbaouoft:
>CI issue opened by @peterbaouoft: https://bugs.centos.org/view.php?id=14466

@ashcrow  this pull request is a different pull request actually =P. This pull request supposed to fix the bug micah reported #1179

jlebon:
The `centos/7/atomic` image just tracks RHELAH releases, so it's always going to be slower. If you want newer things, you can use `centos/7/atomic/smoketested` instead. See https://github.com/projectatomic/papr/blob/master/docs/DISTROS.md for more details.

jlebon:
That said, skopeo is not a package we currently [overlay](https://github.com/CentOS/sig-atomic-buildscripts/pull/245). So you can either special-case the specific tests using feature detection, or see about getting skopeo added to the virt7 SIG list?

peterbaouoft:
:o, thanks for the quick feedback @jlebon :-).
> So you can either special-case the specific tests using feature detection, or see about getting skopeo added to the virt7 SIG list

/me do not know how to do both of them. Any documents I can read to fill in the knowledge gap?

jlebon:
E.g. you can do something like:

```
supports_ostree_copy() {
  skopeo copy --help | grep -q ostree
}
```

or similar and then check that before either skipping or running the test.

This is something we sometimes have to do as well in rpm-ostree tests: https://github.com/projectatomic/rpm-ostree/blob/6c933bbf3c9e395085946eb51eef004cc4405b80/tests/vmcheck/overlay.sh#L92.

Hope that helps!

Atomic/syscontainers.py
```diff
@@ -1694,15 +1694,26 @@ def get_system_images(self, get_all=False, repo=None):
...
         """
+        def get_img_ref_str(ostree_ref_str):
+            return ostree_ref_str.replace(OSTREE_OCIIMAGE_PREFIX, "")
+
         if repo is None:
             repo = self._get_ostree_repo()
             if repo is None:
                 return []
         revs = [x for x in repo.list_refs()[1] if x.startswith(OSTREE_OCIIMAGE_PREFIX) \
-                and (get_all or len(x) != len(OSTREE_OCIIMAGE_PREFIX) + 64)]
```
giuseppe: I think it is better if we leave also the condition on the length.  Otherwise we will catch images like `cafe`, `c0`..

Atomic/syscontainers.py
```diff
@@ -1628,7 +1628,7 @@ def _inspect_system_branch(self, repo, imagebranch):
         image_id = commit_rev
         id_ = None
 
-        if len(branch_id) == 64:
+        if SystemContainers.is_hex(branch_id):
```
peterbaouoft: Here we should also check the length of branch id ==64, because there could be a case, that cafe is the image as giuseppe pointed out earlier

Atomic/syscontainers.py
peterbaouoft: Hmm, Looking more into the code. Looks like we encode the ref via the following function.
`_encode_to_ostree_ref`
```
 def convert(x):
            return (x if str.isalnum(str(x)) or x in '.-' else "_%02X" % ord(x))

        if name.startswith("oci:"):
            name = name[len("oci:"):]
        registry, image, tag = SystemContainers._parse_imagename(name)
        if registry:
            fullname = "%s/%s:%s" % (registry, image, tag)
        else:
            fullname = "%s:%s" % (image, tag)

        ret = "".join([convert(i) for i in fullname])
        return ret
```

and in parse_image_name
```
 sep = image.find(":")
        if sep > 0:
            return reg, image[:sep], image[sep + 1:]
        else:
            return reg, image, "latest"
```

After digging a bit more, we seem always include a ":" for tagging when we encode the ref, no matter what the tag is. (i.e: either latest or other). The encoded ref itself will then always have a ":" which get encoded to `_3A` in unicode. In this case, I still suggest we drop the check of the length for encoded ref here?

But I am also fine with adding it back in this place, if you feel strongly about this =P.

ashcrow: I'll defer to @giuseppe.

Atomic/syscontainers.py
```diff
@@ -1628,7 +1628,8 @@ def _inspect_system_branch(self, repo, imagebranch):
         image_id = commit_rev
         id_ = None
 
-        if len(branch_id) == 64:
+        is_layer_ref = SystemContainers.is_hex(branch_id) and len(branch_id) == 64
```
giuseppe: if we don't use the `len(branch_id) == 64` condition, should we drop it here as well?

peterbaouoft: Hmm, /me thinks. yea you are right! Thanks for pointing it out. Before I was worried that the decoded ref will be the case that you mentioned earlier (i.e: cafe). But thinking more about it, I realized that, when you decode from the ostree ref of a valid image, you will still get the tag, which means the branch id will contain a ":".

As a result, the valid image( image that are not layers and with valid repos / tags) will not be considered as hex due to an extra ":". 

I will have a fix for it, thanks.

peterbaouoft:
Pushed a fixup above. One more question about this PR is that, for bug fix, do we need to have a test for it? (/me is not very familiar with fixing bugs). Other than that, I think this PR is ok to go =P

Atomic/syscontainers.py
```diff
@@ -1912,11 +1923,12 @@ def prune_ostree_images(self):
 
         for i in repo.list_refs()[1]:
             if i.startswith(OSTREE_OCIIMAGE_PREFIX):
-                if len(i) == len(OSTREE_OCIIMAGE_PREFIX) + 64:
+                img_ref_str = i.replace(OSTREE_OCIIMAGE_PREFIX, "")
```
giuseppe: can we use `get_img_ref_str(..)` here?

peterbaouoft: sure, working on it :-)

peterbaouoft:
Pushed a fixup. One last thing I want to confirm other than the test: 

For image layers, will its ref always be hex? Will those ref layers sometimes be not hex?  

Other than that, I had a doubt about whether or not to add test for this fix. (mentioned in above comment)

giuseppe:
Layers will always be hexadecimal. If you would like, we could add a test to verify image refs are never stored as hexadecimals so that our assumption will still be correct

peterbaouoft:
> If you would like, we could add a test to verify image refs are never stored as hexadecimals so that our assumption will still be correct

I would like to. I have a few thoughts about the implementations. Both seem to require more work than expected. I am not sure if I am going in the right direction, mind giving some suggestions? 

1: We create a unit test and a system image / docker image. Then we verify it using our is_hex function. But that would require us to create an image

2: We use integration test (e.g busybox image) and use python script to load json and find digest of the layers and to verify if it is hex

But Both cases can not cover all of the image( can not satisfy never case check). Is it possible that we have some kind of trigger that triggers `is_hex` check whenever an image layer is involved? Thoughts on this? =)

giuseppe:
I think we can simply add a unit tests for `_encode_to_ostree_ref` and verify that for different inputs (e.g. fully qualified image with and without a tag, not fully qualified with and without a tag, `@sha256:` notation)  the return value will include a suffix that make it not a hexadecimal string, no matter of the image name.

ashcrow:
Failures are the storage flake:

```
ERROR: Failed to unmount /var/lib/docker /var/lib/docker/containers /var/lib/docker/overlay2
```

peterbaouoft:
> I think we can simply add a unit tests for _encode_to_ostree_ref and verify that for different inputs (e.g. fully qualified image with and without a tag, not fully qualified with and without a tag, @sha256: notation) the return value will include a suffix that makes it not a hexadecimal string, no matter of the image name.

Yes, thanks for the hint! I think it would cover the check for our discussions above. I pushed a commit containing the test.

Tho, One last thing I am concerned is that, and sorry if I did not explain it well last time. I meant more to have a test to cover the following:
> Layers will always be hexadecimal, we could add a test to verify image refs are never stored as hexadecimals so that our assumption will still be correct

I still can't find a neat way to check for every image layer (with `@sha256:`) we have to be hex. But I believe the layers are coming from docker, which usually are hex.  So....... I am not sure if it is worth to check for that atm. Any thoughts on this?

tests/unit/test_syscontainers.py
```diff
@@ -291,6 +291,37 @@ def test_handle_system_package_files(self, _isi, _raf, _gr):
...
+        3: qualified image(image with registry included) and non-qualified image will behave in the same way
+        when calling _encode_to_ostree_ref
+        """
+
+        qual_img_without_tag = "registry.fedoraproject.org/f27/kubernetes-apiserver"
+        qual_img_with_tag = qual_img_without_tag + ":tag"
+        non_qual_img_without_tag = "cafe"
+        non_qual_img_with_tag = non_qual_img_without_tag + ":tag"
+
+        qual_img_without_tag_ref = SystemContainers._encode_to_ostree_ref(qual_img_without_tag)
```
giuseppe: let's check for `self.assertFalse(ref_is_hex)` for all the cases here.

peterbaouoft: ah, the reason why I did not check other cases is that, the initial img itself is already non-hex. But yea, I will add back the check for other imgs as well. 

and btw, did you see my comment above regarding the other case?

giuseppe:
@rh-atomic-bot r+ 3ae2b07

PR #1184, opened 2018-02-12
ci: use centos/7/atomic/smoketested

jlebon:
The `/alpha` image has been deprecated a while ago and now no longer
exists. Use the `/smoketested` image instead. For more information, see:

https://github.com/projectatomic/papr/blob/master/docs/DISTROS.md

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -28,7 +28,7 @@ timeout: 30m
 inherit: true
 
 host:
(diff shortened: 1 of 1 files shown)


jlebon:
@rh-atomic-bot delegate=miabbott

miabbott:
@rh-atomic-bot r+ a4a22cb

PR #1186, opened 2018-02-13
syscontainers: implement "run IMAGE"

giuseppe:
Implement "run --storage ostree IMAGE" so that the image is installed and executed interactively.

I've implemented it so we can replace a call to docker run in the Open Shift installer:

https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_version/tasks/first_master_containerized_version.yml#L22

that could be better done with atomic since we already need the image to run the master.

Another use case in the future could be to run containers such as Buildah, something like:

`atomic run --storage ostree registry.org/buildah bud /host/$(PWD)/ .`

### DIFF
--- a/Atomic/backends/_ostree.py
+++ b/Atomic/backends/_ostree.py
@@ -165,11 +165,7 @@ class OSTreeBackend(Backend):
 
     def run(self, iobject, **kwargs):
         args = kwargs.get('args')
-        try:
-            name = iobject.name
-        except AttributeError:  # iobject isn't populated
-            raise ValueError('Unable to find an image named {} in {}'.format(
-                args.image, args.storage))
+        name = args.image
         if len(args.command) == 0:
             return self.syscontainers.start_service(name)
         self.syscontainers.set_args(args)
--- a/Atomic/backends/_ostree.py
+++ b/Atomic/backends/_ostree.py
@@ -172,6 +172,7 @@ class OSTreeBackend(Backend):
                 args.image, args.storage))
         if len(args.command) == 0:
             return self.syscontainers.start_service(name)
+        self.syscontainers.set_args(args)
         return self.syscontainers.container_exec(name, args.detach, args.command)
 
     def tag_image(self, src, dest):
--- a/Atomic/backends/_ostree.py
+++ b/Atomic/backends/_ostree.py
@@ -166,6 +166,8 @@ class OSTreeBackend(Backend):
     def run(self, iobject, **kwargs):
         args = kwargs.get('args')
         name = args.image
+        if args.name is not None:
+            raise ValueError("--name is not supported by this backend")
         if len(args.command) == 0:
             return self.syscontainers.start_service(name)
         self.syscontainers.set_args(args)
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -2639,6 +2639,19 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
                     s = os.stat(fullpath)
                     os.chmod(fullpath, s.st_mode | 0o600)
 
+    def _rewrite_config_args(self, orig_config, dest_config, args, tty=None, checkout=None):
+        with open(orig_config, 'r') as config_file:
+            config = json.loads(config_file.read())
+            if checkout is not None:
+                rootfs = os.path.realpath(os.path.normpath(os.path.join(checkout, config['root']['path'])))
+                config['root']['path'] = rootfs
+            config['process']['args'] = args
+            if tty is not None:
+                config['process']['terminal'] = tty
+            with open(dest_config, 'w') as conf:
+                conf.write(json.dumps(config, indent=4))
+            return config
+
     def container_exec(self, name, detach, args):
         is_container_running = self._is_service_active(name)
 
@@ -2656,21 +2669,15 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         else:
             checkout = self._canonicalize_location(self.get_checkout(name))
             if checkout is None:
-                raise ValueError("The container '{}' doesn't exist".format(name))
+                return self._run_once(name, "run-once-{}".format(os.getpid()), args=args)
             if detach:
                 raise ValueError("Cannot use --detach with not running containers")
 
             temp_dir = tempfile.mkdtemp()
             try:
                 orig_config = os.path.sep.join([checkout, 'config.json'])
-                with open(orig_config, 'r') as config_file:
-                    config = json.loads(config_file.read())
-                    new_path = os.path.realpath(os.path.normpath(os.path.join(checkout, config['root']['path'])))
-                    config['root']['path'] = new_path
-                    config['process']['args'] = args
-                    config['process']['terminal'] = tty
-                with open(os.path.sep.join([temp_dir, 'config.json']), 'w') as conf:
-                    conf.write(json.dumps(config, indent=4))
+                dest_config = os.path.sep.join([temp_dir, 'config.json'])
+                config = self._rewrite_config_args(orig_config, dest_config, args, tty=tty, checkout=checkout)
 
                 # The runtime directories are usually created by systemd when the
                 # service starts.  Since we are running the container outside of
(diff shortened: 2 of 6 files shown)


giuseppe:
with this patch I can do:

```
# atomic pull --storage ostree docker.io/gscrivano/buildah
# atomic run --storage ostree docker.io/gscrivano/buildah buildah bud --pull /host/$(PWD)

[CUT OUTPUT]

Complete!
STEP 3: COMMIT containers-storage:[overlay@/var/lib/containers/storage+/var/run/containers/storage:overlay.override_kernel_check=true]@36f6b4e59035e4c86cc52969ccd9209b4ea2468ed9afd558604ccead343b0e02
```

The WIP system container for buildah is here:

https://github.com/giuseppe/atomic-system-containers/tree/buildah

tests/integration/test_system_containers_install.sh
```diff
@@ -138,6 +138,9 @@ ${ATOMIC} pull --storage ostree docker:atomic-test-runonce:latest
 ${ATOMIC} install --system --name=${NAME} --set RECEIVER=Pluto atomic-test-runonce:latest > ${WORK_DIR}/ps.out
 assert_matches "HI Pluto from Saturn" ${WORK_DIR}/ps.out
 
+${ATOMIC} run --storage ostree --set RECEIVER=Pluto atomic-test-runonce:latest echo Hello World > ${WORK_DIR}/ps.out
```
peterbaouoft: Atomic run has no `--set` option right? I believe this is causing ci failures.

Atomic/syscontainers.py
```diff
@@ -2628,21 +2647,15 @@ def container_exec(self, name, detach, args):
         else:
             checkout = self._canonicalize_location(self.get_checkout(name))
             if checkout is None:
-                raise ValueError("The container '{}' doesn't exist".format(name))
+                return self._run_once(name, "run-once-{}".format(os.getpid()), args=args)
```
peterbaouoft: > https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_version/tasks/first_master_containerized_version.yml#L22

Wrong line? This is what I saw: 
```
when:
  - openshift_version is not defined
- not openshift_use_crio_only
```
Might not be relevant but just a nit.

tests/integration/test_system_containers_install.sh
giuseppe: ouch, I assumed it had.  I added a patch to implement it as it is very useful for running containers.

Atomic/syscontainers.py
giuseppe: I changed the logic of this code as I was not really happy about it.

Now it attempts to run a new container if the specified NAME is a fully qualified name.

giuseppe:
Of course I don't mind to answer your questions:

1. that is a great idea, in facts I've added a patch to implement it.

2. the prefix before the colon ("syscontainers:") defines the scope of the patch, I try when possible to keep the first line < 50 characters (sometimes I fail):

https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

tests/integration/test_system_containers_install.sh
```diff
@@ -141,6 +141,9 @@ ${ATOMIC} pull --storage ostree docker:atomic-test-runonce:latest
...
```
peterbaouoft: One quick question, since you modify the container_exec to be only running for images with valid registries. Is atomic-test-runonce qualified tho? (like no network address separated by "/" here right? i.e, i think it is not qualified as image here?) .

If it is not qualified, this test case seems not meaningful to this PR? 

I tested it a bit with copy pasted Decompose. Looks like reg is empty for the test image. https://paste.fedoraproject.org/paste/pAxcE34qZf2fE2t7wM-UTA

Atomic/backends/_ostree.py
```diff
@@ -165,13 +165,10 @@ def delete_container(self, container, force=False):
 
     def run(self, iobject, **kwargs):
         args = kwargs.get('args')
-        try:
-            name = iobject.name
-        except AttributeError:  # iobject isn't populated
-            raise ValueError('Unable to find an image named {} in {}'.format(
-                args.image, args.storage))
+        name = args.image
```
peterbaouoft: Hmm, I am a bit confused about here. We do have atomic run --name where it requires a existing container. I am slightly concerned that user will do something like `atomic run --storage ostree --name firewalld` where firewalld is the container for example. And no args.image is specified, then will there be a problem?

```
if self.name:
             try:
                 be, con_obj = be_utils.get_backend_and_container_obj(self.name)
                 return be.run(con_obj, atomic=self, args=self.args)
             except ValueError:
pass
```

tests/integration/test_system_containers_install.sh
giuseppe: I added fully qualified or having a ':' which cannot be a container name :-)

That means an image having a tag or specify the location such as `docker:`

peterbaouoft: Ah right, k, makes sense. Did not see the ":" check. Thanks

Atomic/backends/_ostree.py
giuseppe: this command was designed for the Docker backend, and containers having labels.  It doesn't fully fit with the way system containers work.  A system container knows the image it was installed from, so we don't need to duplicate this information on the command line.

I think the best way is to raise an error when `--name` is used with the ostree backend.

giuseppe: `atomic run --storage ostree --name firewalld` will fail immediately as you don't specify the image.

giuseppe:
@peterbaouoft I added a patch to raise an error when `--name` is used ⬆️

Atomic/backends/_ostree.py
peterbaouoft: Hmm, sorry, I did not fully understand this implementation. Let me try explaining my POV about this.  I believe once we install the system container, it will be there unless you uninstall it.(and user wants to use it) The usual way we start the system container is via systemctl start container. 

However, some system containers(or in the future) behaves not like that. One example is firewalld, and we can have the following:
1: We install a system container to the system. E.g: firewalld for example
2: We start the service using systemctl start firewalld
3: If we disable the name option, can we still do the following? Like atomic run --name firewalld firewall-cmd xxx? 
(because we want to fully utilize the container once it is installed onto the host right?)

In a sense, those two options (install a container) or run one-shot method are conflicting with each other in this case no? Because, with `atomic run $syscontainer` disabled, then the users who install containers(e.g firewalld) to the host can not access inner commands like firewall-cmd easily.They can only do one-shot option for these system container or run `runc firewalld firewall-cmd` command manually. 

 So my doubt is do we also want to enable the running via "atomic run container" path as well? I am fine with only using one-shot method for this type of system container too, but that would require a broader discussion? (IMO tho). Again, I might misunderstand this part, so feel free to correct me if I was wrong :).

giuseppe: yes, this patch doesn't change what we can currently do.  We are still able to do `atomic run --name firewalld firewall-cmd xxx`.

The argument is considered an image when it is either fully specified or contains a `:`, in that case it cannot be a container as it is not a valid container name.

We surely need to figure out a way to make this clear in the `atomic run` documentation and `--help` output, as we already had this problem before, as the command says IMAGE but we were using it as a container name.

peterbaouoft: >atomic run --name firewalld firewall-cmd xxx

But that will default to docker container only right?( part of my long question above was to ask should we have this working as well)
 the following would no longer work: 
`atomic run --name firewalld --storage ostree firewall-cmd xxx` and I tried the patch a little:

```
[root@localhost atomic]# atomic containers list -a | grep firewalld
   firewalld    firewalld:latest     firewalld  /usr/bin/r 2018-02-16 14:49 inactive   ostree     /usr/bin/r
[root@localhost atomic]# atomic run -n firewalld --storage ostree firewall-cmd status
--name is not supported by this backend
```

> The argument is considered an image when it is either fully specified or contains a :, in that case it cannot be a container as it is not a valid container name.

Yea, I understand that part. Before, I was concerned about not being able to run system container via atomic run --storage ostree $container.

> We surely need to figure out a way to make this clear in the atomic run documentation and --help output, as we already had this problem before, as the command says IMAGE but we were using it as a container name.

Yup, documentation definitely helps ;)

giuseppe:
I'd prefer an error so that won't go unnoticed if it is running in a script.  Thanks for the review!

baude:
LGTM, merging

baude:
@rh-atomic-bot r+

PR #1188, opened 2018-02-13
syscontainers: Refactor do checkout pieces

peterbaouoft:
## Description
Refactor do_checkout pieces. The current PR's commits all work by itself. No functionality impact should be brought by this PR. Will mark this to be WIP tho, as some "missing layer" check is incoming.

## Related Bugzillas
-
-

## Related Issue Numbers
-

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1154,32 +1154,7 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             return options["values"]
 
         try:
-            if rpm_install_content["rpm_file"]:
-                RPMHostInstall.install_rpm(rpm_install_content["rpm_file"])
-            else:
-                for installed_file in rpm_install_content["new_installed_files"]:
-                    util.write_out("Created file {}".format(installed_file))
-
-            self._systemctl_command("daemon-reload")
-            if (tmpfiles_template):
-                self._systemd_tmpfiles("--create", options["tmpfilesout"])
-
-            if options["upgrade_mode"] == SystemContainers.CHECKOUT_MODE_INSTALL:
-                self._systemctl_command("enable", options["name"])
-            elif was_service_active:
-                if options["upgrade_mode"] == SystemContainers.CHECKOUT_MODE_UPGRADE_CONTROLLED:
-                    must_rollback = False
-                    try:
-                        self._systemctl_command("start", options["name"])
-                    except subprocess.CalledProcessError:
-                        must_rollback = True
-                    if must_rollback:
-                        util.write_err("Could not restart {}.  Attempt automatic rollback".format(options["name"]))
-                        self.rollback(options["name"])
-                        self._systemctl_command("start", options["name"])
-                        return {}
-                else:
-                    self._systemctl_command("start", options["name"])
+            self.checkout_container_wrapup(rpm_install_content, tmpfiles_template, options, was_service_active)
         except (subprocess.CalledProcessError, KeyboardInterrupt):
             if rpm_install_content["rpm_installed"]:
                 RPMHostInstall.uninstall_rpm(rpm_install_content["rpm_installed"])
@@ -1189,6 +1164,58 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
             raise
 
         return options["values"]
+
+    def checkout_container_wrapup(self, rpm_install_content, tmpfiles_template, options, was_service_active):
+        """
+        Last stage of checking out a container. It includes outputting results,
+        setup for systemd part, and service starting
+        """
+        # Handle results based on rpm_install_content collected earlier
+        if rpm_install_content["rpm_file"]:
+            RPMHostInstall.install_rpm(rpm_install_content["rpm_file"])
+        else:
+            for installed_file in rpm_install_content["new_installed_files"]:
+                util.write_out("Created file {}".format(installed_file))
+
+        self._systemctl_command("daemon-reload")
+        if (tmpfiles_template):
+            self._systemd_tmpfiles("--create", options["tmpfilesout"])
+
+        # Note, when a rollback for container happens, the values from options will be reset to empty
+        self.checked_out_container_systemd_setup(options, was_service_active)
+
+    def checked_out_container_systemd_setup(self, options, was_service_active):
+        """
+        Handle container service accordingly based on was_service_active and cotnainer name.
+        If the container service was active before, the new container service will be started as well
+        """
+        if options["upgrade_mode"] == SystemContainers.CHECKOUT_MODE_INSTALL:
+            self._systemctl_command("enable", options["name"])
+        elif was_service_active:
+            if options["upgrade_mode"] == SystemContainers.CHECKOUT_MODE_UPGRADE_CONTROLLED:
+                self.start_container_and_rollback_if_necessary(options)
+            else:
+                self._systemctl_command("start", options["name"])
+
+    def start_container_and_rollback_if_necessary(self, options):
+        # Note, here we check if the container needs rollback by starting the
+        # container service. If no rollback is needed, the container service
+        # remains active
+        if(self.is_container_rollback_needed(options["name"])):
+            self.rollback_checked_out_container(options["name"])
+            options["values"] = {}
+
+    def is_container_rollback_needed(self, name):
+        try:
+            self._systemctl_command("start", name)
+        except subprocess.CalledProcessError:
+            return True
+        return False
+
+    def rollback_checked_out_container(self, name):
+        util.write_err("Could not restart {}.  Attempt automatic rollback".format(name))
+        self.rollback(name)
+        self._systemctl_command("start", name)
 
     def _get_preinstalled_containers_path(self):
         return ATOMIC_USR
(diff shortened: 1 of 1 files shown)


peterbaouoft:
Actually, I think the refactoring for missing layer logic in do_checkout, can be done in removing skopeo fallback part in #1180 . I think #1180 can be unblocked now, since we update the centos to be smoketested. And smoketested version already contain skopeo based on yesterday's try. 

So, will lift this WIP, and do the "missing layer refactoring" into #1180. I think this PR should be ready for review

Atomic/syscontainers.py
```diff
@@ -1199,6 +1165,70 @@ def _do_checkout(self, repo, options):
...
+
+    def start_container_and_rollback_if_necessary(self, options):
+        # Note, here we check if the container needs rollback by starting the
+        # container service. If no rollback is needed, the container service
+        # remains active
+        if(self.is_container_rollback_needed(options["name"])):
+            self.rollback_checked_out_container(options["name"])
+            options["values"] = {}
+
+    def is_container_rollback_needed(self, name):
```
giuseppe: I feel it is unnecessary to split to these mini functions.  It makes the code more difficult to read.

Could `is_container_rollback_needed` just be a variable?

Atomic/syscontainers.py
```diff
@@ -1199,6 +1165,70 @@ def _do_checkout(self, repo, options):
...
+        try:
+            self._systemctl_command("start", name)
+        except subprocess.CalledProcessError:
+            return True
+        return False
+
+    def rollback_checked_out_container(self, name):
```
giuseppe: same here, it can be part of the `start_container_and_rollback_if_necessary` function

Atomic/syscontainers.py
```diff
@@ -1045,10 +1045,7 @@ def _do_checkout(self, repo, options):
             raise ValueError("Image {} not found".format(options["img"]))
 
         remote_path = SystemContainers._get_remote_location(options["remote"])
-        if remote_path:
-            exports = os.path.join(remote_path, "rootfs/exports")
-        else:
-            exports = os.path.join(options["destination"], "rootfs/exports")
+        exports = SystemContainers._get_exports_path(remote_path, options["destination"])
```
giuseppe: could we just rewrite it to something like?

```
exports = os.path.join(remote_path or options["destination"], "rootfs/exports")
```

Atomic/syscontainers.py
```diff
@@ -1157,38 +1153,8 @@ def _do_checkout(self, repo, options):
...
-                    except subprocess.CalledProcessError:
-                        must_rollback = True
-                    if must_rollback:
-                        util.write_err("Could not restart {}.  Attempt automatic rollback".format(options["name"]))
-                        self.rollback(options["name"])
-                        self._systemctl_command("start", options["name"])
-                        return {}
-                else:
-                    self._systemctl_command("start", options["name"])
+            self.checkout_container_wrapup(rpm_install_content, tmpfiles_template, options, was_service_active)
```
giuseppe: could we rename it to `finalize_checkout` or something similar?

Atomic/syscontainers.py
```diff
@@ -1199,6 +1165,70 @@ def _do_checkout(self, repo, options):
...
+    def _get_exports_path(remote_path, destination):
+        """
+        Exports path is usually at rootfs/exports, return the exports path
+        based on remote and destination
+        """
+        if remote_path:
+            exports = os.path.join(remote_path, "rootfs/exports")
+        else:
+            exports = os.path.join(destination, "rootfs/exports")
+        return exports
```
giuseppe: since there is only one user of this function, let's simplify it in the caller.

Atomic/syscontainers.py
```diff
@@ -1199,6 +1165,70 @@ def _do_checkout(self, repo, options):
...
+        self._systemctl_command("daemon-reload")
+        if (tmpfiles_template):
+            self._systemd_tmpfiles("--create", options["tmpfilesout"])
+
+        # Note, when a rollback for container happens, the values from options will be reset to empty
+        self.checked_out_container_systemd_setup(options, was_service_active)
+
+    def checked_out_container_systemd_setup(self, options, was_service_active):
+        """
+        Handle container service accordingly based on was_service_active and cotnainer name.
```
giuseppe: can we rename this function to something else?  What about `systemd_service_setup`?

giuseppe: There is a typo in the comment: s|cotnainer|container|

Atomic/syscontainers.py
```diff
@@ -1157,38 +1153,8 @@ def _do_checkout(self, repo, options):
         if options["prefix"]:
             return options["values"]
 
-        sym = os.path.join(self._get_system_checkout_path(), options["name"])
-        if os.path.exists(sym):
-            os.unlink(sym)
-        os.symlink(options["destination"], sym)
-
         try:
```
giuseppe: it looks like this chunk got lost in the new version.  I think we still need to create the symlink

peterbaouoft: ah, this is done in the second commit. I mentioned the reason briefly in the commit msg. Going to mention it here again. The reason for the removal was mainly because of the duplicate calls. 

When prefix is not specified, we are creating a symlink already right? And then one line right below it, we directly return when there is prefix specified. So, the only case where we create a symlink is when prefix is None, which is covered in the following case right?

```
          if not options["prefix"]:
             sym = os.path.join(self._get_system_checkout_path(), options["name"])
             if os.path.exists(sym):
                 os.unlink(sym)
             os.symlink(options["destination"], sym)
 
         # if there is no container service, delete the checked out files.  At this point files copied to the host
         # are already handled.
         if not has_container_service:
             if not remote_path:
                 shutil.rmtree(os.path.join(options["destination"], "rootfs"))
             return options["values"]
 
         if options["prefix"]:
             return options["values"]
```

So, I believe the same creating symlink logic has been called twice originally? That's why I did the removal.

Atomic/syscontainers.py
peterbaouoft: sure

Atomic/syscontainers.py
peterbaouoft: sounds good

Atomic/syscontainers.py
peterbaouoft: sure

peterbaouoft: right, nice catch!

Atomic/syscontainers.py
peterbaouoft: Thanks for the feedback! And ouch, the reason why I tend to create those mini functions was to make each function has only one abstraction level so that each function appears to do only one thing, and to separate concepts with implementation details in some degree.  (May be I did it in the wrong way?). 

One benefit of applying above is that, we can summarize  the functionality of `start_container_and_rollback_if_necessary` in a sentence like the following(with few details), In the function, we first check if container rollback is needed, then if it is, we will rollback the container and set the return value to empty.

But, it might be some what hard for me to get it right for the first few tries( I was still practising with it). So do you mind telling me which part of the code do you find it hard to understand? Ofc, I can put the code all back into one function like before if you want :P.

Atomic/syscontainers.py
giuseppe: True, we are already doing that.

We can drop this part as it is obviously not needed.

Atomic/syscontainers.py
giuseppe: if there are not multiple users of the function and it is easy to follow, I think it is better that we don't break it.

Atomic/syscontainers.py
```diff
@@ -1199,6 +1165,53 @@ def _do_checkout(self, repo, options):
 
         return options["values"]
 
+    def finalize_checkout(self, rpm_install_content, tmpfiles_template, options, was_service_active):
```
giuseppe: looks much better.

Could you prefix the name of these new functions with `_` so it is clear they should be used only within this class?

Atomic/syscontainers.py
```diff
@@ -1199,6 +1165,53 @@ def _do_checkout(self, repo, options):
...
+                util.write_out("Created file {}".format(installed_file))
+
+        self._systemctl_command("daemon-reload")
+        if (tmpfiles_template):
+            self._systemd_tmpfiles("--create", options["tmpfilesout"])
+
+        # Note, when a rollback for container happens, the values from options will be reset to empty
+        self.systemd_service_setup(options, was_service_active)
+
+    def systemd_service_setup(self, options, was_service_active):
```
giuseppe: same here

Atomic/syscontainers.py
```diff
@@ -1199,6 +1165,53 @@ def _do_checkout(self, repo, options):
...
+        """
+        if options["upgrade_mode"] == SystemContainers.CHECKOUT_MODE_INSTALL:
+            self._systemctl_command("enable", options["name"])
+        elif was_service_active:
+            if options["upgrade_mode"] == SystemContainers.CHECKOUT_MODE_UPGRADE_CONTROLLED:
+                self.start_container_and_rollback_if_necessary(options)
+            else:
+                self._systemctl_command("start", options["name"])
+
+    def start_container_and_rollback_if_necessary(self, options):
```
giuseppe: and here

peterbaouoft:
> some minor fixes and I think it is ready to go!

Done in the fixup above, thanks for the review! :-)

giuseppe:
lovely!

@rh-atomic-bot r+ 6bf3409

Issue #1190, opened 2018-02-20
[atomic scan] Get image name in atomic scanner


navidshaikh:
I am writing an atomic scanner for a use case where the scanner needs to know the `image_name` being processed. 

Now, I get the image name by explicitly setting an env variable while invoking the `atomic scan` command like
```
IMAGE_NAME=<image_name> atomic scan [..]
```
what's the better way to get `image_name` without explicitly have to ask for providing env var along with atomic scan command?

rhatdan:
Currently I don't believe we leak that information into the container?
@baude PTAL

navidshaikh:
There is (meta)data with the image itself which could help atomic scanners to implement different use cases. 
For example, accessing `Labels` of image under-scan, within atomic scanner for processing.

How I am achieving that right now is
 - Provide env var specifying image name while invoking `atomic scan` and `os.environ.get` within scanner
 -  Using `Atomic.run.Run().get_label()` to access the Labels.

If we can provide an interface for atomic scanners (say via `Atomic`) to retrieve metadata related information for image under scan, that should be helpful.

baude:
@navidshaikh do you want only on the image name ?

baude:
@navidshaikh and you want all this before the scan is run correct?

navidshaikh:
> do you want only on the image name?
 
1. Image name
2. The dictionary/hash printed by `docker inspect` (only interested in the Labels part right now, but other information can be useful as well)

Labels part can be achieved, if image name can be retrieved.
Like:
```
from Atomic import run
run_object = run.Run()
run_object.image = IMAGE_NAME
label = run_object.get_label("LABEL_NAME")
```

>  and you want all this before the scan is run correct?

Correct. Basically, a way to access that information in the atomic scanner itself.

For eg: My use case is, write an atomic scanner which reads `Labels` of image under test and perform foo, bar actions based on data in `Labels`.

PR #1194, opened 2018-02-21
Provide scanner with image name and info

baude:
Issue #1190 asks for an enhancement to atomic scan where
the scanner could figure out the image|container inspect
information prior to the scan.  We do this now by writing
a file in the scanout/ dir.  The file is JSON formatted
and could be ingested by a scanner that needed that type
of information.  The path should bexi
/scanout/inspect_info.json.

Signed-off-by: baude <bbaude@redhat.com>

## Related Bugzillas
-
-

## Related Issue Numbers
-
-

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/scan.py
+++ b/Atomic/scan.py
@@ -154,6 +154,9 @@ class Scan(Atomic):
         # Create the output directory
         os.makedirs(self.results_dir)
 
+        # Record target information
+        self.record_inspect_info()
+
         docker_args = ['docker', 'run', '-t', '--rm', '-v', '/etc/localtime:/etc/localtime',
                        '-v', '{}:{}'.format(self.chroot_dir, '/scanin'), '-v',
                        '{}:{}:rw,Z'.format(self.results_dir, '/scanout')]
@@ -536,3 +539,18 @@ class Scan(Atomic):
 
     def remediate(self, script, iid, results_dir):
         util.check_call([sys.executable, script, '--id', iid, '--results_dir', results_dir])
+
+
+    def record_inspect_info(self):
+        """
+        Writes inspect information for each object passed to the scanner and
+        stores them in results_dir/inspect_info.json
+        :return: None
+        """
+
+        inspect = []
+        for scan_object in self.scan_list:
+            inspect.append(scan_object.config)
+
+        with open(os.path.join(self.results_dir, 'inspect_info.json'), 'w') as f:
+            json.dump(inspect, f, indent=4, separators=(',', ': '))


baude:
do not merge until @navidshaikh confirms this works for him as he requested the enhancement.

TomSweeneyRedHat:
if @navidshaikh is happy with it, LGTM

Atomic/scan.py
```diff
@@ -536,3 +539,18 @@ def write_persistent_data(self):
...
+    def record_inspect_info(self):
+        """
+        Writes inspect information for each object passed to the scanner and
+        stores them in results_dir/inspect_info.json
+        :return: None
+        """
+
+        inspect = []
+        for scan_object in self.scan_list:
+            inspect.append(scan_object.config)
```
navidshaikh: I think a map/dictionary is better suited here.
Like
```
{
 scan_object.name: scan_object.config
}
```

While most of the time the atomic scanner execute per image, there are options to scan `all(images, containers)`, `all_images`, `all_containers` as well. With this approach, all the scan_object configs are going to be appended, and it can become complex to map which config belong with which scan_target. 

The approach for map/dictionary for `inspect_info.json` should help in both cases where a single scan_target or multiple are provided.

For single scan_target, using the map/dict will provide scanner with
 - Image identity (name, as provided in command line)
 - Its config / inspect info

For multiple scan_target,  using the map/dict will provide scanner with
 - Image identity (name, as provided in command line)
 - Its config / inspect info
 - And mapping for `scan_target:scan_target_config`

baude: @navidshaikh names can be a PITA if the image has multiple names.  how about IDs?

navidshaikh: @baude : IDs sounds good. With `scan_object.name` I meant the identity of the scan_target.

navidshaikh: @baude : Thanks for putting this together. 

Regarding `name` of scan_targets I also took a look in the backendutils module, however the scan targets are then transformed into objects. 

```
231                         _, scan_obj = beu.get_backend_and_image_obj(scan_target)
```

I was wondering if we could shadow the scan_target provided on the command line in the inspect output file.

Given the number of options atomic scan could work with, it can become complex as, if provided `--images` or say `--all`, the best choice is to have `ID`.

navidshaikh: How about having the `SCAN_TARGET` env variable in the atomic scanner, with values based on populated `scan_target` variable?

WDYT?

I'll be happy to collaborate for the PR, if it looks good.

Issue #1191, opened 2018-02-20
'atomic images list' can break if system container is missing files/partially installed


miabbott:
In the error situation covered by [RHBZ#1542144](https://bugzilla.redhat.com/show_bug.cgi?id=1542144), it is possible to end up with a system container that is only partially installed.

In this partial-install state, the `atomic images list` functionality breaks and is unable to show any of the images on the host.

```
# rpm -q atomic
atomic-1.21.1-1.git1170769.el7.x86_64

# atomic images list
[Errno 2] No such file or directory: '/var/lib/containers/atomic/etcd.0/info'

# atomic --debug images list
Namespace(_class=<class 'Atomic.images.Images'>, all=False, assumeyes=False, debug=True, filter=None, func='display_all_image_info', heading=True, ignore=False, json=False, profile=False, quiet=False, truncate=True)
Backends(docker: Active, ostree: Active, )

[Errno 2] No such file or directory: '/var/lib/containers/atomic/etcd.0/info'
Traceback (most recent call last):
  File "/bin/atomic", line 185, in <module>
    sys.exit(_func())
  File "/usr/lib/python2.7/site-packages/Atomic/images.py", line 159, in display_all_image_info
    _images = self._get_images()
  File "/usr/lib/python2.7/site-packages/Atomic/images.py", line 219, in _get_images
    self._mark_used(_images)
  File "/usr/lib/python2.7/site-packages/Atomic/images.py", line 308, in _mark_used
    all_containers = [x.image for x in self.be_utils.get_containers()]
  File "/usr/lib/python2.7/site-packages/Atomic/backendutils.py", line 157, in get_containers
    con_objs += be.get_containers()
  File "/usr/lib/python2.7/site-packages/Atomic/backends/_ostree.py", line 110, in get_containers
    return [self._make_container(x) for x in self.syscontainers.get_containers()]
  File "/usr/lib/python2.7/site-packages/Atomic/syscontainers.py", line 1508, in get_containers
    return self._get_containers_at(checkouts, False, containers) + self._get_containers_at(preinstalled, True, containers)
  File "/usr/lib/python2.7/site-packages/Atomic/syscontainers.py", line 1472, in _get_containers_at
    with open(os.path.join(fullpath, "info"), "r") as info_file:
IOError: [Errno 2] No such file or directory: '/var/lib/containers/atomic/etcd.0/info'
```

This can be reproduced more simply by just installing a system container and then removing the `info` file from the directory:

```
[...]

PR #1195, opened 2018-02-21
WIP:syscontainer: catch missing container info failure

peterbaouoft:
As reported by https://github.com/projectatomic/atomic/issues/1191,
when info file is missing, atomic images list will unexpectedly fail.

This patch is therefore to catch the error and instead output a
warning to notify user. In addition, the container itself will be ignored.

Not sure if we should color/bold the warning message, so I leave it as it is similar to other warning messages. But, I am open to suggestions. :-)

The outputted atomic images list when one of the container info file is missing is like the following:
```
[root@localhost atomic]# ./atomic images list
Container at /var/lib/containers/atomic/hello-world is missing its info file. Please consider reinstalling the container

   REPOSITORY                                            TAG      IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
>  registry.fedoraproject.org/f27/kubernetes-scheduler   latest   cb17034be1ed   2018-02-07 17:48   231.12 MB      ostree    
   hello-world                                           latest   8072a54ebb3b   2018-02-09 19:17   985.0 B        ostree
[root@localhost atomic]# ls /var/lib/containers/atomic/hello-world | grep info
[root@localhost atomic]# 
```
and atomic containers list
```
[root@localhost atomic]# ./atomic containers list
Container at /var/lib/containers/atomic/hello-world is missing its info file. Please consider reinstalling the container

   CONTAINER ID IMAGE                NAME       COMMAND    CREATED          STATE      BACKEND    RUNTIME   
   kubernetes-s registry.fedoraproje kubernetes /usr/bin/k 2018-02-07 17:48 running    ostree     /usr/bin/r
```

## Related Issue Numbers

- https://github.com/projectatomic/atomic/issues/1191
- https://github.com/projectatomic/atomic/issues/1199

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -291,13 +291,20 @@ class SystemContainers(object):
         :param name: name for the container
         :returns: info object, rpm_installed field value, installed_files_checksum field value
         """
-        with open(os.path.join(checkout, name, "info"), "r") as info_file:
-            info = json.loads(info_file.read())
-            rpm_installed = info["rpm-installed"] if "rpm-installed" in info else None
-            installed_files_checksum = info["installed-files-checksum"] if "installed-files-checksum" in info else None
-            if installed_files_checksum is None:
-                installed_files = info["installed-files"] if "installed-files" in info else None
-                installed_files_checksum = {k : "" for k in installed_files}
+        rpm_installed = None
+        installed_files_checksum = None
+        info = {}
+        container_path = os.path.join(checkout, name)
+        try:
+            with open(os.path.join(container_path, "info"), "r") as info_file:
+                info = json.loads(info_file.read())
+                rpm_installed = info["rpm-installed"] if "rpm-installed" in info else None
+                installed_files_checksum = info["installed-files-checksum"] if "installed-files-checksum" in info else None
+                if installed_files_checksum is None:
+                    installed_files = info["installed-files"] if "installed-files" in info else None
+                    installed_files_checksum = {k : "" for k in installed_files}
+        except IOError:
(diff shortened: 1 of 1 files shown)


peterbaouoft:
Hmm, the `atomic uninstall $container` should be properly handled once https://github.com/projectatomic/atomic/issues/1199 is solved. However, `atomic containers delete $container` require a few more tweaks.

Because the way how we handled `atomic containers delete` involves 1: checking container to be existant  or not 2: if it exists, then we go to the corresponding backend to perform the deletion. That would not cause any problem before, but after adopting this patch, the containers without any proper info file will be skipped by `get_containers_at`. IOW, the container will be considered "non-existant". Therefore, when delete the container, the user will not be able to do so, because we think the container is never existent in the first place.

Hmm, if we want to avoid the above scenario mentioned, there are two ways that I am thinking that might work. 
1: Introduce a new method for `get_valid_containers` for `atomic images list` or `atomic containers list`. It will only get the containers which contain info files, and output a warning. Then for other container operations (e.g delete in this case), we will use the old method, but making the "broken container" raising an error.  Then in `atomic containers delete` , we can catch the error and allow deletion to happen properly

2: We can use the similar trick as the `atomic uninstall container` for the deletion case. That case tho will skip the checking existence of container mentioned earlier.

**NOTE:** When info file is not there, the other installed files on the host will no longer be tracked. The only known files are the container folders to be deleteable. So, I am not sure how much attention we  should put in handling this error corner case. Any suggestions on this? @ giuseppe

peterbaouoft:
Close this for now, might come back later at this when there is a response in the linked issue :p

PR #1196, opened 2018-02-22
run: add option --runtime

giuseppe:
it allows to select a different OCI runtime to use for "atomic run --storage ostree"


rhatdan:
LGTM

giuseppe:
Thanks!

@rh-atomic-bot r=rhatdan

Issue #1193, opened 2018-02-21
Minor question about documentation


peterbaouoft:
In https://github.com/projectatomic/atomic/blob/master/docs/atomic-pull.1.md, we have the following documents. 
```
An 'ostree' image refers to an image which is fetched from a remote OSTree repository. The remote has to be already configured in the local OSTree repository:

atomic pull --storage ostree ostree:REMOTE/branch
```

But the code in syscontainers.py (shown below) seems otherwise. 
```
 def _pull_image_to_ostree(self, repo, image, upgrade, src_creds=None):
        if not repo:
            raise ValueError("Cannot find a configured OSTree repo")
        if image.startswith("ostree:") and image.count(':') > 1:
            self._check_system_ostree_image(repo, image, upgrade)
```

From tutorialspoint website https://www.tutorialspoint.com/python/string_count.htm, description of count
> The method count() returns the number of occurrences of substring sub in the range [start, end]

 In code, for "ostree image" we are expecting input to have at least 2 colons right? But in the doc, we said we are expecting a remote/branch? So, there seems to be a conflict here? Or I am misinterpreting it? Thanks in advance :-)!

giuseppe:
I think this must be droppoed. We added `ostree:` for a way to pull images directly from another `ostree` repository but we never used it.  Today `ostree:` has a different meaning with what we are doing with Skopeo.  I've opened a PR here: https://github.com/projectatomic/atomic/pull/1197

PR #1197, opened 2018-02-26
syscontainers: drop pull from a remote ostree

giuseppe:
this code was added for pulling directly from an OSTree repository.  It
wasn't never really used and it expects the image to be fully exploded
in an ostree commit without supporting OCI layers.

Dropping this part also avoid confusion with what we do with Skopeo
where "ostree:" is supported both as source and destination.

Closes: https://github.com/projectatomic/atomic/issues/1193

### DIFF
--- a/Atomic/install.py
+++ b/Atomic/install.py
@@ -182,7 +182,7 @@ class Install(Atomic):
 
     @staticmethod
     def ostree_uri(image_name):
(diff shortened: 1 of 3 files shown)


rhatdan:
LGTM

TomSweeneyRedHat:
LGTM

giuseppe:
thanks!

@rh-atomic-bot r=rhatdan

PR #1200, opened 2018-02-28
images: add cache for dangling images

giuseppe:
do not reload the list of all the images every time we check if an image
is dangling.  This makes "images list" extremely slow when used on the
Docker backend.

On my system (with ~100 images), this patch brings down the wall clock
from ~10 seconds to ~0.5 seconds:

$ /usr/bin/time -v ./atomic images list -a 2>&1 | grep "wall clock"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:09.84

to:

$ /usr/bin/time -v ./atomic images list -a 2>&1 | grep "wall clock"
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:00.58

### DIFF
--- a/Atomic/backends/_containers_storage.py
+++ b/Atomic/backends/_containers_storage.py
@@ -224,7 +224,7 @@ class ContainersStorageBackend(Backend): #pylint: disable=metaclass-assignment
             image_objects.append(self._make_image(image['id'], image))
         return image_objects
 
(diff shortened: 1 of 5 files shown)


giuseppe:
tests are passing, fine to merge?

rhatdan:
@rh-atomic-bot r+

PR #1201, opened 2018-02-28
Begin changing atomic to work with alternatives to docker CLI

rhatdan:
This patch should allow systems with podman installed and no
docker CLI to be able to run.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

### DIFF
--- a/Atomic/Export.py
+++ b/Atomic/Export.py
@@ -27,7 +27,7 @@ def export_docker(graph, export_location, force):
                     raise ValueError("Please delete dangling images before running atomic storage export")
 
             util.write_out("Deleting dangling images")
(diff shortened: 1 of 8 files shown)


rhatdan:
@baude @mheon PTAL

rhatdan:
Actually @baude I think it would be better if you take this over, since it requires a podman backend.

baude:
ok, im just waiting for us to give birth to the backend and then i will start enablement.

rhatdan:
@baude So you want to tie this into the varlink stuff?  I was thinking we would just exec the commands for now.

giuseppe:
not related to this, but another thing we should take care of is the `README.md` file, as it is quite docker-centric :-)

/cc @cgwalters

Issue #1199, opened 2018-02-27
atomic uninstall can not recognize container name properly


peterbaouoft:
Hi I am recently playing around with atomic uninstall command. I noticed this interesting behaviour. Looks like container name is not rendered properly with atomic uninstall? 
 
```
[root@localhost ~]# atomic images list
   REPOSITORY              TAG      IMAGE ID       CREATED            VIRTUAL SIZE   TYPE      
   docker.io/hello-world   latest   f2a91732366c   2017-11-21 00:23   1.85 KB        docker    
   docker.io/hello-world   latest   8072a54ebb3b   2018-02-27 21:23   985.0 B        ostree    
[root@localhost ~]# atomic install --system --system-package=no docker.io/hello-world 
Extracting to /var/lib/containers/atomic/hello-world.0
Warning: /exports directory not found.  Default config files will be generated.
Warning: You may want to modify `/var/lib/containers/atomic/hello-world.0/config.json` before starting the service
systemctl daemon-reload
systemctl enable hello-world
[root@localhost ~]# atomic containers list                                         
   CONTAINER ID IMAGE                NAME       COMMAND    CREATED          STATE      BACKEND    RUNTIME   
   hello-world  docker.io/hello-worl hello-worl run.sh     2018-02-27 21:23 failed     ostree     /bin/runc
[root@localhost ~]# atomic uninstall --storage ostree docker.io/hello-world --name hello-world
[root@localhost ~]# ls /var/lib/containers/atomic/hello-world/     
config.json  hello-world.service  info  rootfs
```

In `_ostree` backend, it also looks like the container name did not get rendered?  WDYT
```
    def uninstall(self, iobject, name=None, **kwargs):
        return self.syscontainers.uninstall(iobject.name)
```

This is produced on Atomic Host:
```
[root@localhost ~]# rpm-ostree status
State: idle
Deployments:
* fedora-atomic:fedora/27/x86_64/atomic-host
                   Version: 27.60 (2018-01-16 16:35:15)
                BaseCommit: 972e5a8158b610fec80f3f73f3372b7bea2b841038f2e246aa7623dbf5b5a751
         PendingBaseCommit: 772ab185b0752b0d6bc8b2096d08955660d80ed95579e13e136e6a54e3559ca9
        PendingBaseVersion: 27.61 (2018-01-17 15:52:47)
              GPGSignature: Valid signature by 860E19B0AFA800A1751881A6F55E7430F5282EE4
           LayeredPackages: man

[...]

PR #1202, opened 2018-02-28
syscontainer: let uninstall fail when name specified

peterbaouoft:
Based on discussion https://github.com/projectatomic/atomic/issues/1199#issuecomment-369153184,
we want to disallow user to specify --name option for uninstalling system containers.

Instead, we should encourage user to delete container using atomic
containers delete

## Related Bugzillas
-
-

## Related Issue Numbers
- https://github.com/projectatomic/atomic/issues/1199
-

## Pull Request Checklist:

- [ ] Unittests
- [ ] Integration Tests

### DIFF
--- a/Atomic/backends/_ostree.py
+++ b/Atomic/backends/_ostree.py
@@ -128,6 +128,8 @@ class OSTreeBackend(Backend):
         return self.syscontainers.install(image, name)
 
     def uninstall(self, iobject, name=None, **kwargs):
+        if name is not None:
+            raise ValueError("System containers do not support --name. Please use atomic uninstall NAME")
         return self.syscontainers.uninstall(iobject.name)
 
     def prune(self):


Atomic/backends/_ostree.py
```diff
@@ -128,6 +128,8 @@ def install(self, image, name, **kwargs):
         return self.syscontainers.install(image, name)
 
     def uninstall(self, iobject, name=None, **kwargs):
+        if name is not None:
+            raise ValueError("System Containers uninstall do not support deletion using container name, Please consider trying atomic containers delete command")
```
giuseppe: What about "System containers do not support--name.  Please use atomic uninstall NAME" ?

giuseppe:
@rh-atomic-bot r+ e7ddcb5

PR #1205, opened 2018-03-08
syscontainers: fix parsing image with dot in version

eyusupov:
## Description
The old code was parsing postgres:9.6 as (postgres:9.6, postgres, 9:6).
This issue is now fixed and method is covered with tests.

## Related Bugzillas
n/a

## Related Issue Numbers
n/a

## Pull Request Checklist:

- [x] Unittests
- [ ] Integration Tests


eyusupov:
Commands to reproduce:
```
atomic pull postgres/postgres:9.6
atomic install --system postgres:9.6
```

peterbaouoft:
LGTM :).

eyusupov:
@peterbaouoft looking at other PR's it looks like it needs to get approved by someone using via `rh-atomic-bot`. Can you please do that?

peterbaouoft:
@eyusupov, ah sorry, I am not one of the maintainers of the repo. let me ping someone who has access rights =). @ashcrow  mind taking a look at this PR as well?

ashcrow:
Thanks @eyusupov!

eyusupov:
Good catch, will update.

Atomic/syscontainers.py
```diff
@@ -2017,18 +2017,18 @@ def get_default_system_name(image):
...
-        reg, image = imagename[:sep], imagename[sep + 1:]
-        if '.' not in reg:
-            # if the registry doesn't look like a domain, consider it as the
-            # image prefix
-            reg = ""
-            image = imagename
-        sep = image.find(":")
-        if sep > 0:
-            return reg, image[:sep], image[sep + 1:]
+        image, _, tag = imagename.partition(":")
```
peterbaouoft: >  but I'm seeing a different result when port number is in the registry

ah right, did not know that will be a problem.  (thanks @ashcrow  for pointing it out and /me took down some notes).  I believe this should be the culprit here, this only takes the leftmost colon, but the tag is usually separated by the rightmost colon.

peterbaouoft: > Good catch, will update

oops, looks like I was a bit late, you can ignore this and the previous comment ;).

eyusupov:
@rh-atomic-bot bot, retest this please

eyusupov:
Squashed the fixup commit

Atomic/syscontainers.py
```diff
@@ -2017,18 +2017,18 @@ def get_default_system_name(image):
...
+            reg, _, prefix = prefix.partition("/")
         else:
-            return reg, image, "latest"
+            reg = ""
+        image, _, tag = image.partition(":")
+        if prefix:
+            image = "%s/%s" % (prefix, image)
+        if not tag:
+            tag = "latest"
+        return reg, image, tag
```
giuseppe: This was some old code that we forgot to refactor to `util.Decompose` when it was  introduced.  Do you think we can just use it here as well or does it do something different ?  Do your tests still pass?

eyusupov: I updated `_parse_imagename` to use `util.Decompose`. Maybe it's worth to remove `_parse_imagename` altogether and add a getter to Decompose returning a repo with an image?

eyusupov: I think there might be some unexpected effects because of `util.Decompose` setting `repo` to `library` if it's empty for docker.io images.

eyusupov:
@giuseppe I pushed two new commits. In the second one I also tried to avoid a change in behavior in case of `library` repo being added implicitly for docker.io by `util.Compose`.
I'm not entirely sure if it's needed - my guess is that something might break after atomic update for an image installed by previous atomic version, but I'm not familiar with the code enough yet to be sure.

eyusupov:
The current behavior is also questionable though - would I get two different ostree refs if I ran some atomic with explicitly specifying `docker.io/library/image` vs 'docker.io/image'?

ashcrow:
Quick ipython check:

```
In [4]: d = util.Decompose('docker.io/library/image')
                                                     
In [5]: d.all                                                                                             
Out[5]: ('docker.io', 'library', 'image', 'latest', '')
                                                                                                          
In [6]: d = util.Decompose('docker.io/image')

In [7]: d.all
Out[7]: ('docker.io', 'library', 'image', 'latest', '')

In [8]: d = util.Decompose('docker.io:5000/library/image:latest')

In [9]: d.all
Out[9]: ('docker.io:5000', 'library', 'image', 'latest', '')

In [10]: d = util.Decompose('docker.io:5000/library/image:v3')

In [11]: d.all
Out[11]: ('docker.io:5000', 'library', 'image', 'v3', '')

In [12]: d = util.Decompose('docker.io:5000/library/image:v3.3')

In [13]: d.all
Out[13]: ('docker.io:5000', 'library', 'image', 'v3.3', '')

```

giuseppe:
> The current behavior is also questionable though - would I get two different ostree refs if I ran some atomic with explicitly specifying docker.io/library/image vs 'docker.io/image'?

Yes, with the current code you will end up with two different ostree refs.  I think it is fine to fix it now (to my knowledge there are no system containers under `docker.io/library`). There should not be problems if we automatically convert the image name to include "/library" when missing and the "update" code should already work with this change.

eyusupov:
@rh-atomic-bot bot, retest this please

eyusupov:
For some reason (probably because of squashing/rebasing) I can't rerun CI tests, can you trigger them somehow?

giuseppe:
could you please rebase this PR on top of the current master?

Once you rebase, there is also another call to `_parse_imagename` that should be taken care of, I needed to amend this patch to get it working here:

```
diff --git a/Atomic/syscontainers.py b/Atomic/syscontainers.py
index b85b3fa..72cb94c 100644
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -1521,11 +1521,11 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
                     config = json.load(config_file)
                     command = u' '.join(config["process"]["args"])
 
-            registry, image, tag = SystemContainers._parse_imagename(image)
-            if registry:
-                image = "%s/%s:%s" % (registry, image, tag)
+            d = util.Decompose(image)
+            if d.registry != "":
+                image = "%s/%s:%s" % (d.registry, d.image, d.tag)
             else:
-                image = "%s:%s" % (image, tag)
+                image = "%s:%s" % (d.image, d.tag)
 
             container = {'Image' : image, 'ImageID' : revision, 'Id' : x, 'Created' : created, 'Names' : [x],
                          'Command' : command, 'Type' : 'system', 'Runtime' : runtime, "Preinstalled" : are_preinstalled}
@@ -2298,10 +2298,10 @@ Warning: You may want to modify `%s` before starting the service""" % os.path.jo
         if name.startswith("oci:"):
             name = name[len("oci:"):]
         d = util.Decompose(name)
-        if d.registry:
-            fullname = "%s/%s:%s" % (d.registry, d.image_with_repo, d.tag)
+        if d.registry != "":
+            fullname = "{}/{}:{}".format(d.registry, d.image, d.tag)
         else:
-            fullname = "%s:%s" % (d.image_with_repo, d.tag)
+            fullname = "{}:{}".format(d.image, d.tag)
 
         ret = "".join([convert(i) for i in fullname])
         return ret
```

The changes in `util.py` can be reverted at this point.

If you are fine with it, could we move the `/library` handling to a followup PR (if you are interested)?  It requires more tweaks, as I've tried to install `busybox` and `docker.io/busybox` and they both failed.

eyusupov:
Oh, looks like I've accidentally rebased my PR onto the old master that I had. Rebasing it back. I'm fine with moving `/library` handling to followup PR, but if we apply your patch, won't the prefix be ignored completely?

giuseppe:
hm.. I've tried with a few images and it seems to work as it used to.  What prefix is ignored?

eyusupov:
`_parse_image_name` used to include repo into the image (like `postgres/postgres`) and `util.Decompose` puts it into separate `repo` property. So in `_parse_image_name` the `image` will be`postgres/postgres` and in `util.Decompose` it will be `postgres`. That's the reason for me adding `_image_with_repo` in `util.Decompose`. 

By the way, in the current version I removed additional code that was handling implicit `library` repo, i.e. the behavior is different now than what it is on master. By moving `/library` handling to follow up PR do you mean I should bring back that code?

giuseppe:
no sorry, it was a mistake on my end.  We need the change in util as well.

The CI seems down:

bot, retest this please

eyusupov:
I think the CI got confused by the rebase to old branch that I did by mistake sometime ago. I can try recreating a PR if there's no other way to make it work.

giuseppe:
we could try that and see if it works.  I didn't see any reaction to my retest request :-(

eyusupov:
https://github.com/projectatomic/atomic/pull/1206

PR #1218, opened 2018-03-29
containers: fix feedback when deleting containers

miabbott:
The feedback when deleting containers was asking the user if they
wanted to delete images, instead of containers.  This updates the
feedback to use the correct terminology.


rhatdan:
LGTM
@rh-atomic-bot r+

PR #1226, opened 2018-04-15
atomic: use runc by default with --user

giuseppe:
## Description

Now that runc has support for rootless containers, let's switch to using it by default.  A newer version of runc that is currently present on Fedora 27 is needed.

Move into atomic some of the transformations that were previously made by bwrap-oci/bubblewrap to adapt the config.json file to run as a non privileged user.  For example, force the `user` namespace to be specified and that a valid uid/gid mapping is provided.

More changes might be needed in future, but for now, this is enough to get the etcd system container to run as a non privileged user using runc.


Atomic/syscontainers.py
```diff
@@ -2779,6 +2785,89 @@ def _correct_dir_permissions_for_user(path):
...
+
+            if not i['destination'].startswith('/dev'):
+                if 'nodev' not in i['options']:
+                    i['options'].append('nodev')
+                if 'nosuid' not in i['options']:
+                    i['options'].append('nosuid')
+            new_mounts.append(i)
+
+        if drop_sys:
+            sys_mount = {
```
peterbaouoft: Minor and question: below is the default generated in runc spec, would the type of mount matter this case?
 ```
"destination": "/sys",
                        "type": "none",
                        "source": "/sys",
                        "options": [
                                "rbind",
                                "nosuid",
                                "noexec",
                                "nodev",
                                "ro"
                        ]
```

File: Atomic/syscontainers.py [hunk 2]
```diff
@@ -2779,6 +2785,89 @@ def _correct_dir_permissions_for_user(path):
...
+                            }
+                            r.append(extra_ids)
+
+            except IOError as e:
+                if e.errno != errno.ENOENT:
+                    raise
+            return r
+
+        if 'uidMappings' not in config['linux']:
+            config['linux']['uidMappings'] = get_subid(os.getuid(), "/etc/subuid")
```
peterbaouoft: Question: I guess we add an extra uid/gid mapping here because user might be  also recognized as subordinate uid? I am not too familiar with the concept of subordinate uid here, but do you think it would be nice to add a comment here to explain a bit? :-)

Atomic/syscontainers.py
```diff
@@ -2779,6 +2785,89 @@ def _correct_dir_permissions_for_user(path):
...
+        drop_sys = len(all_namespaces - namespaces) > 0
+
+        new_mounts = []
+        for i in config['mounts']:
+            if drop_sys and i['destination'].startswith("/sys"):
+                continue
+            if 'options' not in i:
+                i['options'] = []
+
+            if not i['destination'].startswith('/dev'):
```
peterbaouoft: How about `/proc` mount? the default rootless sepc seems not having options like that.
```
{
                        "destination": "/proc",
                        "type": "proc",
                        "source": "proc"
                },
```

Atomic/syscontainers.py
giuseppe: yes, if not all the namespaces are used then mounting sysfs fails.  This is the check I've added and in that case we bind mount it from the host.

Atomic/syscontainers.py
giuseppe: the current user is mapped to root in the container

Atomic/syscontainers.py
giuseppe: for `/proc` it doesn't make any difference, as there are no devices under `/proc`.

giuseppe:
yes, I think the easiest way is to compare the output of `runc spec` and `runc spec --rootless`.  runc is very strict and it errors out immediately if some configuration is not allowed, so we must be a bit more careful that with bwrap-oci (that in any case was implementing only a subset of OCI).

I cannot add a test case yet as it requires a newer runc that what is currently available on Fedora 27, I have tested with a new runc that the etcd system container can still run without any modification to its configuration.

Atomic/syscontainers.py
peterbaouoft: right, I think I see that part, but do you mind elaborating a bit about what will happen for the second uid/gid mapping, which is related to the subuid file we input here, that you added conditionally? :-)

Atomic/syscontainers.py
peterbaouoft: Makes sense, but my only concern is that will overwrite the default (the one without network in its namespaces) `/sys` if there is one.

For instance, if we input default configuration generated by `runc spec --rootless`, the mount type will be overwritten to "bind", would that matter?

Atomic/syscontainers.py
giuseppe: sure, I try to answer to that :)  runc supports more than an user mapped in the usernamespace, and for that it uses the `newuid/newgid` suid tools for setting it up.  We read the configuration file for these two tools and if a mapping for the current is present, then we add it to the configuration file so that runc can request `newuid/newgid` to set it up.

The same  thing is done by bwrap-oci internally, so we never exposed it into atomic:

https://github.com/projectatomic/bwrap-oci/blob/master/run.c#L730

Atomic/syscontainers.py
giuseppe: The default configuration given by `runc spec --rootless` is a bind mount as the one we are adding here

peterbaouoft: wot, ahh, I guess it must be version difference then, my runc version is `runc.x86_64 2:1.0.0-19.rc5.git4bb1fe4.fc27`, and that gives me none for mounting type of `/sys`. SGTM then, thanks for the explanation. :)

giuseppe: the type=none fallbacks to a bind mount in runc :-)  but the correct type is `bind`

Atomic/syscontainers.py
peterbaouoft: ah gotcha, makes sense.

giuseppe:
@rhatdan any comments on this PR?

Atomic/syscontainers.py
rhatdan: Can you do --net=none?

rhatdan:
LGTM
But I have a few questions.

What happens if the user is not listed in /etc/subuid and /etc/subgid?  Do I get a decent error message?

Atomic/syscontainers.py
giuseppe: yes we could do that, that means have a network namespace but not initialize it.  I think that would not be very useful by default, and initialize the network for the rootless case would still require a `setuid` program.  So I preferred to use the host network namespace by default and if the user wishes the config.json file can be edited as we don't offer all the options from the atomic CLI to do it.

giuseppe:
the second part was quite straightforward to implement: https://github.com/projectatomic/atomic/pull/1231

rhatdan:
I thought if a user does not have an entry in the /etc/subuid and /etc/subgid file, he is not allowed ot use userns?

giuseppe:
no, the user is still allowed, but cannot map more than one user from the host:
```
$ id -nu
gscrivano
$ grep -c gscrivano /etc/subuid
0
$ unshare -r id
uid=0(root) gid=0(root) groups=0(root) context=unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023

```

rhatdan:
@rh-atomic-bot r+

PR #1227, opened 2018-04-16
.papr.yaml: use Fedora 27

giuseppe:

## Description

use Fedora 27 for the tests instead of Fedora 26


baude:
bot, retest this please

.papr.sh
```diff
@@ -6,10 +6,6 @@ if test -z "${INSIDE_CONTAINER:-}"; then
 
     if [ -f /run/ostree-booted ]; then
 
-        # by default, the root LV on AH is only 3G, but we need a
-        # bit more for our tests
-        lvresize -r -L +5G atomicos/root
```
jlebon: I think we still need this for `centos/7/atomic/smoketested` for now at least. I suspect this is waiting on inheriting the docker-storage-setup change to `overlay` by default. But I'll look into it.

giuseppe: centos/7/atomic seems to be passing without this change.  Only fedora/27/atomic is failing at the moment (on `test_storage.sh`)

giuseppe:
@jlebon this is the error I get on fedora/27/atomic if I leave that code:
```
### Mon Apr 16 16:54:42 UTC 2018
### https://github.com/projectatomic/atomic/pull/1227
### 0d2e72353e7620bd4ddf5f41221af9e9b0337ec7
### TESTSUITE fedora/27/atomic
### BUILD_ID 12278
>>> ./.papr.sh
+ NO_TEST=
+ test -z ''
+ '[' -f /run/ostree-booted ']'
+ lvresize -r -L +5G atomicos/root
  Insufficient free space: 1280 extents needed, but only 0 available
### EXITED WITH CODE 5 AFTER 0s
```

jlebon:
Yeah, we definitely need to remove it for F27 since it's no longer needed with the "one big LV" change. That change hasn't yet made it to CentOS though. If it's passing, then I guess tests no longer need as much space as before? In that case, we might be able to just completely remove it rather than making it conditional on CentOS.

giuseppe:
some more changes, but the CI is happy now

rhatdan:
LGTM

tests/integration/test_storage.sh
```diff
@@ -35,9 +35,11 @@ setup () {
     # Perform setup routines here.
     copy /etc/sysconfig/docker-storage-setup /etc/sysconfig/docker-storage-setup.atomic-tests-backup
     TEST_DEV_1=/dev/vdb
+    findmnt -o SOURCE
     if findmnt -o SOURCE | grep "^$TEST_DEV_1"; then
-        findmnt -o SOURCE | grep "^$TEST_DEV_1" | uniq | xargs umount
+        exit 77
```
jlebon: Can you expand on this? For some reason, on Fedora, `/dev/vdb1` gets auto-mounted, but not on CentOS. So this is making the test get skipped on Fedora. What happens if you just let it umount and wipe the partition?

giuseppe: I've tried several times to trigger this test today and most of the time it failed (on the `pvmove`), it looks like it is automatically remounted (just my guess), it passed a few times but it failed in all the other cases.  Given that it has been a flake for the past months on Fedora, I think it is safer to skip it if cannot be trusted at all.

giuseppe:
@jlebon I added a fixup and the tests pass now, we could go with this version and see how the CI will behave.

